PR #3, opened 2018-10-08
Allow multiple hosts to support internal DNS in VPC

michalc:


app/app/settings.py
```diff
@@ -2,18 +2,21 @@
...
 DEBUG = False
 
 def aws_fargate_private_ip():
     with urllib.request.urlopen('http://169.254.170.2/v2/metadata') as response:
         return json.loads(response.read().decode('utf-8'))['Containers'][0]['Networks'][0]['IPv4Addresses'][0]
 
 ALLOWED_HOSTS = \
-    [os.environ['ALLOWED_HOST']] if os.environ['ALLOWED_HOST'] == 'localhost' else \
-    [os.environ['ALLOWED_HOST'], aws_fargate_private_ip()]
+    env['ALLOWED_HOSTS'] if env['ALLOWED_HOSTS'] == ['localhost'] else \
```
lgarvey: Wondering what the value of this line is - env['ALLOWED_HOSTS'] is allowed, only if it's set to ['localhost']?

michalc: > Wondering what the value of this line is - env['ALLOWED_HOSTS'] is allowed, only if it's set to ['localhost']?

Not sure what the best thing here is... there are 2 cases

- If doing local dev, then `env['ALLOWED_HOSTS']` would be set to `['localhost']`, and the setting `ALLOWED_HOSTS` should then also be set to `['localhost']`.

- In production, which we treat as every cases where `env['ALLOWED_HOSTS']` is not equal to `['localhost']` then we have 2 fixed hosts in `env['ALLOWED_HOSTS']`, but then also 1 dynamically generated one (this is added on https://github.com/uktrade/jupyterhub-data-auth-admin/pull/3/files/1aaeee7766cceb94f5b7a39f087299b30a629214#diff-3f9737c05a00f8739afebff6239e18f4R19 )

PR #8, opened 2019-03-15
Save SSO ID (to be able to store DB creds later)

michalc:


app/app/models.py
```diff
@@ -1,8 +1,26 @@
...
-from django.db import models
 from django.core.validators import RegexValidator
+from django.db import models
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+
+class Profile(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    sso_id = models.UUIDField(unique=True, default=uuid.uuid4)
```
lgarvey: Presume you're setting a default=uuid=uuid4 because of existing users, but that it'll get overwritten when the user access the app?

PR #55, opened 2019-06-21
Structure for storing required permissions against dataset

michalc:


michalc:
Will have row-based permissions on datasets instead, at least for now

PR #65, opened 2019-06-25
work in progress requesting access to a dataset

markhigham:


app/app/templates/dataset.html
```diff
@@ -65,13 +153,13 @@ <h2 class="govuk-heading-l govuk-!-margin-top-8">Data Links</h2>
                 {% for link in links %}
                     <tr class="govuk-table__row">
                         <td class="govuk-table__cell">
-                            {% if perms.app_sourcelink.change %}
```
markhigham: @michalc here

app/app/templates/dataset.html
```diff
@@ -109,14 +202,7 @@ <h2 class="govuk-heading-l govuk-!-margin-top-8">Data Links</h2>
                         </td>
                     </tr>
                 {% endif %}
-                {% if perms.app_sourcelink.add %}
```
markhigham: @michalc and here too

app/app/models.py
```diff
@@ -272,7 +272,7 @@ class DataSet(models.Model):
             ('REQUIRES_AUTHENTICATION', 'Requires authentication'),
             ('REQUIRES_AUTHORIZATION', 'Requires authorization'),
         ),
-        default='REQUIRES_AUTHENTICATION',
+        default='REQUIRES_AUTHORIZATION',
```
markhigham: @michalc this was requested by the data team - to make the datasets more locked down by default

michalc: That's fine, although just to check that in the creation form the checkbox will be ticked by default as well?

michalc: Ah! I see the code that does this now

app/app/templates/dataset.html
```diff
@@ -4,6 +4,64 @@
...
+    </style>
+{% endblock %}
+
+{% block bottom_page_scripts %}
+
+
+    <script>
+        (function (document) {
+            function ready(fn) {
+                if (document.attachEvent ? document.readyState === "complete" : document.readyState !== "loading") {
```
michalc: This if will never be true, since this script will always run before `DOMContentLoaded`? If so, better to remove, since it's dead code?

app/app/templates/dataset.html
```diff
@@ -4,6 +4,64 @@
...
+        }
+
+        .request-access-container--visible {
+            max-height: 500px;
+        }
+
+    </style>
+{% endblock %}
+
+{% block bottom_page_scripts %}
```
michalc: Since the script will run after domcontentloaded anyway, can this go in the `head` block?

app/app/templates/dataset.html
```diff
@@ -4,6 +4,64 @@
...
+                    fn();
+                } else {
+                    document.addEventListener('DOMContentLoaded', fn);
+                }
+            }
+
+            var $toggle, $requestAccessContainer;
+
+            function onToggleClick(e) {
+                $requestAccessContainer.classList.toggle('request-access-container--visible');
```
michalc: Just to confirm, it's not better to have the request access form on a whole separate page to keep it simpler, and not use Javascript at all?

app/app/templates/dataset.html
```diff
@@ -4,6 +4,64 @@
...
+            }
+
+            ready(function () {
+                console.log('ready')
+                $toggle = document.getElementById('toggleRequestLink');
+                $requestAccessContainer = document.getElementById('requestAccessContainer')
+
+                $toggle.addEventListener('click', onToggleClick);
+
+                {% if form.justification.errors %}
```
michalc: Since this runs after domcontentloaded, does this then result in a transition just after page load?

app/app/templates/dataset.html
```diff
@@ -52,6 +109,44 @@ <h3 class="govuk-heading-l">Contact</h3>
...
+                </div>
+                <div class="govuk-warning-text request-access-container" id="requestAccessContainer">
+                    <div class="govuk-warning-text__text">
+                        <form method="post">
+                            {% csrf_token %}
+                            <fieldset class="govuk-fieldset">
+                                <div class="govuk-form-group">
+                                    <label class="
+                                            {{ form.justification.errors|yesno:"govuk-error-message,govuk-hint" }}"
+                                           id="problem-hint" for="justification">
```
michalc: Is `id="problem-hint"` not used anywhere, and if no, remove?

app/app/templates/dataset.html
```diff
@@ -52,6 +109,44 @@ <h3 class="govuk-heading-l">Contact</h3>
...
+                                        Please provide justification supporting your request for access.
+                                    </label>
+                                    <textarea class="govuk-textarea" id="justification" name="justification" rows="5"
+                                              aria-describedby="justification" required></textarea>
```
michalc: Does the `aria-describedby="justification"` mean that the `textarea` is described by itself? Should it refer to another element?

app/app/templates/dataset.html
```diff
@@ -52,6 +109,44 @@ <h3 class="govuk-heading-l">Contact</h3>
...
+                        <strong>
+                            <span class="govuk-warning-text__assistive">Warning</span>
+                            You do not currently have access to download these links.
+                            <br>
+                            <a class="govuk-link--no-visited-state" href="#" id="toggleRequestLink">Request access
+                                now</a>
+                        </strong>
+                    </div>
+                </div>
+                <div class="govuk-warning-text request-access-container" id="requestAccessContainer">
```
michalc: I might lean to ids being in kebab-case, but no biggy

app/app/templates/dataset.html
```diff
@@ -83,11 +182,16 @@ <h2 class="govuk-heading-l govuk-!-margin-top-8">Data Links</h2>
                 {% for database, schema, table, access in database_schema_table_accesses %}
                     <tr class="govuk-table__row">
                         <td class="govuk-table__cell">
-                            <a
-                                class="govuk-link"
-                                href="{% url 'table_data' database schema table %}">
+                            {% if access %}
```
michalc: Should requesting access surround the download links?

app/app/templates/privacy.html
```diff
@@ -0,0 +1,37 @@
...
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <h1 class="govuk-heading-l">Privacy Policy</h1>
+
+            <p class="govuk-body">
+                {% lorem 50 w  %}
+            </p>
+
+            <p class="govuk-body">
+                {% lorem 50 w random %}
```
michalc: If going live, I would probably lean to not including this?

markhigham: done. Replaced with the official text

app/app/templates/dataset.html
markhigham: Agreed - Removed all this JS and moved to separate view

app/app/templates/dataset.html
markhigham: Agreed. I removed this from the `_base.html`

app/app/templates/dataset.html
markhigham: Done

app/app/templates/dataset.html
markhigham: Done

app/app/templates/dataset.html
markhigham: Done

app/app/templates/dataset.html
markhigham: Done

app/app/templates/dataset.html
markhigham: Done

app/app/templates/privacy.html
markhigham: Done again. The page was removed and changed to a link to privacy policy here https://workspace.trade.gov.uk/working-at-dit/policies-and-guidance/data-workspace-privacy-policy/

PR #70, opened 2019-07-01
Celery

michalc:


app/app/spawner.py
```diff
@@ -35,34 +43,29 @@ class ProcessSpawner():
     '''
 
     @staticmethod
-    def spawn(_, __, application_instance_id, spawner_options, ___, set_id, set_url):
-        cmd = json.loads(spawner_options)['CMD']
-        proc = None
-
-        def _spawn():
-            nonlocal proc
+    def spawn(_, __, application_instance_id, spawner_options, ___):
```
markhigham: do the _ and __ variables represent variables that aren't used. Is this because they were previously used and you didn't want to break calling code?

markhigham: I've just read the rest of the code specifically the get_spawner method call...  this makes sense to me 👍

PR #72, opened 2019-07-01
Hide request access link on empty datasets

niross:


app/app/templates/dataset.html
```diff
@@ -47,7 +47,7 @@ <h3 class="govuk-heading-l">Contact</h3>
         <div class="govuk-grid-column-full">
             <h2 class="govuk-heading-l govuk-!-margin-top-8">Data Links</h2>
 
-            {% if must_request_download_access %}
+            {% if links and must_request_download_access %}
```
markhigham: There are two types of links that are appended to that html table
the others come from `{% for database, schema, table, access in database_schema_table_accesses %}`

niross: Good catch :+1: I've updated the template

PR #77, opened 2019-07-08
(feat) Show running containers in admin

michalc:
This is to build on in later PRs to..
- Show max CPU in previous X hours using prometheus data
- To then use this information to kill containers


michalc:
Will merge later PR

PR #84, opened 2019-07-10
Feature/basic django tests

niross:
Start running django tests as part of deployment.


.envs/test.env
```diff
@@ -0,0 +1,31 @@
...
+ZENDESK_TOKEN=XXXX
+ZENDESK_SERVICE_FIELD_ID=XXX
+ZENDESK_SERVICE_FIELD_VALUE=xxx
+APPLICATION_TEMPLATES__1__NAME=testapplication
+APPLICATION_TEMPLATES__1__NICE_NAME=Test Application
+APPLICATION_TEMPLATES__1__SPAWNER=PROCESS
+APPLICATION_TEMPLATES__1__SPAWNER_OPTIONS__CMD__1=python3
+APPLICATION_TEMPLATES__1__SPAWNER_OPTIONS__CMD__2=/test/echo_server.py
+PROMETHEUS_DOMAIN=some.domain.com
+DJANGO_SETTINGS_MODULE=app.settings
```
michalc: If this is `app.settings`, then it looks like the `app.test_settings` module below is unused? If so, remove it? (I would lean to using the same settings module if possible since there are a fair few things happening in there and so would be good for the tests to use it I think.)

docker-compose-test.yml
```diff
@@ -5,6 +5,8 @@ services:
       context: .
       dockerfile: Dockerfile-test
     image: data-workspace-test
+    env_file:
+      - ./.envs/test.env
```
michalc: If these variables are only required for the django tests, I would lean to either...

- having two different docker-compose files for the two sorts of tests, each with only what they need and no more
- making it so each tests use the variables from the environment. This would probably involve changing test.py to not use its hard coded `APP_ENV` and use its environment.

app/app/tests/common.py
```diff
@@ -0,0 +1,27 @@
+import uuid
+
+from django.test import TestCase
+
+
+class BaseTestCase(TestCase):
+    def setUp(self):
+        self.user_data = {
+            'HTTP_SSO_PROFILE_EMAIL': 'bob.testerson@test.com',
+            'HTTP_SSO_PROFILE_USER_ID': uuid.uuid4(),
```
michalc: A minor thing but this (and the use of uuid4 below) makes each test run not identical? In certain cases things end up ordered by id, so it could result in a difference of behaviour / flakiness, depending on the test.

michalc:
> Start running django tests as part of deployment.

Also just a minor thing, at the moment at least, CircleCI isn't related to deployment. It runs the test and then deployment is a separate manual process in AWS

.envs/test.env
niross: Yeah good catch I wasn't planning to include test_settings.py at all - I was just using that for testing locally.

docker-compose-test.yml
niross: I guess another option is to keep `test_settings.py` above for django tests and get rid of env vars in the docker container.

michalc: Ah and have the values hard-coded in `test_settings.py`? Yes... although I would be slightly anti the tests not using the same settings file, just in case it leads to tests passing when they shouldn't due to a test/production settings file difference. There are some fairly vital bits of code in the settings file.

niross: Ok fair enough. I think i'm leaning towards the second option then - removing APP_ENV altogether :+1:

michalc: Sounds good... actually should be straightforward: hopefully just just be a replace of `APP_ENV` with `os.environ`

PR #89, opened 2019-07-12
Feature/reference data admin

niross:


app/app/models.py
```diff
@@ -319,7 +313,281 @@ class SourceLink(models.Model):
...
+        """
+        Return a list of associated records containing the internal id and row data
+        :return:
+        """
+        records = []
+        if self.field_names:
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    'SELECT dw_int_id, {} FROM {} ORDER BY {}'.format(
+                        ', '.join(self.field_names),
```
michalc: Can the various values be escaped as a defence against sql injection (and elsewhere that has similar usage)? e.g. by using `sql.SQL` and `sql.Identifier`?

I'm not completely sure right now if an injection attack is even possible, but even if there is some logic/validation elsewhere in the code that prevents it, but I suspect it's good to have escaping here in any case, right at the point of use as a last line of defence.

niross: I didn't think they could but I just dived back into the docs and yes it turns out they can

app/app/urls.py
```diff
@@ -135,7 +138,8 @@ def _login_required(request, *args, **kwargs):
     path('request-access/<str:group_slug>/<str:set_slug>', login_required(request_access_view), name='request_access'),
     path('request_access_success/', login_required(request_access_success_view),
          name='request_access_success'),
-
+    path('admin/', include(('app.wd_admin.urls', 'app'), namespace='wd-admin')),
```
markhigham: what is wd short for?

PR #90, opened 2019-07-16
Feature/reference data ui

niross:
Adds reference data sets to user facing site


app/app/templates/_base.html
```diff
@@ -19,6 +19,7 @@
   <link rel="apple-touch-icon" sizes="152x152" href="{% get_static_prefix %}assets/images/govuk-apple-touch-icon-152x152.png">
   <link rel="apple-touch-icon" href="{% get_static_prefix %}assets/images/govuk-apple-touch-icon.png">
   <link rel="stylesheet" href="{% get_static_prefix %}govuk-frontend-including-fonts-2.12.0-v1.min.css">
+  <link rel="stylesheet" href="{% get_static_prefix %}data-workspace.css">
 
   <style>
     .govuk-panel--confirmation--left {
```
michalc: If using a separate CSS file, then this can be moved into it for consistency? (Or, keep all the site-specific CSS in the <style> element?)

app/app/templates/app/referencedataset_detail.html
```diff
@@ -0,0 +1,197 @@
...
+          <tr class="govuk-table__row">
+            <th class="govuk-table__header">Link to the data</th>
+            <th class="govuk-table__header">Format</th>
+            <th class="govuk-table__header">Last Updated</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell">
+              <a class="govuk-link" target="_blank" href="{% url 'reference_dataset_download' object.group.slug object.slug 'json'%}">
```
michalc: Do downloads act a bit weird with `target="_blank"`?

app/app/templates/app/referencedataset_detail.html
```diff
@@ -0,0 +1,197 @@
...
+    </div>
+  </div>
+
+  <div class="govuk-grid-row" style="overflow-x: auto;">
+    <div class="govuk-grid-column-full">
+      <h2 class="govuk-heading-m govuk-!-margin-top-8">
+        Data Preview
+      </h2>
+      <p class="govuk-body">Showing <strong>5</strong> of <strong>{{ object.get_records|length }}</strong> records</p>
+      <table class="govuk-table catalogue-table govuk-!-font-size-16">
```
michalc: Although I think it's already in the codebase, does `catalogue-table` have any associated styles/use? Remove if not?

app/app/templates/app/referencedataset_detail.html
```diff
@@ -0,0 +1,197 @@
...
+        <thead>
+          <tr class="govuk-table__row">
+            {% for field_name in object.field_names %}
+              <th class="govuk-table__header">{{ field_name|title }}</th>
```
michalc: Just to check, `field_name` ever be multi-word? If so, then `| title` will capitalise each word, when I think for headings sentence case should be used (at least according to https://design-system.service.gov.uk/styles/typography/ ). Will `| capfirst` be more appropriate? (and ditto for other uses)

app/app/templates/app/referencedataset_detail.html
niross: I kept target blank because it's not actually a download but displays as a we page. I'm not using `Content-Disposition: attachment` here. Maybe I should be. I will check

app/app/templates/app/referencedataset_detail.html
```diff
@@ -0,0 +1,197 @@
...
+            </tr>
+          {% endfor %}
+        </tbody>
+      </table>
+    </div>
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h2 class="govuk-heading-l govuk-!-margin-top-8">Additional Information</h2>
```
michalc: As in other uses, should the heading only have the first word capitalised?

app/app/templates/app/referencedataset_detail.html
```diff
@@ -0,0 +1,197 @@
...
+          </dt>
+          <dd class="govuk-summary-list__value">
+            <a class="govuk-link" href="{% url 'datagroup_item' object.group.slug %}">
+              {{ object.group.name }}
+            </a>
+          </dd>
+        </div>
+        <div class="govuk-summary-list__row">
+          <dt class="govuk-summary-list__key">Licence</dt>
+          <dd class="govuk-summary-list__value">{{ object.licence|default:"" }}</dd>
```
michalc: Should `licence` not be nullable in the db, and then `|default:""` wouldn't be required (I think)?

app/app/templates/_base.html
niross: Keen to steer clear of inline styles so have added them to the css :+1:

app/app/templates/app/referencedataset_detail.html
niross: I've made it a download also. May as well be consistent.

app/app/templates/app/referencedataset_detail.html
niross: You're correct it's not actually used. I've removed any references to it.

app/app/templates/app/referencedataset_detail.html
niross: Yep, it should only ever have underscores. There is a slight change in how things work, we're going to enforce lowercase for field/column names now. So I guess that should be represented on this page. Or is it still best practice to use sentence case when displaying here?

app/app/templates/app/referencedataset_detail.html
niross: Indeed, made non-nullable

app/app/templates/app/referencedataset_detail.html
michalc: Not entirely sure what's best, but for these "data" headers that might contain underscores, lowercase seems good to me

app/app/dw_admin/forms.py
```diff
@@ -2,6 +2,13 @@
 
 
 class ReferenceDataFieldInlineForm(forms.ModelForm):
+    name = forms.CharField(
+        widget=forms.TextInput(
+            attrs={
+                'style': 'text-transform:lowercase'
```
michalc: If someone enters something (perhaps accidentally) in ALL CAPS or MiXeD CaSe, would it be presented as lower, but stores the mixed case in the db? Would it be better to have validation if we want always lower?

michalc: Oh I see there is validation... in the case, can the validation fail slightly magically, after a user enters something that appears to be in lower case, it can fail because they had caps lock on?

app/app/models.py
```diff
@@ -573,9 +595,12 @@ class ReferenceDatasetField(TimeStampedUserModel):
...
     name = models.CharField(
         max_length=60,
-        help_text='The name of the field. May only contain letters '
-                  'numbers and underscores (no spaces)',
-        validators=[RegexValidator(regex=r'^[a-zA-Z][a-zA-Z0-9_\.]*$')]
+        help_text='The name of the field. May only contain lowercase '
+                  'letters numbers and underscores (no spaces)',
+        validators=[RegexValidator(
+            regex=r'^[a-z][a-z0-9_\.]*$',
+            message='Please only enter lowercase letters, numbers and underscores'
```
michalc: Not sure if it's important, but the allowed characters don't quite match the message (although I also son't think it did before)

app/app/templates/app/referencedataset_detail.html
```diff
@@ -0,0 +1,201 @@
...
+    </div>
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h2 class="govuk-heading-l govuk-!-margin-top-8">Additional information</h2>
+      <dl class="govuk-summary-list">
+        <div class="govuk-summary-list__row">
+          <dt class="govuk-summary-list__key">
+            Published Date
```
michalc: A few more Title Case that should be Sentence case?

app/app/templates/datagroup.html
```diff
@@ -54,6 +53,26 @@ <h2 class="govuk-heading-s">
                        href="{% url 'admin:app_dataset_add' %}">[New]</a>
                 </p>
             {% endif %}
+            {% if model.referencedataset_set.live.exists %}
+              <h2 class="govuk-heading-l govuk-!-margin-top-8">Available Reference Data Sets</h2>
```
michalc: More title case?

app/app/tests/test_views.py
```diff
@@ -56,3 +67,67 @@ def test_dataset_detail_view(self):
...
+            is_identifier=True
+        )
+        response = self._authenticated_get(
+            reverse('reference_dataset_download', kwargs={
+                'group_slug': group.slug,
+                'reference_slug': rds.slug,
+                'format': 'json',
+            })
+        )
+        self.assertEqual(response.status_code, 200)
```
michalc: Would it be good to check a bit more than the status code, i.e. that the body of the download looks reasonable?

michalc: And to add to this, would be good if the test contain some non-ascii data, to make sure there aren't encoding issues? (I _think_ it should all default to utf-8, but I'm not 100%)

app/app/views_catalogue.py
```diff
@@ -150,3 +157,55 @@ def dataset_full_path_view(request, group_slug, set_slug):
...
+        return get_object_or_404(
+            ReferenceDataset,
+            deleted=False,
+            group=group,
+            slug=self.kwargs.get('reference_slug')
+        )
+
+
+class ReferenceDatasetDownloadView(ReferenceDatasetDetailView):
+    lookup_value_regex = r"[\w.]+"
```
michalc: Is this value used? Is it more a DRF thing?

niross: Yep, accidentally left that one in

app/app/templates/app/referencedataset_detail.html
niross: Fixed. This is prevalent throughout the existing code so maybe there is a ticket to be created around this

app/app/views_catalogue.py
```diff
@@ -150,3 +157,55 @@ def dataset_full_path_view(request, group_slug, set_slug):
...
+        if dl_format not in ['json', 'csv']:
+            raise Http404
+        ref_dataset = self.get_object()
+        field_names = ref_dataset.field_names
+        records = []
+        for row in ref_dataset.get_records():
+            record = {}
+            for idx, field in enumerate(field_names):
+                record[field] = row['data'][idx]
+            records.append(record)
```
michalc: This is leaning towards style (and so _not_ doing this is all good) but to me at least comprehensions are clearer here (and I would argue a touch safer due to less mutation)
```
records = [
  {
    field: row['data'][idx]
    for idx, field in enumerate(field_names)
  }
  for row in ref_dataset.get_records()
]
```

app/app/dw_admin/forms.py
niross: Yep happy to remove it and update the help text accordingly

app/app/tests/test_views.py
niross: Ok i've updated the tests here

app/app/models.py
niross: Ok, I've updated the error and help messages accordingly

app/app/dw_admin/forms.py
```diff
@@ -10,6 +10,9 @@ class ReferenceDataFieldInlineForm(forms.ModelForm):
         )
     )
 
+    def clean_name(self):
+        return self.cleaned_data['name'].lower()
```
michalc: Is this function not used? (And if it is, is `lower` unnecessary since uppercase characters fail validation?)

app/app/views_catalogue.py
niross: Hmm... yeah agreed it is easier on the eye

app/app/dw_admin/forms.py
niross: Yes not necessary anymore

PR #91, opened 2019-07-17
Allow the hiding of datasets by marking them unpublished

niross:


michalc:
I suspect that for the tests to pass in `test.py` in the `create_private_dataset` function the DataSet would need to be set as published

michalc:
Ah I see it's done!

niross:
Yes good timing :)

PR #108, opened 2019-07-26
Feature/source link upload

niross:
This adds the ability to upload files to s3 and point source links to them.

* Added an (internal) type field to `SourceLink` with two options, local/external.
* External links will still work as usual.
* Uploading a file to s3 will create a new `SourceLink` with type set to internal.
* The user is able to change the link to an external one at a later time (as requested). This will automatically set the link type to external and delete the file from s3.

https://trello.com/c/S3gwmZa6/192-add-file-upload-for-data-sets


app/app/static/data-workspace.css
```diff
@@ -4,3 +4,34 @@
...
+  padding: 10px 15px;
+  margin: 0 0 0 10px;
+  float:left;
+  border: none;
+  color: #333 !important;
+  cursor: pointer;
+}
+.back-link:hover {
+  background: #ccc;
+}
```
yusufertekin: This css file is already imported in _base.html. I saw that you've added this file to DatasetAdmin as a model specific css file. If you only want to add this bit of css to specific ModelAdmin, you might want to create a new file and refer that in Media inner class of Dataset admin to prevent possible conflicts on templates inherited from _base.html.

app/app/templates/admin/dataset_source_link_upload.html
```diff
@@ -0,0 +1,49 @@
+{% extends "admin/change_form.html" %}
+{% load static admin_urls %}
+{% block extrahead %}
+  {{ block.super }}
+  <script type="text/javascript" src="{% url 'admin:jsi18n' %}"></script>
+  {{ media }}
+  <link rel="stylesheet" href="{% get_static_prefix %}data-workspace.css">
+{% endblock %}{% block extrastyle %}
+  {{ block.super }}
+  <link rel="stylesheet" href="{% static "admin.css" %}">
```
yusufertekin: File does not exists.

app/app/dw_admin/urls.py
```diff
@@ -18,4 +21,9 @@
         view=ReferenceDatasetAdminDeleteView.as_view(),
         name='reference-dataset-record-delete'
     ),
+    path(
+        'app/dataset/<str:dataset_id>/sourcelink/upload/',
+        view=SourceLinkUploadView.as_view(),
+        name='source-link-upload'
+    ),
```
yusufertekin: These views should be login and admin user required. Im able to access all the urls defined here without authentication.

app/app/templates/admin/dataset_source_link_upload.html
```diff
@@ -0,0 +1,49 @@
+{% extends "admin/change_form.html" %}
+{% load static admin_urls %}
+{% block extrahead %}
+  {{ block.super }}
+  <script type="text/javascript" src="{% url 'admin:jsi18n' %}"></script>
+  {{ media }}
```
yusufertekin: This bit of html is already extended from admin/change_form and due to use of {{ block.super }} it's a duplication.

app/app/templates/admin/dataset_source_link_upload.html
```diff
@@ -0,0 +1,49 @@
...
+      {% if form.non_field_errors %}
+        <p class="errornote">
+          {% if errors|length == 1 %}
+            Please correct the error below.
+          {% else %}
+            Please correct the errors below.
+          {% endif %}
+        </p>
+        {{ form.non_field_errors }}
+      {% endif %}
```
yusufertekin: You might want to add 'errors' variable in context then this html code can be removed and django default change_form template has an identical code bit which will display the errors.

app/app/tests/test_admin.py
```diff
@@ -707,3 +706,40 @@ def test_reference_data_record_delete(self):
...
+                'file': file1,
+            }
+        )
+        self.assertContains(response, 'Source link uploaded successfully')
+        self.assertEqual(link_count + 1, dataset.sourcelink_set.count())
+        link = dataset.sourcelink_set.latest('created_date')
+        self.assertEqual(link.name, 'Test source link')
+        self.assertEqual(link.format, 'CSV')
+        self.assertEqual(link.frequency, 'Never')
+        mock_client.assert_called_once()
```
yusufertekin: It might help us to cover more in depth, if we test whether or not mocked boto3 client's put_object is called with expected 
 - Body
 - Bucket
 - Key

app/app/dw_admin/forms.py
```diff
@@ -54,3 +61,58 @@ def clean_identifier(form):
...
+        )
+        client = boto3.client('s3')
+        try:
+            client.put_object(
+                Body=self.cleaned_data['file'],
+                Bucket=settings.AWS_UPLOADS_BUCKET,
+                Key=self.instance.url
+            )
+        except ClientError as ex:
+            raise Exception(
```
yusufertekin: This exception isn't getting caught later on so raises exception to the user directly. 
My suggestions:
- Use the form for validation
- Move s3 file upload to at the end of form_valid function of SourceLinkUploadView
- Don't raise exception, return possibly HTTP 500

app/app/models.py
```diff
@@ -300,6 +303,12 @@ class SourceTable(models.Model):
 
 
 class SourceLink(TimeStampedModel):
+    TYPE_EXTERNAL = 1
+    TYPE_LOCAL = 2
+    _LINK_TYPES = (
```
yusufertekin: Why private naming?

app/app/models.py
```diff
@@ -309,19 +318,62 @@ class SourceLink(TimeStampedModel):
...
+        client = boto3.client('s3')
+        try:
+            client.delete_object(
+                Bucket=settings.AWS_UPLOADS_BUCKET,
+                Key=self.url
+            )
+        except ClientError as ex:
+            raise Exception(
+                'Error deleting file: {}'.format(ex.response['Error']['Message'])
+            )
```
yusufertekin: Exception is not getting handled will reach to users directly.

app/app/tests/test_admin.py
```diff
@@ -707,3 +706,40 @@ def test_reference_data_record_delete(self):
...
         self.assertEqual(num_records - 1, len(reference_dataset.get_records()))
+
+
+class TestSourceLinkAdmin(BaseAdminTestCase):
+    def test_source_link_upload_get(self):
+        dataset = factories.DataSetFactory.create()
+        response = self._authenticated_get(
+            reverse('dw-admin:source-link-upload', args=(dataset.id,))
+        )
+        self.assertContains(response, 'Upload source link')
```
yusufertekin: Might be worth to create a test for possible client error.

app/app/tests/test_models.py
```diff
@@ -277,3 +279,45 @@ def test_delete_record(self):
...
+            id='158776ec-5c40-4c58-ba7c-a3425905ec45',
+            dataset=dataset,
+            link_type=models.SourceLink.TYPE_LOCAL,
+            url='s3://sourcelink/158776ec-5c40-4c58-ba7c-a3425905ec45/test.txt'
+        )
+        link.delete()
+        self.assertFalse(
+            models.SourceLink.objects.filter(id='158776ec-5c40-4c58-ba7c-a3425905ec45').exists()
+        )
+        mock_client.assert_called_once()
```
yusufertekin: It might help us to cover more in depth, if we test whether or not mocked boto3 client's delete_object is called with expected
- Bucket
- Key

app/app/models.py
niross: To check things like `source_link.link_type == source_link.TYPE_EXTERNAL`

app/app/static/data-workspace.css
niross: I've split them out

app/app/templates/admin/dataset_source_link_upload.html
niross: I think this is embedded somewhere in the content block though right? So to make that work we'd need top use the whole block and would lose any customisation in the template?

app/app/models.py
yusufertekin: ClientError for deleting is not getting handled. On the time of error, ugly message will be displayed directly to the user. Only place we use delete is in the Dataset admin. It gets deleted as a part of formset. I couldn't think a clear way to handle this. Options on top of my head which Im not sure if they actually work;
- Create custom formset for dataadmin and override delete_existing() which would raise validation error
- Monkey patch auto generated formset with custom delete_existing()  which would raise validation error

Im happy to merge this as it is, and create a ticket to handle client errors on deletion. This challange/problem might be included in the discussions of scope of usage of django admin framework.

yusufertekin:
Thank you! LGTM except we might think about ClientError handling on the time of deletion.

app/app/models.py
niross: Yeah agreed it's not ideal, Using `delete_existing` on a custom formset will not work as raising a validation error there still shows a 500 error to the user.

One thing that will work, and is a _little_ bit nicer, is to check that the file is accessible to us when validating the formset. If the file doesn't exist we can raise a form level error saying it can't be deleted. If the file does exist we can proceed with the delete. If the file exists and it can't be deleted for whatever reason it will still throw a 500 as it currently does. But yes it's something we should fix when we roll our own admin.

yusufertekin:
LGTM!

PR #109, opened 2019-07-28
Add reference dataset major/minor versioning

niross:
Add major/minor versioning to reference dataset model.

* Increment major version when a schema changes is made
* Increment minor version when a record is added, edited or updated

Trello card: https://trello.com/c/NA3J2yCq/149-reference-data-set-version-numbering


app/app/tests/test_models.py
```diff
@@ -56,6 +56,7 @@ def test_create_reference_dataset(self):
 
         # Ensure the table was created in the db
         self.assertTrue(self._table_exists(ref_dataset.table_name))
+        self.assertEqual(ref_dataset.major_version, 1)
```
yusufertekin: We might want to test schema_version values on the event of 
- creating reference dataset
- deleting reference dataset
- creating reference dataset field
- deleting reference dataset field
- editing reference dataset field

yusufertekin:
Thank you! LGTM!

PR #111, opened 2019-08-07
Bugfix/no datasets ref data fix

niross:
Do not show the 'no datasets available' message if there are reference datasets (and vice versa)

Trello card: https://trello.com/c/AeSzpsR0/216-no-data-sets-available-on-group-page-with-reference-dataset

### Before
Screenshot from 2019-08-07 15-53-51

### After
Screenshot from 2019-08-07 15-54-18


yusufertekin:
LGTM!

PR #132, opened 2019-08-16
(refactor) Bring in mirrors-sync

michalc:
Please forgive the code... it was a rush and always on the to-tidy
list...


yusufertekin:
LGTM!

PR #135, opened 2019-08-16
Feature/200 synching back end tables with the analysts database

niross:
Sync reference data with associated external databases

* Add customisable table name to `ReferenceDataset`
* Add customisable column name to `ReferenceDatasetField`
* Add new `ReferenceDataExternalDatabase` model
* On `ReferenceDataExternalDatabase` creation sync associated ref data tables and data to external database
* On `ReferenceDataset` create/edit sync changes to all external database tables
* On `ReferenceDatasetField` edit sync column changes to all external database tables
* On reference dataset record add/edit/delete sync changes to associated external tables


dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -258,6 +259,20 @@ class ReferenceDataset(DeletableTimestampedUserModel):
...
     )
+    table_name = models.CharField(
+        verbose_name='Table name',
+        max_length=255,
+        unique=True,
+        help_text='Descriptive table name for the field - '
+                  'Note: Must start with "ref_" and contain only letters, numbers and underscores',
+        validators=[
+            RegexValidator(
+                regex=r'^ref_[a-zA-Z0-9_]*$',
```
michalc: Would forcing table names to always be lowercase be better, to avoid having to escape the table names in queries? From https://www.postgresql.org/docs/current/sql-syntax-lexical.html 

> Quoting an identifier also makes it case-sensitive, whereas unquoted names are always folded to lower case.

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -646,3 +778,57 @@ def get_model_field(self):
             max_length=255
         )
         return field
+
+
+class ReferenceDatasetExternalDatabase(TimeStampedUserModel):
```
michalc: Just to check, why allow multiple external databases for the same reference dataset?

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Yeah good call I will update the regex/help text

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Well.. it was to future proof it somewhat. There is no business need as far as I know. Do you think it's not necessary?

.envs/test.env
```diff
@@ -19,6 +19,17 @@ DATA_DB__my_database__USER=postgres
 DATA_DB__my_database__PASSWORD=postgres
 DATA_DB__my_database__HOST=data-workspace-postgres
 DATA_DB__my_database__PORT=5432
+DATA_DB__test_external_db__NAME=testdb1
+DATA_DB__test_external_db__USER=postgres
+DATA_DB__test_external_db__PASSWORD=postgres
+DATA_DB__test_external_db__HOST=data-workspace-postgres
+DATA_DB__test_external_db__PORT=5432
+DATA_DB__test_external_db__TEST__CREATE_DB=False
```
michalc: It looks like CREATE_DB is a oracle setting https://docs.djangoproject.com/en/2.2/ref/settings/#create-db . Is this needed?

dataworkspace/dataworkspace/apps/datasets/migrations/0008_auto_20190813_1544.py
```diff
@@ -0,0 +1,31 @@
...
+                validators=[
+                    django.core.validators.RegexValidator(
+                        message='Column names must start with a letter and contain only '
+                                'letters, numbers, underscores and full stops.',
+                        regex='^[a-zA-Z][a-zA-Z0-9_\\.]*$'
+                    )
+                ]
+            ),
+        ),
+    ]
```
michalc: Can these migrations be squashed one, so that production only goes through the one migration? (To keep things simple in case there are problems)

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -304,16 +319,36 @@ def __str__(self):
...
+        super().__init__(*args, **kwargs)
+        # Stash the current table name so it can be compared on save
+        self._original_table_name = self.table_name
+
     def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
         create = self.pk is None
+        table_changed = self.table_name != self._original_table_name
+        if not create and table_changed:
+            self.schema_version += 1
         super().save(force_insert, force_update, using, update_fields)
```
michalc: Is it worth wrapping this save and the below functions in a transaction, to reduce the chance that they get out of sync if something goes wrong?

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -435,16 +470,29 @@ def _get_record(self, field_name: str, identifier: any) -> Optional[dict]:
...
+        Save a record to the local database and associate it with this reference dataset.
+        Replicate the record in any linked external databases.
         :param internal_id: the django id for the model (None if doesn't exist)
         :param form_data: a dictionary containing values to be saved to the row
         :return:
         """
         self.increment_minor_version()
         if internal_id is None:
-            return self.get_record_model_class().objects.create(**form_data)
+            record = self.get_record_model_class().objects.create(**form_data)
```
michalc: I'm now wondering how this should handle failures saving to the external databases. Say, there are 2 external databases for a reference dataset, and the second one fails and things are left in an inconsistent state. Anything to be done for this case?

dataworkspace/dataworkspace/apps/datasets/migrations/0004_auto_20190813_1536.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 2.2.3 on 2019-08-13 15:36
+
+from django.db import migrations, connection, ProgrammingError
+
+
+def generate_table_name(apps, _):
+    model = apps.get_model('datasets', 'ReferenceDataset')
+    for r in model.objects.all():
+        r._original_table_name = 'refdata__{}'.format(r.id)
```
michalc: Probably a very minor thing, but does _original_table_name need to be on the model? Can it be a local variable?

dataworkspace/dataworkspace/apps/datasets/models.py
michalc: I'm wondering whether actually it's better to _just_ have a function that syncs all the records from the reference data to each dataset, and deleting all that aren't found, rather than responding to individual record changes.

Few reasons:

- Say the reference data is ingested a by some external process into the db: the one case I'm familiar with is when I ingested a list of countries manually. Would be good to be able to sync this, which I don't think is possible in the current code without going through all the records in admin?

- I suspect it's easier to handle error cases, since there will be a single function to sync all the data. For example, could wrap the code in a try/except, and set a flag to say that the last sync errored to prompt the user to try again or report an issue.

- Similarly, say there is a bug and things are a bit wrong, would be easier to fix by just running a single function.

- I suspect a smaller chance of bugs as well, since there would be less code and fewer cases

- It would be easier to do non-synced changes down the line. Say, want to go through a bit of a review/public process.

Potentially this would not be efficient, but I suspect there is not plan to have reference data with more rows than in the low thousands. We could always make it faster later?

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -538,48 +624,71 @@ class ReferenceDatasetField(TimeStampedUserModel):
...
+        """
+        Alter the db column name in the associated table
+        :return:
+        """
+        # Get a copy of the existing model class (pre-save)
+        model_class = self.reference_dataset.get_record_model_class()
+        # Get a copy of the current field
+        from_field = model_class._meta.get_field(self._original_column_name)
+        # Save the changes to the field
+        super().save()
```
michalc: Similar to above, suspect would be good to wrap in a transaction so the save on the local database happens if and only if the save on the remote database happens?

michalc: (Although thinking about it, technically it won't be completely perfect... the remote database could succeed in its update, but then the request time out or something, and we would still be in an inconsistent state...)

dataworkspace/dataworkspace/apps/datasets/models.py
michalc: I'm torn, but I think all-things-being-equal, it looks like it makes things more complicated than they need to be for all known current and future requirements, and I suspect for a good range of unknown ones as well. If for some reason we do need it, there will still be a migration path to do it (albeit will be a bit faffy).

I would say not worth making things definitely more complicated just to avoid an unlikely but doable migration.

dataworkspace/dataworkspace/apps/datasets/migrations/0008_auto_20190813_1544.py
niross: Yes I did consider squashing migrations but i've had issues before mixing schema and runpython so I was trying to avoid any headaches.

https://django.readthedocs.io/en/2.1.x/ref/migration-operations.html

> On databases that do support DDL transactions (SQLite and PostgreSQL), RunPython operations do not have any transactions automatically added besides the transactions created for each migration. Thus, on PostgreSQL, for example, you should avoid combining schema changes and RunPython operations in the same migration or you may hit errors like OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.

dataworkspace/dataworkspace/apps/datasets/models.py
niross: So what are you proposing? A foreign key from `ReferenceDataset` to `Database`? or something else entirely?

michalc: Yes, exactly that

dataworkspace/dataworkspace/apps/datasets/migrations/0004_auto_20190813_1536.py
niross: Yup, not needed at all :+1:

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Yep agree that a "full sync" function would be good. Do you mean run it on every record add/update/delete though?

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Ok, i've updated to use transaction.atomic when a method accesses both internal and external dbs

dataworkspace/dataworkspace/apps/datasets/models.py
niross: OK i've switched to a single external db as a foreign key from `ReferenceDataset`

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Hmm, yeah by all accounts definitely less than 1000 records. Not sure about update frequencies but i would think reference data should not be updated regularly, otherwise it's more of a dataset.

niross: Ok, I've switch to an full sync on record save/delete - I've made it optional as there will be some batch import work coming up where it won't be necessary on every save

dataworkspace/dataworkspace/apps/datasets/models.py
niross: As above I've wrapper in transaction.atomic which should help with this

dataworkspace/dataworkspace/apps/datasets/migrations/0010_auto_20190820_0829.py
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='referencedataset',
+            name='external_database',
+            field=models.ForeignKey(blank=True, help_text='Name of the analysts database to keep in sync with this reference dataset', null=True, on_delete=django.db.models.deletion.CASCADE, to='core.Database'),
+        ),
+        migrations.DeleteModel(
+            name='ReferenceDatasetExternalDatabase',
```
michalc: Can the migrations that aren't necessary be removed?

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -310,16 +333,67 @@ def __str__(self):
             self.name
         )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Stash the current table name & db so they can be compared on save
+        self._original_table_name = self.table_name
+        self._original_ext_db = self.external_database \
+            if self.external_database is not None else None
```
michalc: Is this equivalent to `self._original_ext_db = self.external_database`, without the ternary?

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -310,16 +333,67 @@ def __str__(self):
...
+                self._create_external_database(self.external_database.memorable_name)
+        else:
+            if self.external_database != self._original_ext_db:
+                # If external db has been changed delete the original table
+                if self._original_ext_db is not None:
+                    self._drop_external_database(self._original_ext_db.memorable_name)
+                # if external db is now set create the table and sync existing records
+                if self.external_database is not None:
+                    self._create_external_database(self.external_database.memorable_name)
+                    self.sync_to_external_database(self.external_database.memorable_name)
```
michalc: Would it be better to sync to the new database before deleting the old one? (Not too much in it though since it is essentially a cache)

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -310,16 +333,67 @@ def __str__(self):
...
+
+            # If the db has been changed update it
+            if table_changed:
+                for database in self.get_database_names():
+                    with connections[database].schema_editor() as editor:
+                        editor.alter_db_table(
+                            model_class,
+                            self._original_table_name,
+                            self.table_name
+                        )
```
michalc: Just working through the cases of something going wrong... say for example

- The local table rename worked
- The external table rename failed, and raised an exception

Then the current model would rollback, but (and this I'm not sure on) would the local table _not_ rollback since `with connections[database].schema_editor()` could have been using a different connection? Then the databases would be in an inconsistent state.

I don't think there is perfect way to handle this? Something like...
- If a later database in `self.get_database_names()` change fails, should all the earlier successful changes be undone?
- Should there be checks over the dbs ahead of time that the renaming is likely to work, before anything is changed in any database?

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -310,16 +333,67 @@ def __str__(self):
...
+
+        self._original_table_name = self.table_name
+        self._original_ext_db = self.external_database \
+            if self.external_database is not None else None
+
+    @transaction.atomic
+    def delete(self, **kwargs):
+        # Delete external table when ref dataset is deleted
+        if self.external_database is not None:
+            self._drop_external_database(self.external_database.memorable_name)
```
michalc: To handle the various cases, should this _not_ raise an exception if the external table already doesn't exist. Say, to handle the case. A request to delete _did_ successfully delete the table, but the reply from the db saying so took too long, or the Python server happened to die at that exact moment.

(Also: I keep getting scared by the name `_drop_external_database`... better to be `_drop_external_table`?)

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Yes, fixed, was a throwback from accessing the database name here which would fail if the relation didn't exist.

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Not sure I see why it makes a difference to be honest

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Makes sense, I've added exception handling here and renamed the two table altering methods

dataworkspace/dataworkspace/apps/datasets/models.py
michalc: Ah: I wonder essentially if the table creation and deletion should be more like "ensure exists" or "ensure deleted". Since ReferenceDataSet.table_name is marked as unique, this would probably be robust to various error cases?

niross: In my testing the local table has always rolled back if the external table change fails. 

I have added a unit test in `test_models.py` which forces an error on the external db at this point and ensures both local and external tables remain unchanged.

niross: Oh and for point two, I can add a check that the table is accessible if it helps you sleep at night :)

michalc: > In my testing the local table has always rolled back if the external table change fails.

Ah that's good

> Oh and for point two, I can add a check that the table is accessible if it helps you sleep at night :)

If the local change always rolls back if the external fails, then I think it's unnecessary, so I would be happy without

dataworkspace/dataworkspace/tests/test_models.py
```diff
@@ -889,3 +889,31 @@ def test_create_reference_dataset_external_error(self, mock_conn):
         self.assertFalse(
             self._table_exists('test_create_external_error', database='test_external_db')
         )
+
+    def test_edit_reference_dataset_external_error(self):
+        # Test that when an exception occurs during external database save
+        # the local database is rolled back
```
michalc: 👍

PR #146, opened 2019-08-19
(fix) Fix registry for Fargate: no auth

michalc:
The latest version appears to need authentication set, which doesn't
work in Fargate (at least, not without a bit of faff). However, we are
happy without any auth since the registry is only accessible from inside
the application VPC.


yusufertekin:
So downgrading disables the authentication?

PR #150, opened 2019-08-20
OMISDataset model and its API endpoints

yusufertekin:


dataworkspace/proxy.py
```diff
@@ -478,6 +485,53 @@ def authenticate_by_basic_auth():
...
+                    return False
+
+        @web.middleware
+        async def _authenticate_by_hawk_auth(request, handler):
+            hawk_auth_required = is_machine_to_machine(request)
+
+            if not hawk_auth_required:
+                return await handler(request)
+
+            content = await request.content.read()
```
michalc:
```
content = await request.read()
```

dataworkspace/proxy.py
```diff
@@ -481,6 +492,55 @@ def authenticate_by_basic_auth():
...
+
+        async def seen_nonce(nonce, sender_id):
+            nonce_key = f'nonce-{sender_id}-{nonce}'
+            with await redis_pool as conn:
+                nonce_stored = await conn.execute('GET', nonce_key)
+                if nonce_stored:
+                    return True
+                else:
+                    await conn.execute('SET', nonce_key, '1', 'EX', 5)
+                    return False
```
michalc: There is a bit of a race condition between the `GET` and `SET`... it allows a replay of a request after a `GET` but before the `SET`. To address this, you should be able to pass the `NX` option to `SET` and use the return value of `SET` to determine whether the key previously existed. There is similar a case at https://github.com/uktrade/activity-stream/blob/cf798a7ca7e69b11852c5371e89c1b8c5c6904ab/core/app/app_incoming_redis.py#L17

PR #151, opened 2019-08-20
Feature/159 track and report dataset downloads

niross:
[Trello card](https://trello.com/c/7r5QR5dT/159-track-and-report-dataset-downloads)

- Add new `EventLog` model
- Add new `Dataset` -> `SourceTable` download view
    - This necessitated moving  table download helpers (`table_data`, `table_exists`) to `core.utils` to allow for sharing across django apps 
- Create log entry on dataset/reference dataset download
- Add `EventLog` admin page with filters
- Allow for downloading of filtered `EventLog` records.


dataworkspace/dataworkspace/apps/catalogue/views.py
```diff
@@ -181,9 +202,46 @@ def get(self, request, *args, **kwargs):
         response['Content-Disposition'] = 'attachment; filename="{}"'.format(
             os.path.split(source_link.url)[-1]
         )
+
         return response
 
 
+class SourceTableDownloadView(DetailView):  # pylint: disable=too-many-ancestors
```
michalc: Extremely minor thing, but would be happy to disable too-many-ancestors for all code in the pre-commit config

niross: :+1:

dataworkspace/dataworkspace/apps/catalogue/views.py
```diff
@@ -146,8 +155,6 @@ def get_object(self, queryset=None):
             self.kwargs.get('group_slug'),
             self.kwargs.get('set_slug')
         )
-        if not dataset.user_has_access(self.request.user):
-            return HttpResponseForbidden()
 
         return get_object_or_404(
```
michalc: Just to check that it's desirable to expose 404s for items that don't exist in a dataset that the user doesn't have access to, (by exposing 403s for items that do exist in a dataset that the user doesn't have access to)

michalc: Ah... and below in SourceTableDownloadView 404s are only returned if the user has access to the dataset, so it's a bit inconsistent?

niross: Yep, the problem is I cant throw a forbidden error from `get_object()` so i've moved it into the `get()` method so we can check dataset permissions before querying for links/tables.

PR #160, opened 2019-08-24
(feat) Don't start the application from ajax requests

michalc:


dataworkspace/proxy.py
```diff
@@ -156,17 +156,20 @@ def get_peer_ip(request):
...
         if not host_exists:
-            async with client_session.request('PUT', host_api_url, headers=admin_headers(downstream_request)) as response:
-                host_exists = response.status == 200
-                application = await response.json()
+            if 'x-data-workspace-no-modify-application-instance' not in downstream_request.headers:
+                async with client_session.request('PUT', host_api_url, headers=admin_headers(downstream_request)) as response:
+                    host_exists = response.status == 200
+                    application = await response.json()
+            else:
+                raise UserException('Application stopped while starting', 500)
```
yusufertekin: Question: Is it worth to check application state to validate it's stopped while starting?

michalc: It does, a few lines above? Note that from the point of view of GET requests to the application api, an application only exists if its SPAWNING or RUNNING.

However, I do realise there is a bit of a race condition if for an application that isn't SPAWNING or RUNNING, a user loads two tabs at almost the same time for it. One of the tabs may show an error, because its request will try to start, via the PUT, an application already starting. However a) this is very unlikely, and b) if it happens the single_running_or_spawning_integrity column will prevent there actually being two identical starting applications

yusufertekin: thanks

PR #162, opened 2019-08-24
(refactor) Spawner specifies host pattern

michalc:
This is done as to work towards user-provided apps that will have a more complex hosts, in patterns such as `<gitrepo>-<gitbranch>-<userid>`


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -11,12 +12,32 @@
...
 logger = logging.getLogger('app')
 
 
+def application_template_and_data_from_host(public_host):
+    # Not efficient, but we don't expect many templates. At the time of writing,
+    # no more than 4 are planned
+    matching = [
+        (application_template, host_data)
+        for application_template in ApplicationTemplate.objects.all()
+        for host_data in [re.match('^' + application_template.host_pattern.replace('<user>', '(?P<user>.*?)') + '$', public_host)]
```
niross: Line length seems a bit long, do we have a standard for that here? pep8 says 80/100...

michalc: The dev team of the project typically decides, so happy to defer to you. When the dev team was just me, I set it to 250 in pre-commit, but can change it.

niross: Cool, just had a chat about it, could we make it 120?

michalc: Yes, will change

PR #169, opened 2019-08-27
(fix) Migration when there is existing data

michalc:
The default of '' with a unique index didn't work well with existing data.


michalc:
This doesn't behave as I thought it did

PR #172, opened 2019-08-29
Feature/235 link reference data tables

niross:
[Trello Card](https://trello.com/c/dEVjC2Qf/235-link-reference-data-tables)

- Add a display name flag to reference dataset field
- Add the ability to link a reference dataset field with an external dataset
- Add a selector to the reference dataset record form to link a record to an external dataset record
- On reference dataset download include linked dataset display name and identifier
- Validation
    - A reference dataset cannot link to itself
    - A reference dataset cannot link to a reference dataset that already links to it
    - A linked field cannot be an identifier
    - Linked field data type cannot be changed once created
    - Linked reference dataset cannot be changed if records with existing links exist
    - Ensure any linked to datasets are available on the external database for the linking dataset


dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -766,6 +847,14 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
         :return:
         """
         ref_dataset = self.reference_dataset
+
+        # Disallow linking of two reference datasets to one another
+        if self.data_type == self.DATA_TYPE_FOREIGN_KEY and \
+                self.linked_reference_dataset.fields.filter(linked_reference_dataset=self.reference_dataset).exists():
```
michalc: Does this prevent circular chains of 3 or more datasets?

dataworkspace/dataworkspace/apps/dw_admin/forms.py
```diff
@@ -14,27 +81,152 @@ class ReferenceDataFieldInlineForm(forms.ModelForm):
...
+                    )
+                )
+
+            # Ensure a linked to reference dataset doesn't link back to this dataset
+            if cleaned['linked_reference_dataset'].fields.filter(
+                    data_type=ReferenceDatasetField.DATA_TYPE_FOREIGN_KEY,
+                    linked_reference_dataset=self.reference_dataset).exists():
+                raise ValidationError(
+                    'Unable to link to a dataset that links to this dataset'
+                )
```
michalc: Same question: does this prevent all circular references? (Also... does this duplicate some functionality that's on the model?)

dataworkspace/dataworkspace/apps/datasets/models.py
niross: It didn't, it does now, good catch :+1:

dataworkspace/dataworkspace/apps/dw_admin/forms.py
niross: Fixed above but for the second point it does duplicate it but catching the validation error here means the user will see form errors on submission - otherwise it would be a 500 error

dataworkspace/dataworkspace/apps/datasets/model_utils.py
```diff
@@ -25,3 +25,23 @@ def clean_fields(fields):
...
+    """
+    links = [linked_dataset]
+    while links:
+        linked_dataset = links.pop()
+        if linked_dataset == target_dataset:
+            return True
+        links += set(
+            x.linked_reference_dataset for x in
+            linked_dataset.fields.exclude(linked_reference_dataset=None)
+        )
```
michalc: Is there some structure of links where this get stuck in an infinite loop? During each iteration at most one item is removed from `links`, but any number can be added... not completely sure: thinking about whether it's possible or not...

michalc: I'm suspecting it can only get into an infinite loop if there are pre-existing circular links in the data? So I guess it depends on whether the validation should be robust against that: say, some dodgy manual import caused it. (I would probably lean to yes, it should be robust against that...)

niross: Ok i've added a check to ensure links we've already checked are not checked again - I think this is what you are looking for?

michalc: Yes, I think so

PR #173, opened 2019-08-29
(feat) Page to show temporary S3 credentials

michalc:
This will be used in a later change to show a JavaScript-based S3 file explorer, which will use these credentials: so they won't need to be immediately visible in the page (although they will be visible if they view source).


dataworkspace/dataworkspace/apps/core/views.py
```diff
@@ -86,3 +93,25 @@ def table_data_view(request, database, schema, table):
         table_data(request.user.email, database, schema, table)
 
     return response
+
+
+def file_browser_html_view(request):
+    return \
+        file_browser_html_GET(request) if request.method == 'GET' else \
+        HttpResponse(status=405)
```
niross: Could we not just use a generic view here instead? Which should 405 if a post is request is received? Not that I have a problem with doing it this way. Just wondering what the difference is...

```
class FileBrowserView(View):
    def get(self, request):
        ...
        return render...
```

michalc: I tend to lean to having the response fairly explicit in terms of how it depends on the request. e.g. there is a touch more magic behaviour using `class FileBrowserView(View):`: it's not immediately apparent that anything but a GET would return a 405. And then, it's not clear what the body of the response would be in that case, etc.

Happy to use class-based if it's the team preference though?

niross: No preferences set in stone so happy to leave it :+1:

PR #175, opened 2019-09-02
Feature/201 create reference data loader

niross:
[Trello Card](https://trello.com/c/1dXsm0cr/201-create-reference-data-loader)

- Add new 'csv upload' page accessible from reference dataset admin detail page.
- Validate uploaded data using reference dataset record form field
- Add new log model which stores a record of each row uploaded and the status of the import (added, updated, validation failed)
- On successful upload show user the upload log so they are able to correct any issues.


dataworkspace/dataworkspace/apps/dw_admin/forms.py
```diff
@@ -227,6 +230,28 @@ def clean(self):
...
+        self.reference_dataset = kwargs.pop('reference_dataset')
+        super().__init__(*args, **kwargs)
+
+    def clean_file(self):
+        reader = csv.DictReader(chunk.decode() for chunk in self.cleaned_data['file'])
+        csv_fields = [x.lower() for x in reader.fieldnames]
+        dataset_fields = [x.lower() for x in self.reference_dataset.field_names]
+        if sorted(csv_fields) != sorted(dataset_fields):
+            raise ValidationError(
+                'Please ensure all reference dataset columns are included in the uploaded file'
```
yusufertekin: nit: error message might be misleading if csv file has typo on field name, or contains extra field.

niross: What would you suggest wording wise?

The main upload form has _some_ guidance "The uploaded csv file must contain headers matching columns listed below....<list of columns>"

dataworkspace/dataworkspace/apps/dw_admin/views.py
```diff
@@ -180,6 +193,108 @@ def get_success_url(self):
         )
 
 
+class ReferenceDatasetAdminUploadView(ReferenceDataRecordMixin, FormView):
+    template_name = 'admin/reference_dataset_upload_records.html'
+    form_class = ReferenceDataRecordUploadForm
+    object = None
```
yusufertekin: Sorry, I didn't understand the purpose of this 'object' named attribute ? It recurs on other views as well. But couldn't find any usage of it.

Note: name itself is a bit confusing due to it's a special name.

dataworkspace/dataworkspace/apps/dw_admin/views.py
```diff
@@ -180,6 +193,108 @@ def get_success_url(self):
...
+    upload_log = None
+
+    def get_template_names(self):
```
yusufertekin: nit: might want to use singular to be consistent with return value

dataworkspace/dataworkspace/apps/dw_admin/forms.py
yusufertekin: As I mentioned it's not much important. And Im not good with words but wanted to state what it checks and what it says are a bit different. I'd expect to see something like 
'Please ensure the uploded csv file headers are matching with target reference dataset columns'

niross: Cheers, i've updated it :+1:

dataworkspace/dataworkspace/apps/dw_admin/views.py
niross: Yeah you are right, i'm sure there was a reason for it at some stage - I have no idea what it was now though. Removed

PR #180, opened 2019-09-04
Feature/288 sort reference data set records and fields

niross:
[Trello Card](https://trello.com/c/j3UsMUWW/288-i-want-to-be-able-to-sort-the-order-of-both-reference-data-set-records-and-fields-attributes)

- Allow for sorting reference dataset fields from the reference dataset admin page
- Allow for selection of a sort field and direction for reference dataset records (falls back to creation date)


dataworkspace/dataworkspace/tests/test_views.py
```diff
@@ -258,6 +260,7 @@ def test_reference_dataset_csv_download(self):
             })
         )
         self.assertEqual(response.status_code, 200)
+        # raise Exception(response.content)
```
yusufertekin: probably forgotten

niross: Fixed (and the other one) thanks :)

yusufertekin:
LGTM! One small comment, It might be better if we squash migrations into single file for this PR but it's not blocker. 

How I'd do it. There might be better ways. 
I'd take a copy of reorder python function here dataworkspace/dataworkspace/apps/datasets/migrations/0013_auto_20190903_1142.py
Then delete all migrations introduced by this PR, rebase my branch against most updated master, 
then run make migrations again, and place reorder python operation in the right order to be executed.

PR #187, opened 2019-09-06
Add sentry logging if the SENTRY_DSN environment variable is set

niross:


dataworkspace/dataworkspace/settings/prod.py
```diff
@@ -24,3 +28,9 @@
 # Used when generating URLs for static files, and routed by nginx _before_
 # hitting proxy.py, so must not conflict with an analysis application
 STATIC_URL = '/__django_static/'
+
+if env.get('SENTRY_DSN') is not None:
+    sentry_sdk.init(
+        env['SENTRY_DSN'],
+        integrations=[DjangoIntegration()]
+    )
```
michalc: If this is to be dynamic based on the presence/absence of an environment, should it be in the base settings file?

Or...

If it's in the production settings file, I would lean to not having it dynamic, and failing if the environment variable is not set. It'll be a bit easy to change some settings and not notice that Sentry integration is broken for quite a while

niross: Ok i'll move it to base :+1:  The prod settings are used locally within docker containers (for me anyway) which is the reason for it being dynamic.

PR #190, opened 2019-09-06
(feat) Add files to tools

michalc:


michalc:
Will hold off from merging until get an ok from someone, since it's quite a visible change

PR #191, opened 2019-09-06
Feature/353 create dataset csv downloads with sql queries

niross:
- Allow for admin creation of downloadable custom sql reports
- Ensure all queries run via table/query download are readonly
- Order dataset links on the user facing dataset page by name
- Fixes bug on the event log when log record has no related object


dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -201,18 +201,11 @@ def write(self, value):
                 connect(database_dsn(settings.DATABASES_DATA[database])) as conn, \
                 conn.cursor(name='all_table_data') as cur:  # Named cursor => server-side cursor
 
+            conn.set_session(readonly=True)
```
michalc: I didn't know this was possible. I like it!

dataworkspace/dataworkspace/tests/test_views.py
```diff
@@ -528,3 +528,118 @@ def test_table_download(self):
...
+        self.assertEqual(EventLog.objects.count(), log_count)
+
+    @mock.patch('dataworkspace.apps.core.utils.logger')
+    def test_invalid_sql(self, mock_logger):
+        # Hide exception messages for this test only as its a bit confusing to see in the output
+        import gevent.hub
+        gevent.hub.Hub.NOT_ERROR = (Exception,)
+        query = self._create_query('SELECT * FROM table_that_does_not_exist;')
+        response = self._authenticated_get(query.get_absolute_url())
+        # Should return an empty csv
```
michalc: Good that it doesn't return data, but should it return something more like a 500? In the current form, would it result of a successful download of an empty file?

PR #202, opened 2019-09-13
(feat) Tools page with choice of container size

michalc:
Screenshot 2019-09-13 at 15 07 23


michalc:
Good to be reviewed, but will hold off until have confirmation that it's ok to deploy before mering

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -46,3 +56,68 @@ def application_spawning_html_GET(request, public_host):
...
+    }
+
+    def link(application_template):
+        public_host = application_template.host_pattern.replace('<user>', sso_id_hex_short)
+        # If there are some un-interpolated values, then we can't show a link to this: the app
+        # will be started by the user knowing the link ahead of time. Potentially in a future
+        # version there could be some UI to manage this.
+        return \
+            None if '<' in public_host or '>' in public_host else \
+            f'{request.scheme}://{public_host}.{settings.APPLICATION_ROOT_DOMAIN}/'
```
yusufertekin: Nit: Can we use parentheses for line continuation instead of '\'
like this or any other way. 
```
return (
    None if '<' in public_host or '>' in public_host else
    f'{request.scheme}://{public_host}.{settings.APPLICATION_ROOT_DOMAIN}/'
)
```

dataworkspace/dataworkspace/apps/catalogue/views.py
```diff
@@ -288,7 +286,6 @@ def get(self, request, *args, **kwargs):
 def root_view(request):
     return \
         root_view_GET(request) if request.method == 'GET' else \
-        root_view_POST(request) if request.method == 'POST' else \
         HttpResponse(status=405)
```
yusufertekin: Nit: Parentheses for line continuation instead of '\'

dataworkspace/dataworkspace/templates/tools.html
```diff
@@ -0,0 +1,112 @@
...
+                margin-bottom: 0;
+            }
+            .govuk-summary-list__key {
+                width: 20%;
+            }
+            .govuk-summary-list__actions {
+                width: 80%;
+            }
+        }
+    </style>
```
yusufertekin: Nit: It might be worth to separate styling from template.

dataworkspace/dataworkspace/apps/applications/views.py
michalc: Done

dataworkspace/dataworkspace/apps/catalogue/views.py
michalc: Done

dataworkspace/dataworkspace/templates/tools.html
michalc: I'm going to opt to leave this as-is, for several reasons:

- It's an early version of the page, and so expecting lots of changes; the CSS files have headers that I think cache them for a week; there is no automatic cache busting, which can be forgotten to do manually (e.g. by a manual file name change). These facts together I think mean there is a reasonable chance that the styles would get out of sync with the HTML, and result in a bad user experience and spurious bug reports.

- These styles would most likely not even result in an extra packet sent from the server to the client, so in any case any optimisation by having them in a separate CSS file would be extremely minor (possibly even negative if we have a separate CSS file just for this page). I would prioritise virtually all other considerations.

- This is very much "me" centred, but for me it's easier to have the styles close to the HTML they affect. It's really quick for me to just scroll up and down a few lines, and I never have to find another file. Given: it's probably me going to do the next few versions; but it's still very understandable for others if someone else will work on it; and easy for someone to move the styles to a separate file if they wish, I'm giving a little bit of weight to what's easier to me.

yusufertekin: Sounds reasonable. Thanks for the explanation.

PR #230, opened 2019-10-08
(feat) Dummy Google Data Studio data with basic custom connector

michalc:


dataworkspace/dataworkspace/apps/api_v1/views.py
```diff
@@ -159,3 +159,57 @@ def aws_credentials_api_GET(request):
...
+
+def table_api_schema_GET(request, _):
+    return JsonResponse({
+        'schema': DUMMY_SCHEMA
+    }, status=200)
+
+
+def table_api_rows_view(request, table_id):
+    return \
+        table_api_rows_GET(request, table_id) if request.method == 'GET' else \
```
lgarvey: If this is a common pattern it could be worth breaking request method checking into a decorator similar similar to django's `@require_http_methods`  or `@require_GET`

dataworkspace/proxy.py
```diff
@@ -335,6 +338,42 @@ def application_upstream(proxy_url, path):
 
         return downstream_response
 
+    def authenticate_by_staff_sso_token():
+
+        me_path = 'api/v1/user/me/'
+
+        @web.middleware
+        async def _authenticate_by_staff_sso_token(request, handler):
+            staff_sso_token_required = is_table_requested(request)
```
lgarvey: Wonder if it'd be worth just hitting the endpoint with or without a token and pushing the user through the auth flow if a 401/403 is returned.

michalc: > and pushing the user through the auth flow if a 401/403 is returned.

Ah in this case this isn't hit by a user in a browser, but by the Google Data Studio backend, which does its own redirection of the user to SSO to generate the token, that it then passes to the proxy here.

dataworkspace/dataworkspace/apps/api_v1/urls.py
```diff
@@ -6,12 +6,16 @@
...
+    table_api_rows_view,
 )
 
 
 urlpatterns = [
     path('application/<str:public_host>', csrf_exempt(login_required(application_api_view)),
          name='application-detail'),
     path('application', csrf_exempt(applications_api_view), name='application-list'),
-    path('aws_credentials', csrf_exempt(login_required(aws_credentials_api_view)), name='aws-credentials')
+    path('aws_credentials', csrf_exempt(login_required(aws_credentials_api_view)), name='aws-credentials'),
```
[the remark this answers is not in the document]
michalc: I think it's mostly habit, although I have a vague preference of wanting to see what protection(s) there are/are not on each URL at a glance in one (well, at least fewer) places. But then it's a bit inconsistent with other protections being in the views themselves...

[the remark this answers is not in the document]
michalc: Ah: I think also I like the explicitness of the function call rather than the decorator: every time I look at a decorator I convert it into the function call version in my head, so it's a touch more of a cognitive burden. But it's minor thing. Will ponder

dataworkspace/dataworkspace/apps/api_v1/views.py
michalc: Pondering this one... I guess I find it easier to reason about an `if` than the extra level(s) of indirection added by decorators, e.g. the function that is actually called as the view on the http request is the one returned by the function `require_GET`, which would be passed the one returned by `login_required`, which in turn would be passed by the one returned by `csrf_exempt`, which would be passed the one that's in the source code here. (Or whichever way round these work depending on their order...)

Plus, looking at https://docs.djangoproject.com/en/2.2/_modules/django/views/decorators/http/ , inside `require_http_methods` there is another decorator, `wraps`, which actually means there is another level of indirection, with (I believe) some code to make it look like there is one fewer levels of indirection. And `wraps` then uses partial function application https://github.com/python/cpython/blob/3.7/Lib/functools.py#L78 to do its thing, so there are even more functions being returned from functions.

All in all, I don't think all that code, especially with repeated use of higher level functions, is worth it, and I would argue that a general move to fewer decorators is better than a move to more, both for understandability, and (perhaps down the line) the ability to optimise performance.

PR #241, opened 2019-10-16
Joint datasets

dvoong:
Added the joint datasets section to the datagroup item view


dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -330,6 +330,7 @@ class ReferenceDataset(DeletableTimestampedUserModel):
         (SORT_DIR_ASC, 'Ascending'),
         (SORT_DIR_DESC, 'Descending'),
     )
+    is_joint_dataset = models.BooleanField(blank=True, null=True, default=False)
```
niross: Small thing but I don't think we need to set blank or null if we are setting a default here

dvoong: true, updated

dvoong: I've added the migration script now too

dataworkspace/dataworkspace/templates/datagroup.html
```diff
@@ -51,7 +51,30 @@ <h2 class="govuk-heading-s">
                     </h2>
                   {% endif %}
             {% endfor %}
-            {% if datasets and perms.datasets_dataset_add %}
+
+	    {% if joint_datasets or not datasets %}
+	      <h2 class="govuk-heading-l goveuk-!-margin-top-8">Joint datasets</h2>
```
niross: I think that should be govuk which will fix the top padding on this heading.

dvoong: whoops, fixed it

dvoong:
didn't mean to include that setup mac script into the PR, thanks for catching that

dvoong:
Gove is out! Same for the setup mac script

dataworkspace/dataworkspace/templates/datagroup.html
```diff
@@ -51,7 +51,30 @@ <h2 class="govuk-heading-s">
                     </h2>
                   {% endif %}
             {% endfor %}
-            {% if datasets and perms.datasets_dataset_add %}
+
+	    {% if joint_datasets or not datasets %}
+	      <h2 class="govuk-heading-l govuk-!-margin-top-8">Joint datasets</h2>
```
niross: One more small thing, could we just change 'Joint datasets' to 'Available joint datasets' just to match what we have for the other two sections.

niross: Actually apologies I've been given new information. Disregard this.

PR #260, opened 2019-10-28
Do not grant perms to tables that don't exist

niross:
Check that a table exists before granting permissions to it. Missing tables in the db were causing pgAdmin to fail to load.


dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -133,6 +133,14 @@ def get_new_credentials(database_obj, tables):
             ''')
 
             for schema, table in tables:
+                # Skip granting permissions if the table does not exist in the db
+                cur.execute(sql.SQL('SELECT to_regclass(%s)'), [table])
```
michalc: Should this use the schema name? In production right now, lots of tables are called `L1`

Note also: the schema names can have dots in them, so not sure how to_regclass handles that/what escaping it would need? e.g. `SELECT to_regclass('schema_name.table_name');` might not work if `schema_name` has dots?

It might be more verbose, but various other dynamic table/schema function query the pg_catalogue tables, so that is an option as well?

niross: Makes sense. I've switched to `pg_catalogue` as I couldn't find a nice way to use `to_regclass` as you mentioned :+1:

michalc:
I might be tempted to add a test for when starting an application when a dataset table doesn't exist but no biggie.

PR #262, opened 2019-10-29
spike for data-workspace api endpoints

dvoong:


dataworkspace/dataworkspace/apps/api_v1/datasets/tests/test.py
```diff
@@ -0,0 +1,14 @@
+from django.test import TestCase
```
allait: We've been keeping all tests in `dataworkspace/dataworkspace/tests` that parallels the structure of `apps/`, so could we move these to `dataworkspace/dataworkspace/tests/api_v1/datasets/test_views.py`?

dataworkspace/dataworkspace/apps/api_v1/datasets/tests/test.py
```diff
@@ -0,0 +1,14 @@
...
+from django.urls import resolve
+
+
+class TestFutureInterestCountriesDataset(TestCase):
+
+    url = '/api/v1/dataset/future-interest-countries-dataset'
```
allait: I think the general format for this would be `'/api/v1/dataset/<dataset-slug>`. Looking at our existing dataset URLs, I don't think they include `-dataset` at the end, so this one could be `future-interest-countries`.

dataworkspace/dataworkspace/apps/api_v1/datasets/company_future_interest_countries/views.py
```diff
@@ -0,0 +1,21 @@
...
+from rest_framework.pagination import CursorPagination
+from rest_framework.views import APIView
+
+
+class Pagination(CursorPagination):
+    ordering = ('id',)
+    offset_cutoff = None
+
+
+class CompanyFutureInterestCountriesDatasetView(APIView):
```
allait: Could we make a generic `DatasetView` endpoint that takes `dataset_slug` as an argument but returns 404 for anything that's not `future-interest-countries` for the time being? It shouldn't be that different from this but would be more straightforward to extend in the future.

dataworkspace/proxy.py
```diff
@@ -715,6 +733,63 @@ def authenticate_by_basic_auth():
...
+
+            # Read request content and store it in _read_bytes property to pass data to regarding endpoint after hawk
+            # authentication.
+            content = await request.read()
+            logger.info('Request method %s', request.method)
+            logger.info('URL HOST %s', request.url.host)
+            logger.info('URL Port %s', request.url.port)
+            logger.info('URL Path %s', request.url.path)
+            logger.info('Content %s', content)
+            logger.info('Headers %s', request.headers['Authorization'])
```
allait: I don't think we should log request contents or the auth header as they both could potentially contain sensitive data

dataworkspace/proxy.py
```diff
@@ -715,6 +733,55 @@ def authenticate_by_basic_auth():
...
+    def authenticate_by_hawk_auth():
+        async def lookup_credentials(sender_id):
+            for hawk_sender in hawk_senders:
+                if hawk_sender['id'] == sender_id:
+                    return hawk_sender
+
+        async def seen_nonce(nonce, sender_id):
+            nonce_key = f'nonce-{sender_id}-{nonce}'
+            with await redis_pool as conn:
+                nonce_stored = await conn.execute('GET', nonce_key)
```
dvoong: redis setnx

dataworkspace/proxy.py
```diff
@@ -715,6 +733,55 @@ def authenticate_by_basic_auth():
...
+                return await handler(request)
+
+            # Read request content and store it in _read_bytes property to pass data to regarding endpoint after hawk
+            # authentication.
+            content = await request.read()
+
+            is_authenticated, error_message, _ = await authenticate_hawk_header(
+                lookup_credentials,
+                seen_nonce,
+                1000,
```
dvoong: shorten to 15 seconds

dataworkspace/proxy.py
```diff
@@ -715,6 +738,54 @@ def authenticate_by_basic_auth():
...
+    def authenticate_by_hawk_auth():
+        async def lookup_credentials(sender_id):
+            for hawk_sender in hawk_senders:
+                if hawk_sender['id'] == sender_id:
+                    return hawk_sender
+
+        async def seen_nonce(nonce, sender_id):
+            nonce_key = f'nonce-{sender_id}-{nonce}'
+            with await redis_pool as conn:
+                response = await conn.execute('SET', nonce_key, '1', 'EX', 5, 'NX')
```
michalc: I think the nonce expiry should be longer: if the nonce expires in 5 seconds, the allowed timestamp skew being 15 seconds leaves 10 seconds during which the request can be resent. I would choose 30+ seconds

requirements.txt
```diff
@@ -13,3 +13,6 @@ zenpy==2.0.12
 requests==2.21.0
 django-admin-sortable2==0.7.3
 sentry-sdk==0.11.2
+djangorestframework==3.10.3
```
michalc: Are any features of djangorestframework that aren't also in regular views used? I would be tempted to remove this dependency if everything that can be achieved without this.

dataworkspace/proxy.py
```diff
@@ -423,12 +449,17 @@ def application_upstream(proxy_url, path):
...
             b''
             if 'content-length' not in upstream_headers
             and downstream_request.headers.get('transfer-encoding', '').lower()
             != 'chunked'
-            else downstream_request.content
+            else (
+                await downstream_request.read()
+                if downstream_request.content.at_eof()
+                else downstream_request.content
+            )
```
michalc: I might be tempted to disable Black for this, and have something like...
```python
# fmt: off
data = \
  b'' if 'content-length' not in upstream_headers and downstream_request.headers.get('transfer-encoding', '').lower() != 'chunked' else \
  await downstream_request.read() if downstream_request.content.at_eof() else \
  downstream_request.content
# fmt: on
```
(Not a blocker at all)

requirements.txt
```diff
@@ -13,3 +13,6 @@ zenpy==2.0.12
...
+mohawk==1.1.0
```
michalc: I think mohawk isn't needed in production: can mohawk go into requirements-dev.txt?

dataworkspace/dataworkspace/settings/base.py
```diff
@@ -202,3 +203,9 @@ def aws_fargate_private_ip():
 
 if env.get('SENTRY_DSN') is not None:
     sentry_sdk.init(env['SENTRY_DSN'], integrations=[DjangoIntegration()])
+
+# PAGINATION
+REST_FRAMEWORK = {
+    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.CursorPagination',
+    'PAGE_SIZE': 100,
```
michalc: Is this not used? Remove if not?

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
```diff
@@ -0,0 +1,152 @@
...
+    schema_value_func_for_data_types,
+)
+from dataworkspace.apps.datasets.models import SourceTable
+
+
+def get_streaming_http_response(request, source_table):
+    # validate arguments
+    try:
+        request_dict = json.loads(request.body)
+        fields = request_dict.pop('fields', [])
```
michalc: Does the API need dynamic fields?

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
```diff
@@ -0,0 +1,152 @@
...
+        yield_schema_and_rows_bytes(), content_type='application/json', status=200
+    )
+
+
+class APIDatasetView(APIView):
+    """
+    A GET API view to return the data for the company future countries of interest dataset
+    """
+
+    def post(self, request, dataset_id, source_table_id):
```
michalc: Would it be better to use a GET, especially if the API doesn't need to have dynamic fields?

michalc: ... and then, each page could contain the URL to the next page?

(Assuming we _do_ want it paginated? Is this still a desired feature, or should it all just be the one page? Is the client happy to have pagination?)

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
dvoong: not particularly, it defaults to all fields. I kind of took that from the google data studio api endpoint. You think it should be removed?

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
dvoong: this should be possible, just as long as the primary keys in the $searchAfter are appropriate for putting in the url get query which I think it probably is

dataworkspace/dataworkspace/settings/base.py
dvoong: I think this will eventually need to be in here, there will another api endpoint for reference data, reference data is a little different since it is managed by Django models so I think the approach for that data will be to use the django rest framework. But I'll remove it, and it can be added in a separate pull request

dataworkspace/proxy.py
dvoong: sure, I'll do that

dataworkspace/proxy.py
dvoong: will do that

requirements.txt
dvoong: will do

requirements.txt
dvoong: will do

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
dvoong: if we want to use pagination instead then we do not need to use a streaming http response and all the byte string chunking that comes with it? i.e. we would just need a sql query with a $searchAfter argument and a limit statement?

The pagination vs streaming response is a desired feature so the client (which is me in this case) can download large datasets. From my understanding we can either choose the pagination approach or the streaming response approach but both together don't make much sense

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
michalc: Yes: it sounds like it's not needed, so making the code more complex than it needs to be?

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
michalc: > ... both together don't make much sense

I think it all depends on what you want to/can do in the client...

- If you can handle / want 50mb pages (or some largish fixed number of bytes), then the current combo of streaming response and pagination is good
- If you can handle arbitrary sized pages in the client, then actually there is no need for pagination, just have one big streaming page. (You would have to have a client that can parse and process JSON in a streaming way)
- If you want small (e.g. < 1000 records) pages, then no need for streaming response, just a regular json_dumps page would be simplest/best.

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
```diff
@@ -0,0 +1,184 @@
...
+        select
+            *
+
+        from {schema}.{table}
+
+        {where_clause}
+
+        order by ({primary_key})
+
+        '''.format(
```
michalc: This looks like a string format, rather than a sql.SQL format.  Although it _might_ be safe against injections, it's tricky to know just by looking here. We're quite strict in other places to ensure that all dynamic sql doesn't allow an injection attack by using sql.SQL objects.

Essentially: I think only sql.SQL.format should be used to construct dynamic sql queries.

dvoong: fixed
https://github.com/uktrade/data-workspace/pull/262/files#diff-3d5956f3161509f3404d82e7da36b094R170

dataworkspace/dataworkspace/apps/api_v1/datasets/views.py
```diff
@@ -0,0 +1,200 @@
...
+                break
+        else:
+            yield from yield_chunks(b'], "next": null}')
+        yield from yield_remaining()
+
+    def yield_remaining():
+        if queue:
+            yield b''.join(queue)
+
+    # https://developers.google.com/apps-script/guides/services/quotas#current_limitations
```
michalc: Mini point: slightly misleading comment since this code isn't used by Google?

dataworkspace/proxy.py
```diff
@@ -715,6 +733,54 @@ def authenticate_by_basic_auth():
...
+
+        @web.middleware
+        async def _authenticate_by_hawk_auth(request, handler):
+            hawk_auth_required = is_hawk_auth_required(request)
+
+            if not hawk_auth_required:
+                return await handler(request)
+
+            # Read request content and store it in _read_bytes property to pass data to regarding endpoint after hawk
+            # authentication.
```
michalc: Mini point: now that .at_eof() is used, then no need to comment on _read_bytes? Better to comment that this makes .at_eof() return true?

PR #267, opened 2019-10-31
Feature/auto generate table names

dvoong:
thinking something like this might work @niross


niross:
@dvoong yep looks good :+1:

PR #269, opened 2019-11-06
Add static checks to CircleCI

allait:
### Persist local postgres data when containers are recreated

By default, docker creates anonymous volumes for the postgres container. This volume can be reused when `docker-compose up` is run, but would normally be lost on `docker-compose down` or if a container is rebuilt, taking all postgres data with it.

Creating a named volume makes sure the populated DB persists between image/container changes.

The data can still be reset with `docker-compose down -v`.

### Don't map host ports when running containers for test

`docker-compose run` won't apply published port mappings for the target service, but it still attempts to launch postgres and redis as is, which fails if there's another instance of either already running on the same default port.

We don't need to open ports for test containers to the host, so we can remove the mapping altogether, which still allows tests to talk to the DB but shouldn't conflict with any other running containers.

### Add Makefile targets to run tests using docker compose

Groups commands to run unit and application tests inside docker. Switches to use pytest for running both and sets a project name to avoid accidentally reusing dev DB/redis containers for tests.

### Add static analysis and autoformatting checks to CI

Similar to uktrade/data-flow#8 this adds a make target to check and autocorrect code formatting.

The steps run in CircleCI and can be run locally using `make check` and `make format`. They require requirements-dev.txt packages to be installed.

They checks aren't run inside the test docker container since it only contains dataworkspace/ code, which would exclude other python files from being checked.

This is also not adding mypy at the moment. Because our images are based on alpine mypy has to be built from source in the test container, which either complicates the build process or means we need to install it separately, without listing it in the dev requirements file.

mypy also currently flags a lot of missing annotations, which would need to be corrected first, so I'm leaving it for a separate task for now.

### Run `make format` on existing code


### Manually fix remaining flake8 issues


### Ignore flake8 warnings that are inconsistent with autoformatting

black and pycodestyle disagree on two rules: space after slices and line breaks before binary operators:

https://black.readthedocs.io/en/latest/the_black_code_style.html#line-breaks-binary-operator slice                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     s

In both cases black is supported by latest PEP8 recommendations, so 'm adding the flake8 error and warning to the ignore list.

### Move all test files to tests/

Keeping tests together outside of the main package simplifies a lot of tooling around testing (like making coverage reports not include test code itself) and keeps the test dependencies and imports separate from the main application package.

### Fix api_v1 test failing with a unique constrant error


michalc:
My only minor thing about the formatting is that, for me, ternary operators are less clear, e.g. 

https://github.com/uktrade/data-workspace/pull/269/files#diff-f5ea1cab9ff2485bce8b37b37d209bbbL301

allait:
@michalc yeah, I agree. I think those were reformatted because of the line length limit. One solution would be to rewrite them as a regular if/else or in this case use a dict to look up the correct function.

michalc:
> One solution would be to rewrite them as a regular if/else

I tend to avoid this to ensure that a variable is always set to never get a NameError...

> or in this case use a dict to look up the correct function.

... in this case, a dict seems very reasonable

PR #270, opened 2019-11-07
Add a Jenkins release promotion pipeline for data-workspace

allait:
Screenshot 2019-11-07 at 17 06 43

Adds a pipeline definition for running dev -> staging -> prod release
jobs with manual approval.

This ties together all recent releases to all data-workspace
environments within a single interface and would allow promoting
a specific release version between dev, staging and prod.

Approval steps take place within a "before agent" expression so that
they don't tie up an executor while waiting for user input. The inputs
are also surrounded by milestones so that previous releases are
automatically cancelled if a newer one proceeds to the next stage.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,65 @@
+pipeline {
+  agent none
+
+  stages {
+    stage('prepare') {
+      agent any
+      steps {
+        checkout scm
+        script {
+          pullRequestNumber = sh(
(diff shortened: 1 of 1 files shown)


allait:
@michalc as an option - since `data-workspace-<dev|staging|prod>` just call `ecs-pipeline` with some build parameters we could inline those parameters here and get rid of the individual jobs. It cleans up the jenkins jobs list a little bit and makes it easier to get to the actual deployment command logs, but on the other hand makes it a bit harder to run a hotfix deployment to a specific environment (as this would require setting all the arguments to ecs-pipeline directly).

PR #274, opened 2019-11-12
(feat) Metabase spike

michalc:


metabase/start.sh
```diff
@@ -0,0 +1,9 @@
+# To workaround https://github.com/metabase/metabase/issues/8373
+HOSTNAME=`cat /etc/hostname`
+echo "HOSTNAME: ${HOSTNAME}"
+echo "127.0.0.1 ${HOSTNAME}" >> /etc/hosts
```
allait: Should we do this in the Dockerfile instead?

michalc: I don't think we can... I think it's determined dynamically at runtime(?)

PR #277, opened 2019-11-12
(feat) Pgweb spike

michalc:


pgweb/pgweb-config.py
```diff
@@ -0,0 +1,85 @@
+import itertools
+import os
+import re
+
+
+def normalise_environment(key_values):
```
allait: Do we need this function here? We're only using `DATABASE_DSN` from the environment, or am I missing something?

pgweb/start.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -e
+
+python3 pgweb-config.py
```
allait: `export DATABASE_URL=$(python3 pgweb-config.py)` and skip writing the env file entirely?

pgweb/pgweb-config.py
michalc: Ah, DATABASE_DSN isn't the environment variable, it's `DATABASE_DSN_<some-name>`, with (in this context) `<some-name>` not being known

This is set at https://github.com/uktrade/data-workspace/blob/master/dataworkspace/dataworkspace/apps/applications/spawner.py#L170

(It is a bit overcomplex for the case of a single database, but is coded up for more than one...)

pgweb/start.sh
michalc: Sounds good!

PR #285, opened 2019-11-22
(feat) Workaround 50mb URLFetch limit in Google Data Studio

michalc:


dataworkspace/dataworkspace/apps/api_v1/views.py
```diff
@@ -450,31 +424,145 @@ def table_api_rows_POST(request, table_id):
...
+    def query_vars_search_after(fields_sql, schema_sql, table_sql, primary_key_sql):
+        search_after = request_dict['$searchAfter']
+
+        return (
+            sql.SQL(
+                'SELECT {},{} FROM {}.{} WHERE ({}) > ('
+                + ','.join(['%s'] * len(search_after))
+                + ') ORDER BY {}'
+            ).format(
+                primary_key_sql,
```
allait: Could we quote/escape these variables here? From looking at the code I believe they're not external, so we're not at risk of injection, but it'd be easier to audit if they were escaped when they're combined with the query. A test could also be useful.

michalc: Ah this is escaped: note the `format` is not a string format, it's a `sql.SQL` format, which does the escaping.

Are you you looking for a test that ensures its escaped? (There is a bit of a test that runs this code, by passing `$searchAfter` into the endpoint)

michalc:
@allait 

> Is it called by GDS backend? 

Yes

> How is the request authenticated

In https://github.com/uktrade/data-workspace/blob/master/dataworkspace/proxy.py#L457 the proxy checks that the token is valid against SSO

> and how are credentials passed to GDS?

GDS behaves like a typical SSO application, in that...

- A user visits it...
- ... it will redirect to SSO (the user will login if necessary) and it redirect back to a URL
- ... and the url with have a code in as a query string parameter, with which GDS requests a token from SSO
- ... and this is the token passed with requests and checked by proxy.py

Typically, the user would already logged into SSO since they've just come from Data Workspace, so often the above process is quick/unnoticeable.

> could you describe a bit how the integration between Google Data Studio and this API endpoint works? 

A typical flow
- A user will click on a Google Data studio link next to a table in Data Workspace, and follows the instructions to add it as a data source in a report
- This link has the ID of a table in, and GDS uses this to query its two endpoints, `rows` and `schema`
- The schema endpoint converts the metadata of the columns in Postgres to the JSON format that GDS is expecting
- When querying for rows, GDS also can request only a subset of the columns (and no other columns can be returned)
- The rows endpoint also must return the schema of the requested columns
- All results are ordered by Postgres primary key of the table to make pagination work consistently
- Slightly unfortunately, there are two pagination mechanisms
   1. When initially adding the data source in GDS, it requests 100 rows to try and guess the schema of data (even though we tell it the schema...)
   2. The 50mb pagination added here

dataworkspace/dataworkspace/apps/api_v1/views.py
allait: Ah, I missed that! Yeah, that's fine then 👍

PR #287, opened 2019-11-26
Add published version to reference datasets

marcuspp:
This change allows unpublished datasets to consolidate changes in to one major or minor update.

It does this by using a new field called published_version which whilst published mirrors that of version.
Upon being unpublished and changes made the numbers diverge (version number never being surfaced in the admin to avoid confusion). On republishing the numbers are checked and the next major or minor version is then set to both variables.

To test: 
 - Create a reference dataset 
 - Make changes add fields add data
 - Check the version number is yet to be set
 - Publish
 - Check version number now 1.0
 - Make changes
 - Check the version number follows current behaviour
 - Unpublish
 - Make multiple updates
 - Check version number does not change
 - Publish
 - Check version number is incremented only once (either major or minor depending on the changes)

Comments:
 - Unsure if the html templates should be updated to use published_version property rather than the version
 - Unsure if the approach taken to what data is shown in the admin is correct. Should it show version also so the user is aware changes have been made or is displaying both confusing.
- Migration sets defaults on fields which can cause db locking on large tables. This table has a hand full of rows but should it be split up anyway?

Screenshot 2019-11-26 at 11 30 44


niross:
In response to your comments. Which I forgot about :man_facepalming: 

> Unsure if the html templates should be updated to use published_version property rather than the version

I would say yes, they should be using `published_version` - might want to check with pm/dm though.

> Unsure if the approach taken to what data is shown in the admin is correct. Should it show version also so the user is aware changes have been made or is displaying both confusing.

Again maybe check with Adrian/Lucas but I don't think there is any reason for the "non published" version numbers to be shown anymore.

> Migration sets defaults on fields which can cause db locking on large tables. This table has a hand full of rows but should it be split up anyway?

I think it will be ok as it is. I can do a release out of hours if you are worried.

marcuspp:
Thanks @niross I have updated the FE template and left all admin as is after discussing things with the DM/PM

PR #369, opened 2020-01-07
Feature/external reference datasets

niross:
### Description of change

- Give users permissions to access reference dataset tables via tools and visualisations
- Fixes for various issues encountered when making changes to external reference datasets

### Checklist


niross:
> Looks good 
> 
> Is it worth having regression tests for the fixed bugs?

Cheers, yes it totally is. I haven't been able to recreate the issue anywhere but staging though. I will have another crack at it anyway.

PR #373, opened 2020-01-08
Add a datasets search page with filters

allait:
### Description of change

Screen Shot 2020-01-08 at 16 03 13-fullpage

### Don't duplicate query arguments when proxying admin requests

When sending the request to the admin app proxy passed the query parameters as a separate `query` argument to `handle_http`, as well as adding them to the `upstream_url`. This results in each query value being duplicated in the Django `QueryDict`.

Removing query string from the URL lets `handle_http` only leaves one copy of the parameters provided in `query`.

### Add a Django forms widget for a Design System checkboxes component

We need to render the search filters as a collection of checkboxes. Normally we'd add the HTML directly to the template, but in this case it involves a lot of templating logic for setting correct checkbox ids, labels and setting the checked value based on the provided form data.

Django forms already handles a lot of this, but the default templates for CheckboxSelectMultiple have a different layout than the GOV.UK Design System component, so we need to write our own custom widget and related templates.

Apart from custom classes and div structure, one big difference in Django and GOV.UK form layout if that Django considers the form field label to be outside the input template. The label is rendered when the whole form is rendered, but not when an individual field is (for example using `{{ form.field }}`). In Design System however the field heading is inside the field's div and fieldset tags, which should be rendered by `{{ form.field }}`.

One way around this would be to use grouped lists of `choices` and render the name of the group as the heading for the set of checkbox inputs. The problem is that the grouped choices aren't supported out of the box by `ModelChoiceField`, so we'd need to write additional logic to group queryset results.

Instead, we add a new `group_label` argument to the widget constructor, which should contain the desired heading for the field and is passed to the template context.

### Add a datasets search page with filters

Adds a new /datasets/ page with search and filters across datasets, data cuts and reference datasets.

Search and filters are validated by a Django form and applied to a dataset and (if required) reference dataset queries. The queries are converted to dictionaries of common fields, which makes it possible to combine and sort them with a `.union`, allowing us to search across two separate models.

The `use` filter controls which types of datasets should be included, with a special "0" value for reference datasets (to keep the same type as the DataSet.type while avoiding clashing with potential future types).

This change also includes CSS rules for the search field, which currently doesn't have a component in the GOV.UK Design System (https://github.com/alphagov/govuk-design-system-backlog/issues/132).

The rules are based on the GOV.UK search box.

Things that are currently missing:

* Individual pagination links, currently there's only `Next` and `Previous`
* Refreshing the results when a filter is selected/deselected without clicking the search button
* Search box default text label, which requires JavaScript to hide the text if there's user input in the field

### Hide the new search page behind a feature flag

The search page is still in development, so we're placing it behind a `datasets-search` feature flag, which should only be enabled for superusers/staff in production.

### Checklist

* [X] Has the [CHANGELOG](https://github.com/uktrade/data-workspace/blob/master/CHANGELOG.md) been updated?


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -174,3 +182,62 @@ def get_template_names(self):
...
+@waffle_flag('datasets-search')
+def find_datasets(request):
+    form = DatasetSearchForm(request.GET)
+
+    if form.is_valid():
+        query = form.cleaned_data.get("q")
+        use = form.cleaned_data.get("use")
+        source = form.cleaned_data.get("source")
+
+    datasets = filter_datasets(DataSet.objects, query, source, use)
```
michalc: If the form isn't valid for whatever reason, query, source and use won't be defined, and a NameError would be raised here?

dataworkspace/dataworkspace/templates/datasets/index.html
```diff
@@ -0,0 +1,68 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Welcome to Data Workspace</h1>
+    <p class="govuk-body-l">
+      Data Workspace is where DIT staff and partners can access, analyse and securely share data relevant to DIT.
+    </p>
+  </div>
+
+    <div class="govuk-grid-column-two-thirds">
+      <div class="search-field govuk-!-margin-bottom-9">
+        <!-- <label class="govuk-label search-field__label" for="search">Search data</label> -->
```
michalc: Should this be removed?

dataworkspace/dataworkspace/apps/datasets/views.py
allait: Yes, but since the form doesn't have any required fields this only happens if the user modifies the query string by hand. I've added a redirect back to the search page here just so we don't get a 5xx in Sentry.

dataworkspace/dataworkspace/templates/datasets/index.html
allait: I'm going to bring it back once the JS for hiding it after input is added.

dataworkspace/dataworkspace/apps/datasets/views.py
michalc: ok!

dataworkspace/dataworkspace/templates/datasets/index.html
michalc: ok

dataworkspace/dataworkspace/static/data-workspace.css
```diff
@@ -40,3 +40,110 @@
...
+
+.search-field__submit {
+ border:0;
+ cursor:pointer;
+ border-radius:0;
+ position:relative;
+ padding:0;
+ width:40px;
+ height:40px;
+ background-image:url('/__django_static/assets/images/search-button.png');
```
michalc: The files browser uses FontAwesome icons I think. Just to check if we should be continuing to use those for consistency? [And if licensing is all good etc..., both for the FA icons, and for this one?]

dataworkspace/dataworkspace/templates/datasets/index.html
```diff
@@ -0,0 +1,68 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-third">
+    {{ form.use }}
+    {{ form.source }}
+  </div>
+  <div class="govuk-grid-column-two-thirds">
+    {% for dataset in datasets %}
+    <div class="search-result">
+      <h2 class="govuk-heading-m">
+        <a class="govuk-link" href="{{ dataset.get_absolute_url }}">{{ dataset.name }}</a>
```
niross: I don't think `get_absolute_url` is working here. It's an empty string on my dev env anyway.

michalc:
> Search box default text label, which requires JavaScript to hide the text if there's user input in the field

Just to check, would the placeholder attribute be enough? (Also: even if it's not quite perfect, should making it fancy with JS be left until later?)

dataworkspace/dataworkspace/static/data-workspace.css
```diff
@@ -40,3 +40,110 @@
...
+.js-enabled .search-field__label {
+ position:absolute;
+ left:15px;
+ top:1px;
+ z-index:1;
+ color:#6f777b
+}
+
+.search-field-wrapper {
+ display:table;
```
michalc: The file browser uses flexbox to position things. Admittedly older browsers won't support this, but for an internal product where all(?) users will be on managed devices (/ at the very least, should _not_ be using older browsers on data workspace), I think this would be acceptable. Could do this as well for consistency?

dataworkspace/dataworkspace/templates/datasets/index.html
```diff
@@ -0,0 +1,68 @@
...
+    </p>
+  </div>
+
+    <div class="govuk-grid-column-two-thirds">
+      <div class="search-field govuk-!-margin-bottom-9">
+        <!-- <label class="govuk-label search-field__label" for="search">Search data</label> -->
+        <div class="search-field-wrapper">
+          <input type="search" name="q" id="search" title="Search" class="govuk-input search-field__item search-field__input" value="{{ query }}" aria-controls="">
+          <div class="search-field-submit-wrapper search-field__item">
+            <input class="search-field__submit" type="submit" value="Search">
```
michalc: Should a button be used for ease of styling/ability to put HTML/images inside it, and for consistency with the file browser?

dataworkspace/dataworkspace/templates/datasets/index.html
allait: Oops, nice catch. This was working before I combined the queries with union, but now that they return a list of dictionaries we obviously don't have access to the DataSet methods. I've rebuild the URL in template and removed the admin "Edit" for now (it requires more information to rebuild and is available from the details page).

dataworkspace/dataworkspace/static/data-workspace.css
allait: I don't know where the icon came from originally, but since it's published under Crown Copyright we should have permission to use it and I know other projects in GDS do.

allait:
@michal there's no placeholder text at the moment (which is why <label> is commented out). The JS is needed exactly for the reason you highlight in the GOV.UK screenshot, so I'll add the label together with JS.

I believe `placeholder` attribute has some accessibility issues (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input - "Placeholders are not accessible"), which is why GOV.UK is using a label. 

In general this is mostly taken from GOV.UK publishing as-is, and since it's pretty self-contained I'd prefer not change it too much unless we have a specific reason to. There are discussions on the Design System about a search field component, I'd expect us to switch to one once it becomes available, so hopefully the CSS isn't here for long.

PR #374, opened 2020-01-08
(fix) Duplicate query string params being received

michalc:
### Description of change

Fix duplicate query string parameters being received by the admin application. Usually, this doesn't make a difference. However, on endpoint users `getlist`, which is sensitive to duplicate parameters, and its behaviour breaks

### Checklist


dvoong:
I have a good feeling about this

PR #414, opened 2020-02-04
Migrate docker debian

ejayesh:
### Description of change
Migrated R and Python docker images from Alpine to Debian.

### Checklist

* [ x] Have tests been added to cover any changes?
* [ x] Has the [CHANGELOG](https://github.com/uktrade/data-workspace/blob/master/CHANGELOG.md) been updated?
* [ x] Has the README been updated (if needed)?

### DIFF
--- a/jupyterlab-python/Dockerfile
+++ b/jupyterlab-python/Dockerfile
@@ -1,5 +1,29 @@
-FROM alpine:3.8
+FROM debian:buster-slim
 
+# Must be the same as the s3sync user in the s3sync container so it
+# can modify files
+RUN \
+    groupadd -g 4356 jovyan && \
+    useradd -u 4357 jovyan -g jovyan -m
+
+RUN \
+    apt-get update && \
+        apt-get install -y --no-install-recommends \
+        autoconf=2.69-11 \
+        build-essential=12.6 \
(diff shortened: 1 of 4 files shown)


jupyterlab-python/Dockerfile
```diff
@@ -1,5 +1,29 @@
...
+        git=1:2.20.1-2+deb10u1 \
+        openssh-client=1:7.9p1-10+deb10u1 \
+        texlive-xetex=2018.20190227-2 \
+        texlive-generic-extra=2018.20190227-2 \
+        texlive-fonts-recommended=2018.20190227-2 \
+        wget=1.20.1-1.1 \
+        ca-certificates
+
+# Set the locale
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y locales && \
```
allait: Do we need a new RUN command here or can this be chained to the previous one to keep the number of layers low?

michalc: Ah good spot. We can and will chain. However, treating this as an optimisation, and I believe @ejayesh will do this tomorrow in another PR. (From testing, with this image load times seem to be very almost the same as they are currently)

PR #418, opened 2020-02-05
cleanup docker files

ejayesh:
### Description of change
Reduce unnecessary RUNs in Docker image

### Checklist

* [x] Has the README been updated (if needed)?

### DIFF
--- a/jupyterlab-python/Dockerfile
+++ b/jupyterlab-python/Dockerfile
@@ -11,6 +11,7 @@ RUN \
           texlive-generic-extra=2018.20190227-2 \
           texlive-fonts-recommended=2018.20190227-2 \
           wget=1.20.1-1.1 \
+          sudo=1.8.27-1+deb10u1 \
           ca-certificates && \
       echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen && \
       locale-gen en_US.utf8  && \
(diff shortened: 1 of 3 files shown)


jupyterlab-python/Dockerfile
```diff
@@ -102,9 +98,14 @@ COPY jupyter_notebook_config.py /etc/jupyter/jupyter_notebook_config.py
 
 RUN \
     echo '[global]' > /etc/pip.conf && \
-    echo 'index-url = https://s3-eu-west-2.amazonaws.com/mirrors.notebook.uktrade.io/pypi/' >> /etc/pip.conf
-
-EXPOSE 8888
+    echo 'index-url = https://s3-eu-west-2.amazonaws.com/mirrors.notebook.uktrade.io/pypi/' >> /etc/pip.conf && \
+    apt-get remove --purge -y \
+      wget && \
```
michalc: I think this remove/clean won't reduce the size of the image, since it's not in the same later as things getting installed?

michalc: (And ditto in the other Dockerfile?)

michalc: (Ultimately as well... I think we can cope with wget left on the image... probably makes virtually no difference... I may have been micro/prematurely optimising before..._

ejayesh: Agreed, the purpose of the PR was also to re-order and remove unnecessary steps.

ejayesh: Removing wget saves 20Mb :-)

jupyterlab-r/Dockerfile
```diff
@@ -101,24 +101,28 @@ RUN \
...
-    "--NotebookApp.ip='0.0.0.0'", \
-    "--NotebookApp.allow_remote_access=True", \
-    "--NotebookApp.port=8888"]
+#CMD ["/opt/conda/bin/jupyter", \
+#    "lab", \
+#    "--config=/etc/jupyter/jupyter_notebook_config.py", \
+#    "--NotebookApp.token=''", \
+#    "--NotebookApp.ip='0.0.0.0'", \
+#    "--NotebookApp.allow_remote_access=True", \
+#    "--NotebookApp.port=8888"]
```
michalc: Should be uncommented?

ejayesh: fixed.

PR #430, opened 2020-02-10
Dataset admin accessibility

niross:
### Description of change

**Admin dataset access**
- View an unpublished dataset by visiting the url directly (unpublished datasets do not appear in search results)
- Access related source tables for an unpublished dataset from within tools

**Dataset soft delete**
- Deleting a dataset via the admin sets the `deleted` flag rather than deleting from the db entirely
- Deleted datasets are hidden from search results and admin listings

**Misc**
- Switch some tests to pytest from django test cases - This allows for parameterizing? user/staff clients 

### Checklist


dataworkspace/dataworkspace/templates/datasets/referencedataset_detail.html
```diff
@@ -24,6 +24,9 @@
 {% endblock %}
 
 {% block inner_content %}
+  {% if not model.published %}
```
allait: This should be `object.published`, otherwise banner shows up all the time. Or even better we could rename `object` to `model` in this template so we don't keep running into this.

niross: Good catch :+1 i've updated it to bring it inline with the other dataset templates

PR #434, opened 2020-02-13
Add spinner to 'tool loading' page

samuelhwilliams:
### Description of change

We find that users have low levels of confidence that their tool is
loading properly when everything on the page is static. By adding a
dynamic spinner we hope to increase confidence that 'work is being
done'.

This solution is lifted from conversation on the GOV.UK Design System
repo: https://github.com/alphagov/govuk-design-system-backlog/issues/28

Ticket: https://trello.com/c/ODpRBsdj/828

## Show the thing
It looks a bit jarring in the gif, but it moves smoothly in reality...

spinner-green

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 
+## 2020-02-12
+
+## Added
+
+- A loading spinner on the 'tool spawning' page.
+
+
 ## 2020-02-11
 
 ### Changed
--- a/dataworkspace/dataworkspace/static/data-workspace.css
+++ b/dataworkspace/dataworkspace/static/data-workspace.css
@@ -156,9 +156,9 @@ Comment: https://github.com/alphagov/govuk-design-system-backlog/issues/28#issue
 .loading-spinner {
   border: 12px solid #ffffff;
   border-radius: 50%;
-  border-top-color: #1d70b8;
-  width: 80px;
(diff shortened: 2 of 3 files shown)


samuelhwilliams:
Eventually (😬) managed to get this tested via BrowserStack for core supported devices (as per https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices). It looks OK on all of them.

michalc:
I would love for the countdown to be restored, to avoid the spinner. Or if not a countdown, some sort of progress bar, maybe even hardcoded to 2 mins to get all the way across. This could be the first stage of a way of hooking into the various stages of loading an application (and displaying them maybe), or maybe using historical loading times to inform how long it takes.

Note, for visualisations... a) they tend to load up quicker than 2 mins, and b) a person may see the loading screen after someone else was started the visualisation. e.g. see the screen for the first time, it says 2 minutes approximately when there are 5 seconds to go.

(Or even if you're the same person that started the loading... you can close the browser, do something else for a minute, and then come back... it doesn't really have an indication of how much is left)

samuelhwilliams:
Updated the colours to be more consistent with the rest of the page as per feedback from Berkan.

@michalc I think we're going for an untimed implementation at the moment, but obviously we'll see how well this works to reassure users. Having a timed solution feels like more effort, and if we avoid the effort by hard-coding e.g. a 2 minute length, then that also comes with its own set of problems (weirdness around things finishing "early" or still needing work after 2 minutes and users losing confidence again). Definitely is good input though, and something we can continue to keep in mind as we look at maturing this page.

PR #435, opened 2020-02-13
(feat) Allow the port of an application to be overridden

michalc:
### Description of change

Allow the port of an application to be overridden, and open 9000. This means that, for example, in JupyterLab, a user can start their own webserver on port 9000 and develop on it, opening up

jupyterlab-[sso-id]--9000.domain.com

to view their in-progress application.

### Checklist

### DIFF
--- a/dataworkspace/proxy.py
+++ b/dataworkspace/proxy.py
@@ -532,7 +532,10 @@ async def async_main():
 
         # fmt: off
         data = \
-            b'' if 'content-length' not in upstream_headers and downstream_request.headers.get('transfer-encoding', '').lower() != 'chunked' else \
+            b'' if (
+                'content-length' not in upstream_headers
+                and downstream_request.headers.get('transfer-encoding', '').lower() != 'chunked'
+            ) else \
             await downstream_request.read() if downstream_request.content.at_eof() else \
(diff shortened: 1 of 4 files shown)


dataworkspace/proxy.py
```diff
@@ -268,6 +268,11 @@ def get_peer_ip(request):
         public_host, _, _ = downstream_request.url.host.partition(
             f'.{root_domain_no_port}'
         )
+        public_host, _, port_override_str = public_host.partition('--')
```
allait: Apart from it making partition straightforward here, is there another reason to use `--` instead of `-`? Things like Slack for example love replacing it with an em dash, although not in a middle of the word, so maybe that's not a real concern.

michalc: It is to make the partition straightforward, including for the cases when there is no `--9000`, but also to give a little bit more of a visual difference between the part of the domain that determines which task to route to, and which is the port on that task.

If the em issue ends up too annoying, I suspect we can change it though. For one thing, I am hoping that it won't have to be typed in manually, e.g. can add a button in JupyterLab to launch a new tab with it automatically populated, so if it needs to be changed we can just change the code in the button.

PR #437, opened 2020-02-14
Build dataworkspace docker container in Jenkins

allait:
### Build dataworkspace docker container in Jenkins

Previously, our images were build by quay.io using a Github trigger and we needed to wait for the build to finish before triggering a release job in Jenkins or it would pick up an old master image tag.

This adds a step to the Jenkins pipeline that builds a docker image for dataworkspace admin and pushes it to quay before proceeding with the release to dev. This way, we can kick off the release pipeline automatically after merge and the correct image, tagged with commit SHA will get released. This also allows us to resume older pipelines paused before staging/prod since they'll be tied to a unique immutable tag.

This merges existing prepare step and a k8s agent build step into one, since both require a workspace checkout. The agent is defined within the stage, allowing pipeline itself to use `agent none` and avoid blocking an executor while waiting for manual approval.

The build step is locked, preventing simultaneous builds, but making it safe to merge things in quick succession without causing a release race.

### Run ecs-pipeline jobs directly

This removes the intermediate data-workspace-{env} jobs and runs ecs-pipeline with required parameters directly.

This reduces the number of simultaneous jobs running in Jenkins during release and makes it easier to get to the logs in case of failure.

One of the benefits of a single release pipeline is that it shows new commits since last job run, making it easier to tell what changes are about to go live. This benefit is reduced if some releases are going through separate jobs, so the idea is to remove individual env release jobs once this is merged. In case we ever need a quick hotfix release to prod while dev/stage pipeline is blocked/broken we can always use ecs-pipeline job directly.

In addition, this adds the locks around release jobs making sure simultaneous merges don't overtake each other between stages.

### Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,17 +1,67 @@
 pipeline {
   agent none
 
+  parameters {
+    string(name: 'GIT_COMMIT', defaultValue: 'master', description: 'Commit SHA or origin branch to deploy')
+  }
+
   stages {
-    stage('prepare') {
-      agent any
+    stage('build') {
+      agent {
+        kubernetes {
+        defaultContainer 'jnlp'
+        yaml """
+            apiVersion: v1
+            kind: Pod
+            metadata:
+              labels:
+                job: ${env.JOB_NAME}
+                job_id: ${env.BUILD_NUMBER}
+            spec:
+              nodeSelector:
+                role: worker
+              containers:
+              - name: builder
+                image: gcr.io/kaniko-project/executor:debug
+                imagePullPolicy: Always
+                command:
+                - cat
+                tty: true
+                volumeMounts:
+                - name: jenkins-docker-cfg
+                  mountPath: /kaniko/.docker
+              volumes:
+              - name: jenkins-docker-cfg
+                configMap:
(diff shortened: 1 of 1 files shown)


allait:
@michalc if this gets approved/merged, we probably need to disable the Git hook in quay.io itself, so it doesn't override Jenkins builds for branches. We could keep it for `master`, since we'll be tagging main releases with a SHA.

Jenkinsfile
```diff
@@ -1,27 +1,75 @@
...
+                    path: config.json
+        """
+        }
+      }
       steps {
-        checkout scm
+        checkout([
+            $class: 'GitSCM',
+            branches: [[name: params.GIT_COMMIT]],
+            userRemoteConfigs: [[url: 'https://github.com/uktrade/data-workspace.git']]
```
allait: This is weirdly required because without it `master` doesn't work, but `origin/master` does, but the later can't be used as an image tag since it gets replaced by SHA, but passed as a string to release jobs 🤷‍♂

michalc:
@allait Let me know just before merge, and I'll disable the quay.io build

PR #453, opened 2020-02-21
(feat) Visualisations page

michalc:
### Description of change

- Fetches a list of Visualisation projects from GitLab on a new page
- This page is accessible to superusers and users who have the develop_visualisations permission
- There is a new top level tab, visible only to those how have develop_visualisations explicity, so superusers can maintain the view of most users as they wish.

This is is the first of a sequence of changes on this feature. Right now this isn't that useful, but wanting feedback from production, e.g. for speed of page loads that depend on synchronous responses from GitLab, and feedback from other developers on permissions/code etc. Am hoping we can live without having to do some ajax-y requests from the browser, and without server-side caching, but we will see.

Also: not really looking for feedback on the very minimal/rough UI at this point, this PR is for permissions & wiring.

Screenshot 2020-02-24 at 12 19 50

### Checklist


dataworkspace/dataworkspace/apps/applications/urls.py
```diff
@@ -4,9 +4,11 @@
 from dataworkspace.apps.applications.views import (
     application_spawning_html_view,
     tools_html_view,
+    visualisations_html_view,
 )
 
 urlpatterns = [
     path('', login_required(tools_html_view), name='tools'),
+    path('', login_required(visualisations_html_view), name='visualisations'),
```
allait: I don't think this should be here

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -148,3 +150,22 @@ def tools_html_POST(request):
             request, 'Stopped ' + application_instance.application_template.nice_name
         )
     return redirect(redirect_target)
+
+
+def visualisations_html_view(request):
+    return (
```
allait: I honestly don't feel like the saved 12 characters on two `return` statements are worth the reversed logic layout caused by double ternary compared to:

```python
if not request.user.has_perm('applications.develop_visualisations'):
    return HttpResponseForbidden()

if not request.method == 'GET':
    
return visualisations_html_GET(request)
```

At which point you could probably inline the `_GET` function without losing any readability since the edge conditions are handled at the beginning and don't require nesting the rest of the code.

dataworkspace/dataworkspace/context_processors.py
```diff
@@ -1,8 +1,27 @@
...
+    # during early development, we hide the tab itself from the interface
+    # unless the user has been given the explicit permission to develop
+    # visualisations. This is so until the interface is a bit more final, most
+    # superusers still see the application as most users will, which is good
+    # for demos, but can optionally see the WIP interface as needed.
+    #
+    # This I suspect is _not_ cached, and so results in a database query on
+    # every page load. This is slightly unfortunate, but should be acceptable
+    # for an easy query.
+    can_see_visualisations_tab = request.user.user_permissions.filter(
```
allait: Is this different from using `{% if perms.applications.develop_visualisations %}` in the template?

dataworkspace/dataworkspace/templates/_base.html
```diff
@@ -63,6 +63,14 @@
             </a>
           </li>
           {% endif %}
+          {% if can_see_visualisations_tab %}
+          {% url 'visualisations:root' as visualisations_root_url %}
+          <li class="govuk-header__navigation-item{% if request.get_full_path == visualisations_root_url %} govuk-header__navigation-item--active{% endif %}">
+            <a class="govuk-header__link" href="{{ visualisations_root_url }}">
```
allait: There's another set of links in the footer, duplicating the header ones. We should probably add it there as well.

michalc: Will do

dataworkspace/dataworkspace/context_processors.py
michalc: Yes, I believe so. I don't want superusers to see to the visualisation tab unless that have the explicit permission. However, I think perms.applications.develop_visualisations would return true for superusers even if they don't have the permission.

dataworkspace/dataworkspace/apps/applications/views.py
michalc: Can do this

dataworkspace/dataworkspace/apps/applications/urls.py
michalc: Oh yes, will remove

dataworkspace/dataworkspace/templates/_base.html
michalc: Done

dataworkspace/dataworkspace/apps/applications/urls.py
michalc: Removed

dataworkspace/dataworkspace/apps/applications/views.py
michalc: Done

dataworkspace/dataworkspace/context_processors.py
allait: Oh yeah, you're absolutely right, it would.

I'd just show it for the superusers to be honest. It's people on the team, so everyone understands it's not generally visible / still in progress, but it can be useful to make people aware that it's coming. And it saves us remembering to remove this later.

michalc: If it's ok, for now, I would prefer to not. Yes, it's it's all ok for people in the team, but when people in the team are showing things to others, doing demos etc, I think it's a bit early for it to be so visible in the UI, where the only way for them to turn it off is to not be a superuser. (I think also by my understanding Alex was quite keen on it only being visible for users with a certain setting, although superusers weren't discussed explicitly)

PR #461, opened 2020-02-24
Test parallel circleci test

samuelhwilliams:
### Description of change
Run our tests and linting in parallel so that tests complete faster and linting failures are reported sooner.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,17 +1,23 @@
 version: 2
+
 jobs:
-  build:
+  docker-test:
     docker:
-      # Test and code run in _other_ docker containers, so mostly doesn't matter
       - image: circleci/python:3.7
     steps:
       - checkout
       - setup_remote_docker
       - run:
-          name: Run tests
+          name: Run test
           command: |
             touch .envs/dev.env
             make docker-test
+  check-linting:
+    docker:
+      - image: circleci/python:3.7
+    steps:
+      - checkout
+      - setup_remote_docker
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,17 +1,23 @@
...
+  docker-test:
     docker:
-      # Test and code run in _other_ docker containers, so mostly doesn't matter
       - image: circleci/python:3.7
     steps:
       - checkout
       - setup_remote_docker
       - run:
-          name: Run tests
+          name: Run test
```
allait: There is more than 1 though 😄

samuelhwilliams: Typo 🤷‍♂️

samuelhwilliams:
I'm prepared to be wrong, but I'd be surprised if it has a hard requirement on 2 before starting any. I think it's more likely that, if just one is free, it will start doing one of the jobs. And when another comes free, it will then do the second. In which case we don't really lose anything.

This mostly came from wanting to have linting come back before the tests are complete. We could reorder them and run linting first, but I don't think it really gets us anything - if I'm right with the assumption above.

samuelhwilliams:
Linting is quick yes, but with dependencies it's not instant (30-60s). I think it's fine to run in parallel.

I'm not sure how it's slower to procure two than one if they're procuring in parallel.

samuelhwilliams:
Also it's nice to see the separate github status checks for linting and tests 🙈😂

PR #462, opened 2020-02-24
Fix intermittent test failures

allait:
### Merge settings modules into one

Previously, tests were using the prod settings module and .local one wasn't ever used since the ENV variable is set in Dockefile and overrides the manage.py default.

Since most of our configuration currently is done in environment variables this merges existing prod config into base to reflect the fact that it's used in all contexts.

I'm keeping the settings/ as a package for the time being since it's possible we might end up moving a lot of non-secret config values from ENV into separate Django settings files.

### Fix DATABASES_DATA config getting out of sync in test environment

When running tests Django modifies DATABASES setting to add `test_` prefix to all database names.

Since internally we're using `DATABASES_DATA` to connect to dataset DBs (making sure we don't accidentally query the admin DB), the tests using dataset DB would actually run against a different DB than the test one (`postgres` instead of `test_postgres` in our case). The original DB is not cleaned up / recreated after tests, so any issues in it will persist and are the main cause of some of our intermittent test failures.

This moves `DATABASES_DATA` initialization after the `DATABASES` one and references the `DATABASES` dictionaries directly, which means they'll reference the same objects that get updated in `DATABASES`, making sure they stay in sync.

### Split test databases into dataworkspace and datasets

Up till now, our dev and test environments were using the same `postgres` database for both Data Workspace tables and as the datasets DB.

This made it easier to setup an environment and makes it slightly easier to create test data since we can use Data Workspace tables as datasets, which some of the application tests do.

There is a number of downsides though:

* Using `postgres` makes it hard to tell whether we're doing set up and teardown correctly since it always exists.
* Using a single DB is different from what we're doing in every other environment, so can potentially hide bugs
* Django test set up considers DATABASE records with the same name to be aliases, so the setup/teardown process is different and appears to be less predictable

Splitting the databases requires updating some of the existing tests and making sure our local postgres container creates both on startup.

This modifies the postgres container start up script to create everything needed for application tests and local application dev in one Dockerfile, since we mostly want the same things in both.

This also fixes the application tests that were previously using the auth_user table as a dataset to use the test_dataset fixture set up in postgres start up script.

### Update sample.env with separate DBs / missing vars

This splits local env DBs into admin 'dataworkspace' one and the 'datasets' one.

This requires rebuilding postgres container, which would reset local data.

### Set a flag to not migrate dataset DBs in Django 3.0

Currently, Django will run all migrations on all databases when setting up test environment.

We don't really need this on dataset DBs, but there's no way to override this behaviour in Django 2. Django 3 has an undocumented `MIGRATE` flag that skips migrations, so I'm adding to dataset DBs test configuration in case we upgrade / it gets backported.

### Checklist


samuelhwilliams:
This looks great 👍 Going to give it a few runs through locally, but it looks like it will make things a lot more reliable.

PR #464, opened 2020-02-25
Combine dataworkspace Dockerfile into multi-stage build

allait:
Our test docker image was being built on top of the live one using
a separate Dockerfile. This created two issues:

* To build an up-to-date test image the base image needs to be built
  with a separate command, which is easy to miss when not using the
  make target.
* Since base image copies the dataworkspace/ files any change to them
  would invalidate the cache for all test image layers, requiring all
  dev requirements to be reinstalled during the test run. This adds an
  additional 3-4 minutes to the test runtime.

This merges Dockerfile with Dockerfile-test into a 3-stage build
process, separating the base image with common dependencies from
live and test images. By default, when running `docker build` a
live image will be built and tagged. When running a build with a
`test` target, a test image is created, which installs test
dependencies and copies application files on top of the common base.

The main downside of this approach is that since there's currently no
way to skip a build stage, building a live image will also build the
test one, increasing Jenkins build times.

### Checklist


samuelhwilliams:
Assuming we only want to build the live image on Jenkins, I think if you set `--target live` on the Kaniko executor and then just have the `live` stage before the `test` stage in the Dockerfile, it shouldn't go on to build the `test` image?

michalc:
A related thing: it looks like we're going all-in with Debian (and specifically, debian:buster-slim) on tools and visualisations. I wonder if it would be good to move essentially _all_ parts of the infrastructure onto that at some point, to reduce the net dependencies of the system.

(Mentioning now in case there are some Alpine-related issues when tweaking the Dockerfiles. It might be good to move to Debian first if we end up doing that anyway...)

allait:
@michalc I had the same thought and gave it a try while working on this. The image size jumps from 236MB to 470MB, and some Debian packages are behind what we're running on alpine now (like the latest parallel is 2016, which doesn't have `--halt done`). But it also lets us move all python stuff to requirements.txt since it's compatible with published wheels, so I'll try to finish that up for another PR.

allait:
@samuelhwilliams I think the default `docker build` action should tag a live image, so `live` has to be the last stage. For that ticket it looks like there's a proper dependency resolution for stages in docker itself with buildkit, so maybe we just need to wait for kaniko to support that

michalc:
@allait 

> I had the same thought and gave it a try while working on this.

Ah cool!

>  The image size jumps from 236MB to 470MB

I suspect in these case it doesn't matter too much (unlike tools where the user will be sitting waiting for things to load), although I do wonder what would be taking all that space...

> (like the latest parallel is 2016, which doesn't have --halt done

... It's pretty easy to download + compile in the build, and it doesn't take too long (I have experimented with this for one of the tools builds). 

Not really sure what's best TBH

PR #466, opened 2020-02-25
Move Data Workspace admin docker image to debian base

allait:
### Move Data Workspace admin docker image to debian base

Changes our base docker image from alpine to debian-buster with python 3.7.

Debian comes with an older version of parallel, so we need to replace `--halt ...,done` with `--halt ...,fail`, which would interrupt other jobs if one fails.

### Move python dependencies to requirements.in and pip-compile them

Now that we're using Debian as a base docker image and we can install binary python packages through wheels, we can move all python dependencies into a requirements file.

This moves our dependencies list to `requirements.in` and uses `pip-compile` to generate a full requirements.txt, listing all direct and third-party requirements.

requirements.in are unpinned (apart from django) since we don't have any other known incompatibilities with the latest package versions.

### Checklist


michalc:
> Debian comes with an older version of parallel, so we need to replace --halt ...,done with --halt ...,fail, which would interrupt other jobs if one fails.

This makes me just a tiny bit nervous/feels like a step backwards. If I understand `--halt ...,fail", If one process exits with a zero exit code, i.e. a success, then the docker image will continue on, when it should stop so ECS will start another one.

I would lean to using a more recent parallel for this reason

michalc:
> python:3.7-slim-buster

Actually... I would lean to using debian:buster-slim (and apt-get install python etc), rather than the pre-build Python image. for more visibility, control and consistency: less to investigate if something suddenly stops working for example.

samuelhwilliams:
@allait are we merging this or shall we close it off for now?

PR #485, opened 2020-03-03
Add permissions for managing unpublished datasets

samuelhwilliams:
### Description of change

We want "subject matter experts" (SMEs) to be able to create, view and
update unpublished datasets to take a load of work away from our PM/DMs.

Ticket: https://trello.com/c/KBM2SCBN/873

### Checklist


dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -1,11 +1,41 @@
...
+
+def dataset_type_to_manage_unpublished_permission_codename(dataset_type: int):
+    if dataset_type == DataSetType.REFERENCE.value:
+        return 'datasets.manage_unpublished_reference_datasets'
+    elif dataset_type == DataSetType.MASTER.value:
+        return 'datasets.manage_unpublished_master_datasets'
+    elif dataset_type == DataSetType.DATACUT.value:
+        return 'datasets.manage_unpublished_datacut_datasets'
+
+    raise ValueError(f"Unknown dataset type: {dataset_type}")
```
andyhd: Couldn't this be a dict mapping ints to strings?

samuelhwilliams: Yep it could, will change.

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -1,11 +1,41 @@
...
+
+    if not dataset.published:
+        raise Http404('No dataset matches the given query.')
+
+    return dataset
+
+
+def get_dataset_type(dataset: Union[DataSet, ReferenceDataset]):
+    if isinstance(dataset, ReferenceDataset):
+        return DataSetType.REFERENCE.value
```
andyhd: Would adding a `type` property, to `ReferenceDataset` to return this value make this utility function unnecessary?

samuelhwilliams: Yes, it would. I guess it's a toss-up between whether it's weirder to add a `type` to reference datasets where they don't really _have_ a type. I could go either way, probably.

andyhd: I mean adding a `@property` rather than a model field. Not a big deal, anyway.

samuelhwilliams: Done 👍

PR #498, opened 2020-03-06
Try splitting out unit and integration tests runs

samuelhwilliams:
### Description of change
Split out unit and integration tests to see if we can get feedback faster.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,12 +84,11 @@ jobs:
             set -x
             mkdir -p caches
             make docker-build | tee docker-build.log
-            cat docker-build.log | grep '\-\-\->' | grep -v 'Using cache' | grep -v 'Running in' | sed -e "s/[ >-]//g" | sort | uniq > layers.txt
+            echo -e "caches/images.tar\n" > layers.txt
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
Looks like we can use smaller instances, which should offset the cost of running more instances. Also with small instances, and most people presumably on medium by default, we should have a lower chance of waiting in a queue as the queue is per resource-class (https://circleci.com/docs/2.0/configuration-reference/#resource_class).

Looks like with this setup we get linting back in ~30 seconds, unit tests back in ~4 minutes, and integration tests back in 6 minutes. As opposed to running them all together and waiting about ~10 minutes.

The docker build stage takes about 3 minutes, so if we ever upgraded circleci and got the layer caching, linting would be almost instant (assuming we moved it to run inside docker), unit tests would take 1 minute, and integration tests would take about 3 minutes.

samuelhwilliams:
Can't seem to get circleci to use the cached layers from our quay.io builds, which I thought might be a work-around to using circleci's docker layer caching. We always get a cache bust on COPY commands, despite hashes/permissions looking identical.

Found some SO threads about similar issues, but no solution: https://stackoverflow.com/questions/54574821/docker-build-not-using-cache-when-copying-gemfile-while-using-cache-from

PR #509, opened 2020-03-09
(feat) Link to preview visualisations

michalc:
### Description of change

This adds

- Links to preview visualisations, of the form `<host_exact>--<commit-id>.rootdomain/`
- The ability to parse these hosts properly, to know which application this is, and what the docker tag is (and the docker tag is the full `<host_exact>--<commit-id>`)

It does not add any sort of triggering of the build if the tag doesn't exist. If it doesn't exist, the spawn will error.

### Checklist

* [x] Have tests been added to cover any changes? [partially: the key permissions-related behaviour]


test/test_application.py
```diff
@@ -307,6 +307,78 @@ def add_async_cleanup(self, coroutine):
...
+                'related_emails': [],
+                'first_name': 'Peter',
+                'last_name': 'Piper',
+                'user_id': '7f93c2c7-bc32-43f3-87dc-40d0b8fb2cd2',
+            }
+        }
+        sso_cleanup, _ = await create_sso(is_logged_in, codes, tokens, auth_to_me)
+        self.add_async_cleanup(sso_cleanup)
+
+        await asyncio.sleep(15)
```
samuelhwilliams: Do we have a plan for not needing to use these any more? It feels like it will quickly become annoying if every additional test we add means the suite takes another 30 seconds. I guess you have a good idea of whether this is feasible or not based on what's happening behind the scenes?

michalc: Yes: we can poll, using a much shorter interval, whether whatever it is that we're waiting for is ready. I never quite got around to coding that up.

PR #514, opened 2020-03-10
(feat) GitLab runner to run visualisation docker builds

michalc:
### Description of change

The GitLab runner runs in the same VPC and subnets as the notebooks to avoid it having any direct internet access: they will run essentially untrusted code: the Dockerfiles of visualisations.

### Checklist

### DIFF
--- a/infra/ecs_main_gitlab.tf
+++ b/infra/ecs_main_gitlab.tf
@@ -419,7 +419,8 @@ resource "aws_instance" "gitlab" {
 
   vpc_security_group_ids      = ["${aws_security_group.gitlab-ec2.id}"]
   associate_public_ip_address = "false"
-  key_name                    = "michal"
+  key_name                    = "${var.gitlab_key}"
+
   subnet_id                   = "${aws_subnet.private_with_egress.*.id[0]}"
   user_data                   = <<EOF
   #!/bin/bash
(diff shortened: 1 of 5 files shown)


infra/ecs_main_gitlab.tf
```diff
@@ -475,3 +475,69 @@ resource "aws_eip" "gitlab" {
...
+
+resource "aws_launch_configuration" "gitlab_runner" {
+  name_prefix     = "${var.prefix}-gitlab-runner-"
+  # This is the ECS optimized image, although we're not running ECS. It's
+  # handy since it has everything docker installed, and cuts down on the
+  # types of infrastructure
+  image_id        = "ami-0749bd3fac17dc2cc"
+  instance_type   = "t3a.medium"
+  security_groups = ["${aws_security_group.gitlab_runner.id}"]
+  key_name        = "michal"
```
samuelhwilliams: What is this for?

michalc: Ah, it's the name of my private key in AWS. Will make it a variable.

michalc: Done

infra/ecs_main_gitlab.tf
```diff
@@ -479,3 +480,69 @@ resource "aws_eip" "gitlab" {
...
+    create_before_destroy = true
+  }
+}
+
+resource "aws_launch_configuration" "gitlab_runner" {
+  name_prefix     = "${var.prefix}-gitlab-runner-"
+  # This is the ECS optimized image, although we're not running ECS. It's
+  # handy since it has everything docker installed, and cuts down on the
+  # types of infrastructure
+  image_id        = "ami-0749bd3fac17dc2cc"
```
samuelhwilliams: Looks like this is the second place we're hard-coding this AMI - do we want to use it via a variable as well?

michalc: I'm happy no really. Happy with the two instances being independent from each other (say, to then change can do one at a time), and the various dev/staging/production environments would all use the same one.

samuelhwilliams: :+1

PR #520, opened 2020-03-14
(feat) Trigger GitLab pipeline if tag doesn't exist

michalc:
### Description of change

There will probably be later changes to make things more robust + give
more visibility to errors and logs

### Checklist


dataworkspace/dataworkspace/apps/applications/spawner.py
```diff
@@ -198,13 +205,46 @@ def spawn(
...
+                    application_instance.commit_id,
+                    ecr_repository_name,
+                    tag,
+                )
+                if 'id' not in pipeline:
+                    raise Exception('Unable to start pipeline: {}'.format(pipeline))
+                pipeline_id = pipeline['id']
+
+                for _ in range(0, 60 * 5):
+                    gevent.sleep(3)
```
allait: Why are we sleeping on both ends of the loop?

michalc: Because I made a mistake ;-) Will change

PR #527, opened 2020-03-17
(feat) Take GitLab pipeline into account when checking spawner status

michalc:
### Description of change

When determining the spawner status, take GitLab pipeline status and time into account, otherwise the spawner is marked as stopped because it takes too long for the task to be created

### Checklist

*~ [ ] Have tests been added to cover any changes?~

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/spawner.py
+++ b/dataworkspace/dataworkspace/apps/applications/spawner.py
@@ -335,7 +335,7 @@ class FargateSpawner:
         spawner_application_id_parsed = json.loads(spawner_application_id)
         cluster_name = spawner_options['CLUSTER_NAME']
 
-        three_minutes_ago = datetime.datetime.now() - datetime.timedelta(seconds=180)
+        three_minutes_ago = datetime.datetime.now() - datetime.timedelta(minutes=3)
         twenty_seconds_ago = datetime.datetime.now() - datetime.timedelta(seconds=20)
 
         # We can't just depend on connectivity to the proxy url, since another
--- a/dataworkspace/dataworkspace/apps/applications/spawner.py
+++ b/dataworkspace/dataworkspace/apps/applications/spawner.py
@@ -165,6 +165,7 @@ class FargateSpawner:
     ):
 
         try:
+            pipeline_id = None
             task_arn = None
             options = json.loads(spawner_options)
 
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/apps/applications/spawner.py
```diff
@@ -223,12 +224,16 @@ def spawn(
...
                     raise Exception('Unable to start pipeline: {}'.format(pipeline))
                 pipeline_id = pipeline['id']
+                application_instance.spawner_application_instance_id = json.dumps(
+                    {'pipeline_id': pipeline_id, 'task_arn': None}
+                )
+                application_instance.save(
+                    update_fields=['spawner_application_instance_id']
+                )
 
                 for _ in range(0, 60 * 5):
```
samuelhwilliams: I know that you haven't changed this in this PR, but... Any reason this isn't just `range(0, 300)`? 🤔The way it's written kinda implies it's 5 minutes, but each sleep is for 3 seconds, so that feels a little misleading.

michalc: Agreed. Will change

dataworkspace/dataworkspace/apps/applications/spawner.py
```diff
@@ -321,48 +323,80 @@ def spawn(
...
-                if not proxy_url
-                else get_task_status()
-            )
+            logger.info(spawner_options)
+            spawner_options = json.loads(spawner_options)
+            spawner_application_id_parsed = json.loads(spawner_application_id)
+            cluster_name = spawner_options['CLUSTER_NAME']
+
+            now = datetime.datetime.now()
+            fifteen_minutes_ago = now + datetime.timedelta(seconds=-15 * 60)
```
samuelhwilliams: I think I've seen this "add a negative timedelta" thing a few places in the codebase, so take or leave this, but to me it feels more readable as `now - datetime.timedelta(minutes=15)` (i.e. two changes, one is using `minutes` rather than `seconds`, and the other is subtracting a positive timedelta).

dataworkspace/dataworkspace/apps/applications/spawner.py
michalc: Changed

dataworkspace/dataworkspace/apps/applications/spawner.py
michalc: Agreed...

- Will change to minutes
- Subtracting the positive delta, I remember trying this originally and there being some issue with that, but will investigate to see if that is really the case

michalc: Made both changes (not sure what was up with me not being able to subtract the positive delta before... I suspect I just got confused...)

PR #530, opened 2020-03-17
Make sure they really do have access

samuelhwilliams:
### Description of change
There was a logic error in the "You have access" filter, which meant
that if _anyone_ had been granted access to the dataset then it would
appear. It was also resulting in each dataset appearing multiple times
in the results. This should fix these problems.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/tests/datasets/test_views.py
+++ b/dataworkspace/dataworkspace/tests/datasets/test_views.py
@@ -210,18 +210,47 @@ def test_find_datasets_filters_by_source(client):
 
 @pytest.mark.django_db
 def test_find_datasets_filters_by_access():
-    user = factories.UserFactory.create()
+    user = factories.UserFactory.create(is_superuser=False)
+    user2 = factories.UserFactory.create(is_superuser=False)
     client = Client(**get_http_sso_data(user))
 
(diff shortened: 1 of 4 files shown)


dataworkspace/dataworkspace/tests/datasets/test_views.py
```diff
@@ -210,16 +210,39 @@ def test_find_datasets_filters_by_source(client):
...
+        user_access_type='REQUIRES_AUTHORIZATION',
+    )
+
+    datacut = factories.DataSetFactory.create(
+        published=True,
+        type=DataSetType.DATACUT.value,
+        name='Datacut',
+        user_access_type='REQUIRES_AUTHORIZATION',
     )
+    factories.DataSetUserPermissionFactory.create(user=user2, dataset=datacut)
```
allait: Should we add a public non-reference dataset to this to check that it gets displayed correctly?

samuelhwilliams: master1 is a public master dataset.

samuelhwilliams: I'll rename the variables

PR #531, opened 2020-03-18
Actually de-duplicate datasets properly

samuelhwilliams:
### Description of change
I tried to fix this before in
https://github.com/uktrade/data-workspace/pull/530 but hadn't understood
the problem properly.

When a user filters for access as well as use (but not reference
datasets), then we get multiple instances of each dataset returned if
that dataset doesn't require authorisation but has multiple people
specifically authorised to use it.

This fix adds proper querying on public datasets to ensure that only one
instance is returned, and also changes the context passed into templates
to a list of dicts, rather than a list of dataset objects, which is
consistent with what's passed in when "reference" datasets are filtered
for.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/views.py
+++ b/dataworkspace/dataworkspace/apps/datasets/views.py
@@ -111,8 +111,14 @@ def filter_datasets(
     )
 
     if user and access and datasets.model is not ReferenceDataset:
-        access_filter = Q(user_access_type='REQUIRES_AUTHENTICATION') | Q(
-            datasetuserpermission__user=user
+        access_filter = (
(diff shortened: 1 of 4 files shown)


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -161,6 +167,10 @@ def find_datasets(request):
                 'uuid', 'name', 'slug', 'short_description', 'search_rank'
             )
         )
+    else:
+        datasets = datasets.values(
```
allait: We could do this first and then only do `union` inside `if`, just to keep that list of fields in 2 places instead of 3

PR #533, opened 2020-03-18
(feat) GitLab behind VPN

michalc:
### Description of change

Also opens up port 80 between the admin application and GitLab for the
API: so the communication doesn't go via the NAT/public internet, and so
we don't have to open up the GitLab load balancer to the address of the
NAT instance.

### Checklist

### DIFF
--- a/infra/security_groups.tf
+++ b/infra/security_groups.tf
@@ -313,6 +313,18 @@ resource "aws_security_group" "admin_service" {
   }
 }
 
+resource "aws_security_group_rule" "admin_service_egress_http_to_gitlab_service" {
+  description = "egress-http-to-gitlab-service"
+
+  security_group_id = "${aws_security_group.admin_service.id}"
+  source_security_group_id = "${aws_security_group.gitlab_service.id}"
+
+  type        = "egress"
+  from_port   = "80"
+  to_port     = "80"
+  protocol    = "tcp"
+}
+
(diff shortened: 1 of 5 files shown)


infra/security_groups.tf
```diff
@@ -313,6 +313,18 @@ resource "aws_security_group" "admin_service" {
   }
 }
 
+resource "aws_security_group_rule" "admin_service_egress_http_to_gitlab_service" {
+  description = "egress-https-to-ecr-api"
```
samuelhwilliams: This description doesn't sound correct - copy/paste error?

infra/security_groups.tf
```diff
@@ -876,6 +888,18 @@ resource "aws_security_group_rule" "gitlab_service_ingress_http_from_nlb" {
   protocol    = "tcp"
 }
 
+resource "aws_security_group_rule" "gitlab_service_ingress_http_from_admin_service" {
+  description = "ingress-https-from-nlb"
```
samuelhwilliams: Same here

infra/security_groups.tf
michalc: Ah yes, will change

michalc: Changed

infra/security_groups.tf
michalc: Changed

PR #535, opened 2020-03-19
(fix) Link to GitLab on visualisations page

michalc:
### Description of change

Data Workspace now has GITLAB_URL set as a private domain to allow it to resolve to a private domain, i.e. gitlab.jupyterhub. This won't work from the public internet.

### Checklist


michalc:
@allait Yes... but...

- if it's ok would prefer to do this as an optimisation later. Essentially, the design is somewhat in flux, and we may end up needing this request anyway for another reason
- plus, even the message that is displayed I would like to do-away-with, and create people in GitLab automatically, removing the need for this request

PR #547, opened 2020-03-23
Remove intermediary queue from CSV data downloads

allait:
### Remove intermediary queue from CSV data downloads

When a user downloads a CSV we used to spawn a gevent greenlet that would fetch results from DB and put them in a queue. Django then returned a streaming response that was reading data from that queue to return to the client. This allowed us to prefetch results, making downloads potentially faster on quick connections since the query is executed in parallel with the request being streamed.

This approach however turned out to be prone to creating hard to debug problems. It used to cause incomplete downloads on slow connections without reporting the error to the user until we tweaked queue timeouts and linked greenlet error handling together. Since the queue still required a timeout it was still possible for a very slow connection to fail an otherwise successful download if it took too long to fetch data.

Another recent issue potentially caused by this is that if a client has disconnected a spawned greenlet was left behind with a stuck DB transaction. That transaction would idle for hours, locking DB tables and preventing data-flow from swapping dataset tables.

To avoid these and to potentially make this code easier to reason about we remove the queue and stream DB response directly to WSGI. We still use `.fetchmany` to `yield` things in chunks as yielding each row slowed down the response (psycopg2 still uses itersize to fetch results in bulk, but WSGI server has to call the stream function a lot more)

### Add a DB statement timeout to CSV downloads

To avoid accidentally creating long running idle transactions we set a 5 minute statement timeout on all data download queries.

The timeout applies to each `FETCH` so should be long enough to allow complex queries to finish while preventing a leftover/stuck query from blocking dataset table swap for hours.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/core/utils.py
+++ b/dataworkspace/dataworkspace/apps/core/utils.py
@@ -8,7 +8,6 @@ import secrets
 import string
 import csv
 import gevent
-import gevent.queue
 from psycopg2 import connect, sql
 
 import boto3
@@ -397,12 +396,9 @@ def table_exists(database, schema, table):
 
 def streaming_query_response(user_email, database, query, filename):
     logger.info('streaming_query_response start: %s %s %s', user_email, database, query)
-    cursor_itersize = 1000
-    queue_size = 3
-    queue_timeout = 60 * 2
-    bytes_queue = gevent.queue.Queue(maxsize=queue_size)
+    batch_size = 1000
 
-    def put_db_rows_to_queue():
+    def yield_db_rows():
         # The csv writer "writes" its output by calling a file-like object
         # with a `write` method.
         class PseudoBuffer:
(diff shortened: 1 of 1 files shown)


dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -416,71 +413,52 @@ def write(self, value):
...
+                    yield csv_writer.writerow(
+                        [column_desc[0] for column_desc in cur.description]
                     )
+
                 bytes_fetched = ''.join(
                     csv_writer.writerow(row) for row in rows
                 ).encode('utf-8')
-                bytes_queue.put(bytes_fetched, timeout=queue_timeout)
+
+                yield bytes_fetched
```
michalc: Should the `if not rows: break` be above this line? If there are exactly a multiple of `batch_size` rows, then this would yield the empty string, which I suspect should be avoided. (The previous version could have had this problem as well admittedly...)

allait: Oh yeah, good catch! I don't think it affects the response, but no reason to have that extra query in there. Fixed

dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -416,71 +413,52 @@ def write(self, value):
...
-            cur.itersize = cursor_itersize
-            cur.arraysize = cursor_itersize
 
-            try:
-                cur.execute(query)
-            except Exception as ex:
-                gevent.get_hub().parent.throw(ex)
+            # set statements can't be issued in a server-side cursor, so we
+            # need to create a separate one to set a timeout on the current
+            # connection
```
michalc: If this lasts for the entire connection, then would it be clearer to set it after creating the connection, and not nest the cursors?

```python
with connect(
    database_dsn(settings.DATABASES_DATA[database])
) as conn:
   with conn.cursor() as cur:
      cur.execute("SET statement_timeout={0}".format(query_timeout))
   with conn.cursor(name='data_download'):
     ...
```

allait: Are there any issues with multiple active cursors? Separate `with` would require another level of indentation, and this already gets pretty nested

michalc: I guess I don't know to be honest. However for me, it's clearer (in spite of the indentation) since initially I thought the cursors had to be nested for some reason. No biggie.

PR #548, opened 2020-03-24
(refactor) Intermediate template

michalc:
### Description of change

Refactor to add intermediate template in chain to support the visualisations side navigation not being inside the `main` html element

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/templates/_base.html
+++ b/dataworkspace/dataworkspace/templates/_base.html
@@ -132,11 +132,7 @@
 
     {% block breadcrumbs %}{% endblock %}
 
-    <main class="govuk-main-wrapper {% block main-wrapper-class %}{% endblock %}" id="main-content" role="main">
-      {% include 'partials/messages.html' %}
-
-      {% block inner_content %}{% endblock %}
-    </main>
+    {% block main %}{% endblock %}
   </div>
 
   <footer class="govuk-footer" role="contentinfo">
--- a/dataworkspace/dataworkspace/templates/_base.html
+++ b/dataworkspace/dataworkspace/templates/_base.html
@@ -133,13 +133,7 @@
     {% block breadcrumbs %}{% endblock %}
 
     <main class="govuk-main-wrapper {% block main-wrapper-class %}{% endblock %}" id="main-content" role="main">
-      {% for message in messages %}
-      <div class="govuk-panel govuk-panel--confirmation govuk-panel--confirmation--left govuk-!-padding-5 govuk-!-margin-bottom-7">
-        <h1 class="govuk-panel__title govuk-!-font-size-27">
-          {{ message }}
-        </h1>
-      </div>
-      {% endfor %}
+      {% include 'partials/messages.html' %}
 
       {% block inner_content %}{% endblock %}
     </main>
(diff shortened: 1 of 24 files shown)


dataworkspace/dataworkspace/templates/_base.html
```diff
@@ -132,17 +132,7 @@
...
-      <div class="govuk-panel govuk-panel--confirmation govuk-panel--confirmation--left govuk-!-padding-5 govuk-!-margin-bottom-7">
-        <h1 class="govuk-panel__title govuk-!-font-size-27">
-          {{ message }}
-        </h1>
-      </div>
-      {% endfor %}
-
-      {% block inner_content %}{% endblock %}
-    </main>
+    {% block content %}{% endblock %}
```
samuelhwilliams: I think it would be nice to keep/make our block names consistent with the GOV.UK Design System, which in this case would probably call this block `main` rather than `content` (`content` being a block within `main`). I'm not sure how closely we're aligned with their block names at the moment, but having done this transition before I feel there are good reasons to try to stay in sync (easier for people to understand GOV.UK Design System based on our templates, and vice-versa, quicker for people to get up-to-speed and make changes with more confidence, etc).

https://design-system.service.gov.uk/styles/page-template/

michalc: > I think it would be nice to keep/make our block names consistent with the GOV.UK Design System,

I am good with that...

> which in this case would probably call this block main rather than content

... but note, this would in some cases contain just a `main` element, but sometimes a row with a side navigation and a main element, e.g. as in the commit in the later PR https://github.com/uktrade/data-workspace/pull/549/commits/36eaffec67d059f759ea7de36317b3308e0a1232

I would be more pro renaming what is currently `inner_content` to `main`, since that will always go into the `main` element

But happy either way.

samuelhwilliams: Given the `main` block is a wrapper around the `main` element, I don't think it's egregious for that to sometimes include extra stuff (e.g. a sidebar). But if we wanted to keep that separation, we could add an _extra_ block (e.g. `sidebar`) that goes before the `main` block, which is generally just empty?

michalc: > Given the main block is a wrapper around the main element, I don't think it's egregious for that to sometimes include extra stuff (e.g. a sidebar). 

Fair enough, will rename

> But if we wanted to keep that separation, we could add an extra block (e.g. sidebar) that goes before the main block, which is generally just empty?

Happy no (and I don't think it's quite so straightforward, since with/without side nav has a different structure beyond presence / absence of the sidebar?)

michalc: Renamed

dataworkspace/dataworkspace/templates/_main.html
```diff
@@ -0,0 +1,9 @@
+{% extends '_base.html' %}
+
+{% block main %}
+<main class="govuk-main-wrapper {% block main-wrapper-class %}{% endblock %}" id="main-content" role="main">
+  {% include 'partials/messages.html' %}
+
+  {% block inner_content %}{% endblock %}
```
samuelhwilliams: This block then just becomes `content` in the GOV.UK Design System

michalc: Will rename

michalc: (Actually... will rename in another PR)

PR #550, opened 2020-03-24
(fix) 'Skip to main content' link

michalc:
### Description of change

The base href is set to `/` on the root domain, partially to make links on the spawning and error pages, which are on other domains, work as expected. However, this means that links with just a #fragment will always go to the #fragment on the root path of the root domain, unless prefixed by the path of the current page.

out

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 2020-03-24
+
+### Changed
+
+- Fix the "Skip to main content" link: it would always link to the home page
+
 ## 2020-03-23
 
 - Fix Google Data Studio link
--- a/dataworkspace/dataworkspace/templates/_base.html
+++ b/dataworkspace/dataworkspace/templates/_base.html
@@ -67,7 +67,7 @@
     document.body.className = document.body.className + ' js-enabled';
   </script>
 
-  <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
+  <a href="{{ request.path }}#main-content" class="govuk-skip-link">Skip to main content</a>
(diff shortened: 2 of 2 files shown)


allait:
I'm confused, what happens exactly? My impression was `#fragment` is handled entirely by the browser as "add fragment to current URL", which is different to navigating to a new page. I'm not sure that the latter works just as well with the screen readers

michalc:
@allait Ah, I realise the report wasn't clear. I have added a clearer (hopefully) description and a short gif.

allait:
@michalc sorry, my bad, I missed the fact that we have a `<base href=...>` on the page. Is there a way we could only set it on the error/app spawning pages or do we need it anywhere else?

I think the problem with setting a full URL instead of fragment is that it triggers a page refresh, and I'm not sure that a screen reader will then treat it the same way as a fragment jump.

michalc:
> I think the problem with setting a full URL instead of fragment is that it triggers a page reload, and I'm not sure that a screen reader will then treat it the same way as a fragment jump.

When I tested it in Chrome, it didn't? What browser does it do this in?

michalc:
@allait 

>  Is there a way we could only set it on the error/app spawning pages or do we need it anywhere else?

Maybe... but...

- Those pages would also need a skip to main content link. So if the base href were still the root domain, then its skip to main content link can't just be the  #fragment, or it similarly would send the user to another page.

- I would be a bit nervous of using the same templates with/without a base href (or where the base href were different), especially on the slightly less tested spawner page

Thinking about an alternative... essentially the base href is so we don't have to specify what domain everything is on. We could make it so all links (and urls to style/javascript files) aways explicitly contain the full domain (which would be the ROOT_DOMAIN most of the time), and remove the base href?

(And note also, the spawning page can be on an "inner" page, e.g.  https://jupyterlabpython-9c57e86a.data.trade.gov.uk/my-page/, although I suspect this isn't relevant to the various solutions right now...)

michalc:
All good!

PR #556, opened 2020-03-27
(feat) Visualisation user page

michalc:
### Description of change

For now, just showing the users that have access. In a later PR, making this editable

Screenshot 2020-03-27 at 09 56 55

### Checklist


dataworkspace/dataworkspace/templates/_visualisation.html
```diff
@@ -0,0 +1,108 @@
+{% extends '_main_and_nav.html' %}
+{% load static %}
+{% load core_tags %}
+
+{% block page_title %}Visualisations - {{ block.super }}{% endblock %}
+
+{% block head %}
+  {{ block.super }}
+  <style>
+    .branch-icon-small {
```
samuelhwilliams: I know we have inline styles all over the place, but this is going to make our CSS a lot harder to maintain if we keep doing it this way. Can we at least pull this out into a separate file that is just included using a script src?

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -324,17 +306,101 @@ def branch_sort_key(branch):
...
+    request,
+    template,
+    gitlab_project,
+    branches,
+    current_menu_item,
+    template_specific_context,
+):
+    # For templates that inherit from _visualisation.html. This is factored
+    # out, in a way so that any context variables required, but not passed from
+    # the view have a chance to be caught by linting
```
samuelhwilliams: 👍

dataworkspace/dataworkspace/templates/_visualisation.html
michalc: I would lean to no, at least for now...

- I actually find this really easy to maintain: easy to add/remove/tweak styles, see what's being used etc, since the styles for the html are often just in the same file as the html they apply to

- The CSS files at the moment have long expires headers, with no way to cache bust other than a manual file name change. Having them inline means there is no chance the html will ever be out of sync with the styles from the point of view of the browser.

- And for new visitors, a separate file would be an extra http request to the server (and the visualisations pages are slow enough as it is...) (yes, the styles are then transferred on each page, but it would be far less than even 1 ip packet)

samuelhwilliams: I'm not going to be a stick in the mud on this, at least for now, given this isn't the first example. But I'm not sure I agree with your three points.

1) Maybe you do, but I'm not sure that's generally true as the site develops - and it's likely to lead to styles being subtly different between pages, which can be annoying and weird to debug and investigate. Centralised styles generally mean consistent styles.
2) It's relatively easy to add cache-busting to files with the same name, e.g. by adding a query param to the style src (such as a hash of the file contents, app release ID, etc).
3) Sure this is valid, but I don't generally agree that it's a super argument unless we're really optimising things tightly. Which I don't think we are yet, nor probably need to be.

PR #561, opened 2020-03-30
Add an api endpoint for event log entries

niross:
### Description of change

Adds an api endpoint for event log entries for consumption by data flow. Includes the addition of django rest framework as there are at least 3 more endpoints to be added in the near future.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/api_v1/eventlog/serializers.py
+++ b/dataworkspace/dataworkspace/apps/api_v1/eventlog/serializers.py
@@ -1,14 +1,7 @@
-from django.contrib.auth import get_user_model
 from rest_framework import serializers
 
 from dataworkspace.apps.datasets.models import DataSet, ReferenceDataset
 from dataworkspace.apps.eventlog.models import EventLog
-
-
-class EventLogUserSerializer(serializers.ModelSerializer):
-    class Meta:
-        model = get_user_model()
-        fields = ('id', 'email', 'first_name', 'last_name')
 
 
 class EventLogDatasetSerializer(serializers.ModelSerializer):
@@ -43,12 +36,11 @@ class EventLogRelatedObjectField(serializers.RelatedField):
 
 class EventLogSerializer(serializers.ModelSerializer):
     related_object = EventLogRelatedObjectField(read_only=True)
-    user = EventLogUserSerializer()
     event_type = serializers.SerializerMethodField()
 
     class Meta:
         model = EventLog
-        fields = ('id', 'timestamp', 'event_type', 'user', 'related_object', 'extra')
+        fields = ('id', 'timestamp', 'event_type', 'user_id', 'related_object', 'extra')
 
     def get_event_type(self, obj):
         return obj.get_event_type_display()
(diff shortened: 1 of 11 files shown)


dataworkspace/dataworkspace/apps/api_v1/eventlog/serializers.py
```diff
@@ -0,0 +1,54 @@
...
+from rest_framework import serializers
+
+from dataworkspace.apps.datasets.models import DataSet, ReferenceDataset
+from dataworkspace.apps.eventlog.models import EventLog
+
+
+class EventLogUserSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = get_user_model()
+        fields = ('id', 'email', 'first_name', 'last_name')
```
allait: Are we sure we need to include personal information in this? If we're just looking at some stats maybe ID is enough?

niross: Yeah good call. I will raise it with the data people. Just looking at the trello card though It mentions

> Analyse usage: what are the top downloads from a group of users who comprise a team (e.g. East of England ITAs) or profession (e.g. statisticians)

> Audit: investigate downloads by user and by catalogue item in the event of a breach

For the first item we would need email at least so maybe we could lose names... I will check.

niross: Yes so checked with the relevant data management people - this has been signed off.

dataworkspace/dataworkspace/settings/base.py
```diff
@@ -323,3 +324,9 @@ def aws_fargate_private_ip():
 
 
 CKEDITOR_BASEPATH = STATIC_URL + 'ckeditor/ckeditor/'
+
+
+REST_FRAMEWORK = {
+    'DEFAULT_PAGINATION_CLASS': 'rest_framework.pagination.LimitOffsetPagination',
```
samuelhwilliams: Do we want to use cursor-based pagination here instead? I'm not sure how many records are in the eventlog, but it feels like it could have a fair number of writes - so if we just use limit/offset, and writes occur during the data-flow task, we'll end up with duplicates in the data-flow DB. It also feels like it could be (or grow to be) a fairly long dataset, so limit/offset with page size at 100 could be pretty inefficient. Maybe?

Feels probably like a prime use-case for cursor rather than limit/offset (https://www.django-rest-framework.org/api-guide/pagination/#cursorpagination).

dataworkspace/dataworkspace/apps/api_v1/eventlog/serializers.py
niross: Ok so on second thoughts I have switched this to user id only here. I will create another PR with a user endpoint as this data will be needed by a few datasets.

dataworkspace/dataworkspace/settings/base.py
niross: Yep makes sense. I've swapped it out :+1:

PR #571, opened 2020-04-02
Instrument ELK APM

samuelhwilliams:
### Description of change
Based on https://readme.trade.gov.uk/docs/howtos/app-metrics.html

Currently not tested locally as I don't have access to ELK. But have requested access and will check it out when I get it.

### Checklist


samuelhwilliams:
Not able to test this locally as we can't access apm.ci.uktrade.io from our home laptops, even on VPN. That will be fixed but not very soon.

Michal's added the env var we need to all environments so we should just be able to test this in dev.

PR #572, opened 2020-04-02
Instrument Django-Debug-Toolbar, if installed

samuelhwilliams:
### Description of change
The debug toolbar adds some nice profiling info (e.g. sql queries for
the page, load time, headers/settings, etc) which can be handy in a dev
environment.

We only load the middleware if we're in debug mode and the library is
available, and we only add the library to the dev requirements, so it
should never be available in a production (even dev/staging)
environment.

### Show it (right-hand side)
Screenshot 2020-04-02 at 17 27 04

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/settings/base.py
+++ b/dataworkspace/dataworkspace/settings/base.py
@@ -70,6 +70,15 @@ MIDDLEWARE = [
     'csp.middleware.CSPMiddleware',
 ]
 
+if DEBUG:
+    try:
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Would probably have to stick it in requirements.txt for that, if you're using docker to run data-workspace, as I don't think that installs requirements-dev?

allait:
Oh yeah, that's a good point. We could run test container in dev docker-compose instead of live as well, which would give us dev dependencies

PR #587, opened 2020-04-07
(feat) Release to production button

michalc:
### Description of change

Initial version of button. Slightly awkward to have to preview first, but better than me releasing it every time

Screenshot 2020-04-07 at 09 35 59

Screenshot 2020-04-07 at 09 36 43

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/templates/visualisation_branch.html
+++ b/dataworkspace/dataworkspace/templates/visualisation_branch.html
@@ -21,7 +21,33 @@
 {% endif %}
 
 <section class="commit govuk-!-padding-5">
-    <h2 class="govuk-heading-m">Latest commit: {{ latest_commit.short_id }}</h2>
+    <div class="commit-row">
+        <h2 class="commit-left govuk-heading-m">Latest commit: {{ latest_commit.short_id }}</h2>
+        <form method="POST" class="commit-right">
+            {% csrf_token %}
+
+            {% if must_preview_latest_commit_to_release %}
+            <button class="govuk-button govuk-!-margin-bottom-1" data-module="govuk-button" type="submit" disabled="disabled" aria-disabled="true">
+                Release<span class="govuk-visually-hidden"> commit {{ latest_commit.short_id }}</span> to production
+            </button>
+            <p class="govuk-body govuk-!-margin-bottom-3 govuk-!-font-size-14">Preview required before release</p>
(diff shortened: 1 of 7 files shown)


dataworkspace/dataworkspace/templates/visualisation_branch.html
```diff
@@ -21,7 +21,33 @@ <h2 class="govuk-heading-m govuk-heading-production">
...
+    <div class="commit-row">
+        <h2 class="commit-left govuk-heading-m">Latest commit: {{ latest_commit.short_id }}</h2>
+        <form method="POST" class="commit-right">
+            {% csrf_token %}
+
+            {% if must_preview_latest_commit_to_release %}
+            <button class="govuk-button govuk-!-margin-bottom-1" data-module="govuk-button" type="submit" disabled="disabled" aria-disabled="true">
+                Release<span class="govuk-visually-hidden"> commit {{ latest_commit.short_id }}</span> to production
+            </button>
+            <p class="govuk-body govuk-!-margin-bottom-3 govuk-!-font-size-14">Preview required before release</p>
```
samuelhwilliams, line 33: I'm not a designer, but I wonder if this button should just become a "preview visualisation" button if that's a required step. Then when it's been previewed, it can just become the "Release to production" button. That way, there's always a clear "call to action" for users that steps them through the process.

Feels like it might be slightly better to me - though it could also be a bit magical with a button in one place changing purpose/function. 🤷‍♂

michalc: This crossed my mind, but...

- Release and preview are such radically different things, and release is somewhat "dangerous", so I suspect it's better that the preview link/button does not turn into a release button to minimise the risk of accidentally releasing when the user meant to preview.
-Tthe preview link would then still need to be visible somewhere even if previously previewed, so that leaves either having it duplicated, or moving location depending on state, neither of which I think are good
- I am deliberately leaving preview as a link and not a button to suggest that it doesn't really "do" anything and is safe to click at any point, while leaving the release button as the only thing that does anything
- This is meant to be an interim step to not requiring a preview, so the release button in the top right will hopefully be active from the get-go in future versions

samuelhwilliams: Ok, sounds sensible 👍

PR #597, opened 2020-04-09
Visualisations catalogue item model

samuelhwilliams:
### Description of change
Add a model to the application for visualisations in the search
catalogue, including an admin interface. The model is one-to-one linked
to a visualisation application template, which contains the launch
configuration for the vis. Some of the information is kept in sync
between visualisation templates and the catalogue item (at the moment
just the name).

### Checklist


dataworkspace/dataworkspace/apps/datasets/migrations/0040_visualisationcatalogueitem.py
```diff
@@ -0,0 +1,108 @@
...
+                (
+                    'id',
+                    models.UUIDField(
+                        default=uuid.uuid4,
+                        editable=False,
+                        primary_key=True,
+                        serialize=False,
+                    ),
+                ),
+                ('name', models.CharField(max_length=255)),
```
allait, line 34: We need a slug here similar to datasets in order to make URLs a bit more friendly

samuelhwilliams: I think we can probably used `visualisation_template.name` as the slug? It's URL friendly and should be guaranteed to exist.

samuelhwilliams:
@allait Ok - have added a slug field. Have also added `published_at` and `updated_at` fields, which are part of the design for the data visualisation page.

PR #606, opened 2020-04-14
Add visualisation catalogue detail page

samuelhwilliams:
### Description of change
Adds the catalogue detail page for visualisations.

Ticket: https://trello.com/c/GX0ULwT1/945

### Show it
screencapture-dataworkspace-test-8000-datasets-28d3ee21-947f-4965-86d8-3c0ab42180db-2020-04-14-17_26_17

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/constants.py
+++ b/dataworkspace/dataworkspace/apps/datasets/constants.py
@@ -10,3 +10,4 @@ class DataSetType(enum.Enum):
     REFERENCE = 0
     MASTER = 1
     DATACUT = 2
+    VISUALISATION = 3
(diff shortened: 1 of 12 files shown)


dataworkspace/dataworkspace/apps/datasets/constants.py
```diff
@@ -10,3 +10,4 @@ class DataSetType(enum.Enum):
     REFERENCE = 0
     MASTER = 1
     DATACUT = 2
+    VISUALISATION = 3
```
allait, line 13: Are we happy with this? I kind of broke this enum when I added reference value to it, which isn't actually stored in Datasets table, but there was a vague plan to merge the tables and 0 felt like a special enough number. I've run into this as well with visualisations and started doing a separate string value for their "use" filter, which makes "none selected means all selected" a bit fiddly.

I guess what I'm saying is - this feels hacky, but makes my life easer as well, so I'm happy to go with it if we feel it's a reasonable trade off 😅

samuelhwilliams: I'm OK keeping it for now. I agree it's a bit hacky.

PR #609, opened 2020-04-15
(refactor) Reduce usage of application 'name', to remove later

michalc:
### Description of change

This is so there will be less to setup for each visualisation.

The change of the database role _should_ be safe no matter the
characters in the host_basename: it gets escaped using sql.Identifier
before being sent to the database.

The usage of `name` in views.py appears to be unused in the template, so
it doesn't need to be replaced by anything.

### Checklist


dataworkspace/dataworkspace/apps/applications/migrations/0003_auto_20200415_1143.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [('applications', '0002_remove_applicationtemplate_host_pattern')]
+
+    operations = [
+        migrations.AlterField(
+            model_name='applicationtemplate',
+            name='name',
+            field=models.CharField(
+                help_text='Used in URLs: only lowercase letters allowed',
+                max_length=128,
+                validators=[django.core.validators.RegexValidator(regex='^[a-z]+$')],
```
samuelhwilliams, line 18: Is it reasonable to also allow dashes/underscores here?

PR #617, opened 2020-04-16
(feat) Automatically create visualisation

michalc:
### Checklist


dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -637,6 +617,82 @@ def branch_sort_key(branch):
...
+                    spawner_options='{}',
+                    application_type='VISUALISATION',
+                    user_access_type='REQUIRES_AUTHORIZATION',
+                    gitlab_project_id=gitlab_project['id'],
+                    visible=False,
+                )
+                VisualisationCatalogueItem.objects.create(
+                    name=gitlab_project['name'],
+                    slug=path_and_dns_safe_name,
+                    short_description=gitlab_project['description'],
```
samuelhwilliams, line 676: I was going to question this - as it looks like Gitlab lets you put quite a lot of text in this field (including paragraphs), whereas we only want a single sentence. But generally it looks like on gitlab it's currently either not used or a single sentence anyway, so this is probably fine.

PR #621, opened 2020-04-17
Add a Data Vis UI catalogue edit page

samuelhwilliams:
### Description of change
Add a page to the data visualisation UI that allows users to edit the
information that will be displayed on the catalogue page (i.e. this is
functionally-equivalent to the Django admin page).

Some things of note:
* Descriptions for reference/datacut/master datasets use a richtext
field editor for descriptions. This requires blowing a hole in our CSP
header, and generally provides features that aren't needed or desirable
for our content creators. The visualisation description therefore
supports basic markup - paragraphs and lists only. We can extend this
further if we need additional work in the future.
* We'll probably want to do some work around simplifying how we render
forms directly using GOV.UK Design System styles - at the moment it's
quite manual and fiddly, including specifying classes in Python code.

Ticket: https://trello.com/c/CvOWjmvD/978

### Show it
screencapture-dataworkspace-test-8000-visualisations-9-catalogue-item-2020-04-20-11_44_40

screencapture-dataworkspace-test-8000-visualisations-9-catalogue-item-2020-04-20-11_45_28

### Checklist


dataworkspace/dataworkspace/templates/visualisation_catalogue_item.html
```diff
@@ -0,0 +1,55 @@
...
+</div>
+{% endif %}
+
+<h1 class="govuk-heading-l govuk-!-margin-bottom-6">
+  <span class="govuk-caption-l">{{ gitlab_project.name }}</span>
+  Catalogue item
+</h1>
+
+<!-- TODO: Link up `published` -->
+<p class="govuk-body">These details are visible in the catalogue once the visualisation has been <a class="govuk-link" href="#">published</a>.</p>
```
michalc, line 35: Can this not be a link until it the published page exists? (Since it feels like a bit of a bug that the link doesn't go anywhere, at least to me)

dataworkspace/dataworkspace/apps/applications/forms.py
```diff
@@ -0,0 +1,29 @@
...
+class VisualisationsUICatalogueItemForm(GOVUKDesignSystemModelForm):
+    class Meta:
+        model = VisualisationCatalogueItem
+        fields = [
+            'short_description',
+            'description',
+            'enquiries_contact',
+            'secondary_enquiries_contact',
+            'information_asset_manager',
+            'information_asset_owner',
```
michalc, line 16: I've noticed that the VisualisationCatalogueItem form in admin is a bit slow... and I suspected due to loading all of the users in the database 4 times. I might be tempted to just have text inputs here, and for the user to enter the email address of the various contacts.

For one thing, similar to giving access to a visualisation, there is probably going to be the need to be able to add a user that _isn't_ in the database by adding their email address, so it's going to have to do something a bit more than choosing a user from a drop down: i.e. from the given email address search SSO if the email isn't already in the system.

(And to achieve this, I would probably not use the Django form system, and just populate the HTML manually. To me, that's more flexible and explicit)

dataworkspace/dataworkspace/templates/visualisation_catalogue_item.html
samuelhwilliams: Sure.

dataworkspace/dataworkspace/apps/applications/forms.py
samuelhwilliams: We chatted a bit about this off-site. Will stick with this for the MVP, but we will want to circle back to this fairly soon in order to address 1) performance, 2) data controls, and 3) user experience.

samuelhwilliams:
@michalc updated as per our discussion

dataworkspace/dataworkspace/apps/applications/forms.py
```diff
@@ -0,0 +1,91 @@
...
+class _NiceEmailValidationModelChoiceField(ModelChoiceField):
+    def clean(self, value):
+        if value:
+            EmailValidator(message=self.error_messages['invalid_email'])(value)
+        return super().clean(value)
+
+
+class VisualisationsUICatalogueItemForm(GOVUKDesignSystemModelForm):
+    enquiries_contact = _NiceEmailValidationModelChoiceField(
+        queryset=get_user_model().objects.all(),
```
michalc, line 18: Would this actually query for all users in the database under the hood, i.e. take them all from the database and load them into memory?

samuelhwilliams: I don't believe so - I believe it's just the basis for later filtering on using the value of the field - i.e. when a user enters an email address, it's matched against all of the rows - rather than, say, just staff.

michalc: Ah ok

PR #622, opened 2020-04-17
Add reference code and reference numbers to datasets and sources

niross:
### Description of change

Bring dataset source download inline with naming conventions

The naming conventions call for a unique reference to be prepended to any outputs generated by data workspace. The reference should identify the source of the data (eg. DH for data hub, EW for export wins) and be followed by a sequential number (padded to 5 digits).

- Add nullable `refernce_number` field to dataset sources (link, query, view, table)
- Adds nullable foreign key from `DataSet` to new `DatasetReferenceCode` model
- `DatasetReferenceCode` model contains a code and a counter 
- When a dataset is assigned a reference code, it's related sources are assigned reference numbers by incrementing the counter
- When a source is downloaded the reference code and number are prepended to the file name followed by a dash (eg. DH00001-datahub-interactions.csv, EW23412-export-wins-wins.csv).

### Checklist


dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -110,6 +110,28 @@ def __str__(self):
...
+        ordering = ('code',)
+
+    def __str__(self):
+        return self.code
+
+    def get_next_reference_number(self):
+        self.counter = F('counter') + 1
+        self.save(update_fields=['counter'])
+        self.refresh_from_db()
+        return self.counter
```
samuelhwilliams, line 132: Is it worth explicitly wrapping this section with a `transaction.atomic`? It doesn't look like we currently have atomic requests at the moment.

niross: Yes good call, I've added it in case we do use it outside the admin at some point :+1:

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -110,6 +110,29 @@ def __str__(self):
...
 
 
+class DatasetReferenceCode(TimeStampedModel):
+    code = models.CharField(
+        max_length=20,
+        unique=True,
+        help_text='Short code to identify the source (eg. DH for Data Hub, EW for Export Wins)',
+    )
+    description = models.TextField(null=True, blank=True)
+    counter = models.IntegerField(default=0)
```
samuelhwilliams, line 120: Maybe also a `unique_together` on code+counter?

samuelhwilliams:
Looks good. Will 👍 after resolving the comment above

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Will that make a difference if the code already has to be unique?

samuelhwilliams: No, I totally read this differently for a sec and thought that this was part of each sourcelink/query/etc.

PR #624, opened 2020-04-18
(feat) More grant locking

michalc:
### Description of change

Recent changes seem to have fixed "tuple concurrently updated" when GRANTing on the database, but the error now occurs later in the granting process.

At the moment, these permissions are created synchronously on starting an application, so the locking would slow down this request. However, treating that as better than erroring. Ideally they would be created as part of the subsequent background task that runs when starting an application, but leaving that until later.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -471,13 +471,14 @@ def delete_unused_datasets_users():
                                 )
                             )
 
-                        cur.execute(
-                            sql.SQL(
-                                'REVOKE ALL PRIVILEGES ON DATABASE {} FROM {};'
-                            ).format(
-                                sql.Identifier(database_name), sql.Identifier(usename)
+                            cur.execute(
+                                sql.SQL(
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -471,23 +471,27 @@ def delete_unused_datasets_users():
...
-                        )
 
                         for schema in schemas:
                             for schema_revoke in schema_revokes:
                                 try:
-                                    cur.execute(
-                                        sql.SQL(schema_revoke).format(
-                                            sql.Identifier(schema),
-                                            sql.Identifier(usename),
+                                    with cache.lock(
```
samuelhwilliams, line 486: Would it be better to do this lock one level up - i.e. outside of the `for schema_revoke in schema_revokes` loop? It would make the lock hold for longer period in one go, but there'd be fewer locks/unlocks for the function to get hung up on.

michalc: I'm not sure at all which would be better... but seems a reasonable argument: will change it.

PR #632, opened 2020-04-20
Add visualisations to catalogue search results

allait:
### Description of change

image

### Add "purpose" to the search result descriptions

Adds the filter name label based on the dataset type to each result.

We add the type to the reference datasets by making Django ORM generate a `SELECT ..., 0 as "type"` query.

Django orders model fields before any additional fields like annotations in generated SQL statement which means doing a UNION between a model field in one query and an extra field in another results in different columns getting connected into a single result one (which sometimes leads to a type error and sometimes succeeds with values ending up in the wrong place). To avoid this, we alias model field `type` with `.annotate` in the datasets query, making sure it's added to the end of SELECT field list.

### Add visualisations to catalogue search results

Visualisation follow the same rules as datasets:

* they are filtered by text query
* Access filter only displays visualisations the user has permission to launch (either public ones or the ones the user has an explicit permission for)
* Unpublished visualisations are visible in search results for users with a permission to manage them
* Source filter for visualisations is based on source tag of the connected datasets

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/views.py
+++ b/dataworkspace/dataworkspace/apps/datasets/views.py
@@ -12,7 +12,7 @@ from django.conf import settings
 from django.contrib.postgres.search import SearchQuery, SearchRank, SearchVector
 from django.core.paginator import Paginator
 from django.core.serializers.json import DjangoJSONEncoder
-from django.db.models import F, Q
+from django.db.models import IntegerField, F, Q, Value
 from django.forms import model_to_dict
 from django.http import (
     Http404,
@@ -78,7 +78,6 @@ def filter_datasets(
     source,
     use=None,
     user=None,
-    form=None,
 ):
     search = SearchVector('name', 'short_description', config='english')
     search_query = SearchQuery(query, config='english')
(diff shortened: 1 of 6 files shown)


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -134,38 +133,94 @@ def filter_datasets(
     return datasets
 
 
+def filter_visualisations(query, access, source, user=None):
+    search = SearchVector('name', 'short_description', config='english')
+    search_query = SearchQuery(query, config='english')
+
+    if user and user.has_perm(
+        dataset_type_to_manage_unpublished_permission_codename(
+            DataSetType.VISUALISATION.value
```
samuelhwilliams, line 142: I think this is probably where adding this perm for visualisations has led us in the wrong direction - sorry. This perm doesn't do anything at the moment and doesn't really fit into the existing model for visualisation perms, which depends on checking permissions in gitlab. I wonder if we should delete it.

To match the existing perms, I think we need to filter to include unpublished datasets where `gitlab_has_developer_access(request.user, visualisation.visualisation_template.gitlab_project_id)` is True, which can't be done directly in SQL. The easiest thing to do might be to hit the gitlab API and find all projects where this holds true, store them as `developer_gitlab_project_ids`, and then do a `visualisation__visualisation_template__gitlab_project_id__in=[developer_gitlab_project_ids]`? Or something to that effect.

Does that make sense? @allait @michalc 

An alternative would be to somehow sync the perms between gitlab and our "manage unpublished visualisations" perm in django.

michalc: Easier still might be to not show unpublished visualisation records at all to anyone for now, and leave it until later?

(And then... I would be a bit scared of hitting the GitLab API for every search since it is slow...)

samuelhwilliams: `gitlab_has_developer_access` is now at least partially cached - but yes, we'd definitely need to pay some particular attention to that and possibly cache more aggressively.

If we don't show them in the search then we probably need to give them a "preview catalogue item" link/button from data vis UI, otherwise they'll struggle to see how it actually looks on the page (which is useful for drafting/reviewing it).

allait: I agree with Michal - accessing Gitlab API from the homepage feels a bit much. I'm also not sure developer access and managing the catalogue item will necessarily always stay the same, so it's not obvious to me that they should be based on the same access check.

But for now, can we populate the permission when the user creates catalogue page or somewhere else along the journey?

I feel like this getting out of sync with gitlab is potentially confusing, but since it's view only it shouldn't be critical. And if the permission is updated somewhere a new contributor would potentially visit before searching for the unpublished visualisation page it might work out better than time-based caching.

Alternatively we keep them manually assigned by admins for now. I think that's better than pulling it altogether.

allait: Also I think being able to visit the visualisation page itself currently relies on the same permission, so even if we're going to go with the direct link from visualisation management pages that part would need to get integrated with Gitlab.

samuelhwilliams: > But for now, can we populate the permission when the user creates catalogue page or somewhere else along the journey?

> Alternatively we keep them manually assigned by admins for now. I think that's better than pulling it altogether.

Either/or OK with me 👍

PR #646, opened 2020-04-22
Allow peer ip to work with no proxies

samuelhwilliams:
### Description of change
If running data workspace entirely locally, there are no
`x-forwarded-for` entries, so `get_peer_ip` dies when trying to retrieve
that header.

If there are no expected proxies, and no x-forwarded-for entries, we
should just directly use the peer ip from the request. If either
x-forwarded-for is set, or we expect there to be some proxies, then we
keep the current behaviour.

### Checklist


samuelhwilliams:
Going to opt for making it easier to run the nginx proxy locally instead.

PR #652, opened 2020-04-23
Allow users to unapprove visualisations

samuelhwilliams:
### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/views.py
+++ b/dataworkspace/dataworkspace/apps/applications/views.py
@@ -794,9 +794,11 @@ def visualisation_approvals_html_GET(request, gitlab_project):
     approvals = VisualisationApproval.objects.filter(
         visualisation=application_template, approved=True
     ).all()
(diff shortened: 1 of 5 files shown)


dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -794,9 +794,14 @@ def visualisation_approvals_html_GET(request, gitlab_project):
     approvals = VisualisationApproval.objects.filter(
         visualisation=application_template, approved=True
     ).all()
-    already_approved = any(filter(lambda x: x.approver == request.user, approvals))
+
+    try:
+        approval = next(filter(lambda a: a.approver == request.user, approvals))
```
michalc, line 799: Very minor/style thing, but I think `next` takes a second parameter that is returned in the case of an empty iterator

samuelhwilliams: Oh yes it does. Good point!

PR #659, opened 2020-04-24
Add publishing page to data vis UI

samuelhwilliams:
### Description of change
Adds a page to the Data Vis UI that allows a data vis creator to
publish/unpublish both the production visualisation URL, and the
catalogue item. These respectively control whether the visualisation
can accessed at its prod URL, and whether other users can discover the
visualisation exists at all in the Data Workspace catalogue.

Ticket: https://trello.com/c/FJLZZZAP/980

### Show it
screencapture-dataworkspace-test-8000-visualisations-6-publish-2020-04-24-13_04_30
screencapture-dataworkspace-test-8000-visualisations-6-publish-2020-04-24-13_04_35

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/views.py
+++ b/dataworkspace/dataworkspace/apps/applications/views.py
@@ -18,7 +18,7 @@ from django.db import IntegrityError, transaction
 from django.http import HttpResponse, Http404
 from django.shortcuts import redirect, render
 from django.urls import reverse
-from django.utils.encoding import force_text
+from django.utils.encoding import force_str
 
 from dataworkspace.apps.api_v1.views import (
     get_api_visible_application_instance_by_public_host,
@@ -146,11 +146,10 @@ def tools_html_GET(request):
                     'link': link(application_template),
                     'instance': application_instances.get(application_template, None),
                 }
-                for application_template in ApplicationTemplate.objects.all().order_by(
-                    'nice_name'
-                )
+                for application_template in ApplicationTemplate.objects.all()
+                .filter(visible=True, application_type='TOOL')
+                .order_by('nice_name')
                 for application_link in [link(application_template)]
-                if application_template.visible
             ],
             'appstream_url': settings.APPSTREAM_URL,
             'your_files_enabled': settings.YOUR_FILES_ENABLED,
@@ -454,7 +453,7 @@ def visualisation_users_with_access_html_POST(request, gitlab_project):
                 user_id=request.user.pk,
                 content_type_id=content_type_id,
                 object_id=user.id,
-                object_repr=force_text(user),
+                object_repr=force_str(user),
                 action_flag=CHANGE,
                 change_message=f'Removed visualisation {application_template} permission',
             )
@@ -551,7 +550,7 @@ def visualisation_users_give_access_html_POST(request, gitlab_project):
             user_id=request.user.pk,
             content_type_id=content_type_id,
             object_id=user.id,
-            object_repr=force_text(user),
+            object_repr=force_str(user),
             action_flag=CHANGE,
             change_message=f'Added visualisation {application_template} permission',
         )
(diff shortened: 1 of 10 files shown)


samuelhwilliams:
@allait @michalc updated to include the following:

* Require visualisation to be published before catalogue item can be
* Require all catalogue item fields to be complete (except eligibility criteria)

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -903,3 +902,212 @@ def visualisation_datasets_html_GET(request, gitlab_project):
...
+            catalogue_item.restrictions_on_usage,
+            catalogue_item.personal_data,
+        ]
+    )
+
+
+def _render_visualisation_publish_html(
+    request, gitlab_project, catalogue_item=None, errors=None
+):
+    if not catalogue_item:
```
michalc, line 961: Minor thing, but catalogue_item seems to always be `None`, so no need for this if?

michalc: (and then no need for the named parameter `catalogue_item` in this function?)

dataworkspace/dataworkspace/apps/core/templatetags/core_filters.py
```diff
@@ -36,3 +36,9 @@ def minimal_markdown(text):
             strip=True,
         )
     )
+
+
+@register.filter
+def non_breaking(text):
+    """Replace normal ASCII spaces and hyphens with non-breaking equivalents - for display only."""
+    return text.replace(' ', ' ').replace('-', '‑')
```
michalc, line 44: Just to check that copy/pasting a domain with this change is fine? (I did a very brief test, and Chrome seemed to change the domain to a punycoded version `xn--`)

dataworkspace/dataworkspace/templates/visualisation_publish.html
```diff
@@ -0,0 +1,99 @@
...
+      {% if visualisation_published %}
+      Unpublish this visualisation to stop users from accessing it at
+      <a class="govuk-link" href="{{ visualisation_link }}">{{ visualisation_domain | non_breaking }}</a>.
+      {% else %}
+      Publish this visualisation to make the latest release available for users at
+      <a class="govuk-link" href="{{ visualisation_link }}">{{ visualisation_domain | non_breaking }}</a>.
+      {% endif %}
+    </p>
+  </div>
+  <div class="govuk-grid-column-one-third">
```
michalc, line 52: Very minor thing, and not sure what's best in terms of accessibility, but to my eye it's neater to align the buttons to the right in these columns. Would be more visually consistent with the Release to production button.

dataworkspace/dataworkspace/templates/visualisation_publish.html
```diff
@@ -0,0 +1,99 @@
...
+    <form method="POST" action="{{ request.path }}">
+      {% csrf_token %}
+      {% if visualisation_published %}
+      <input id="unpublish-visualisation" name="action" type="hidden" value="unpublish-visualisation" />
+      <input type="submit" class="govuk-button" value="Unpublish visualisation" />
```
michalc, line 57: Another minor thing, but I think you can use a button with a `name="action"` and a `value="..."`, so you don't need the hidden element?

michalc: ^ Also then, you can use a single form around everything... and also it might make sense to use `aria-labelledby` to link the buttons to their explanatory text?

dataworkspace/dataworkspace/apps/core/templatetags/core_filters.py
samuelhwilliams: Copy/pasting probably does break, but the underlying link is fine. I'm honestly not sure if I'd class it as a usability concern, but I do understand that some people _might_ copy the text rather than the link ... so I could be persuaded to not use this.

dataworkspace/dataworkspace/apps/applications/views.py
samuelhwilliams: Missed hooking it up to `_set_published_on_catalogue_item`, where it saves a trip to the db.

dataworkspace/dataworkspace/templates/visualisation_publish.html
samuelhwilliams: Sure, can do. Don't think it affects accessibility much either way.

dataworkspace/dataworkspace/apps/core/templatetags/core_filters.py
michalc: I think better to not, essentially due the combination of:

- I think people so far have been emailing/messaging the links to visualisations about a lot (since there isn't another way to distribute them), so they would be copying them from available places
- I think the behaviour if it happens looks quite dodgy, even suspicious (i.e. "what's causing the address to change in the address bar? Am I being hacked?"), so even if it's rare, it doesn't make DW look good.

Alternatives:

- Could be able to be done with CSS?
- Maybe we don't need to show the address here?
- Or maybe a line break before it enough, and allow the address to wrap?
- Or maybe the layout with the text on the left/button on the right just isn't suitable?

dataworkspace/dataworkspace/templates/visualisation_publish.html
samuelhwilliams: Not sure that using a consolidated button adds much, but have done it. Not convinced about using a single form because it moves it to a much higher level. That's fine semantically but I guess it feels less neatly contained - so without a good reason I've just left it as-is.

Added the aria labels too - good suggestion.

dataworkspace/dataworkspace/apps/core/templatetags/core_filters.py
samuelhwilliams: I'm going to go with not showing the address.

PR #661, opened 2020-04-24
Visualisation access request zendesk and notify

allait:
### Description of change
image

-----

image

-----

image

-----
image

### Send an email notification to visualisation contacts on access request

Sends an email through GOV.UK Notify to primary and secondary contacts with the info about visualisation access requests.

The email contains a link to the "give access" page with the user email address and related Zendesk ticket ID in an encrypted token.

This allows us to pre-populate the email field and update the ticket in Zendesk if the visualisation owner grants access.

We're encrypting the token to avoid leaking the email in web server logs. Another approach would be to send the user ID with HMAC (to avoid exposing user emails by modifying the ID), but encryption achieves the same result without having to look up the user object on the "give access" page.

### Pre-fill user email when giving access by following an email link

Pre-fills requester email address on "give access" page when the visualisation developer clicks on the email link.

If the access is granted this adds a comment to the related Zendesk ticket and marks it as solved, but only if the access is given to the same email address listed in the original request.

### Add full request access flow for data visualisations

Adds eligibility criteria and goal form for visualisation access requests.

Updates the templates to mention "data visualisation" instead of "dataset" for visualisation requests.

### Checklist


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -348,14 +355,21 @@ def eligibility_criteria_view(request, dataset_uuid):
...
+            'dataset': dataset,
+            'is_visualisation': isinstance(dataset, VisualisationCatalogueItem),
+        },
+    )
 
 
 @require_http_methods(['GET', 'POST'])
 def request_access_view(request, dataset_uuid):
-    dataset = find_dataset(dataset_uuid, request.user)
+    dataset = find_dataset_or_visualisation(dataset_uuid, request.user)
```
michalc, line 372: Minor thing, but should `dataset_uuid` be changed to `dataset_uuid_or_visualisation_id` or similar throughout, including in the route definition `<str:dataset_uuid>/request-access`, and `dataset` be changed to `dataset_or_visualisation`?

(Or, should there be a separate route for each request access view, and each wouldn't have any branching?)

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -477,16 +479,19 @@ def visualisation_users_give_access_html_view(request, gitlab_project_id):
     if not gitlab_has_developer_access(request.user, gitlab_project_id):
         raise PermissionDenied()
 
+    token = request.GET.get("token")
+    token = decrypt_token(token.encode('utf-8')) if token else {}
```
samuelhwilliams, line 483: Minor - but possibly worth renaming this to `token_data`?

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -556,6 +564,13 @@ def error(email_address_error):
             change_message=f'Added visualisation {application_template} permission',
         )
 
+    if email_address == token.get("email", "").strip().lower():
+        update_zendesk_ticket(
+            token["ticket"],
+            comment=f"Access granted by {request.user.email}",
+            status="solved",
+        )
```
samuelhwilliams, line 572: This feels like a good idea but I think it will be worth letting data/PM know that if the approvers don't follow the link in the email, the zendesk ticket won't get resolved. So if a ticket is still open after 5 days the first thing to do is check if access has already been granted.

I have a sneaking suspicion that the "check the user is OK" process might take people away from their email, and then they're less likely to click the specific link and more likely to organically come back to the page via the data vis UI flow, which would leave the ticket open.

What do you think?

dataworkspace/dataworkspace/apps/datasets/views.py
allait: I've done separate views initially, but with eligibility criteria flow included the logic is mostly the same. I agree that the variable name is not ideal, but that's what most templates use at the moment, and changing everything to `dataset_or_visualisation` doesn't feel like a great outcome. Maybe we need to agree on a new name for "catalogue item" going forward.

dataworkspace/dataworkspace/apps/applications/views.py
allait: :+1:

dataworkspace/dataworkspace/apps/applications/views.py
allait: I've looked into Zendesk API search and there doesn't seem to be an easy way to attach a reference to a ticket to make it possible to find things related to a specific request.

We could start storing access requests and accept/decline them in vis UI instead of entering the email manually, which would allow us to store zendesk ID alongside the request, but this requires redesigning a large chunk of the flow, so I think for the initial version we'll have to check this manually.

PR #668, opened 2020-04-27
Auth checkbox on data vis UI

samuelhwilliams:
### Description of change
Adds a checkbox to the "catalogue item" page in the data vis UI for data vis creators to set whether the vis is available to all logged-in users, or just to a set of explicitly-authorised people.

If it's open for all logged-in users, creators don't need to provide eligibility criteria or manage users with access - so we hide both those sections.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Automatically lowercase emails in data vis UI catalogue page before searching db.
 
+## 2020-04-27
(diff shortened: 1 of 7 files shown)


michalc:
Ah also: is the value of user_access_type logged logged when its changed, and by whom?

samuelhwilliams:
@michalc We did consider putting it under the users section, but ended up with a decision to keep it on the catalogue.

Re: logging that's a good point - let me add a log event.

I'll call out your final point during the run-through tomorrow - but at the moment product is happy with it working this way.

PR #712, opened 2020-05-12
Trial syncing datasets/permissions to QuickSight

samuelhwilliams:
### Description of change
Rudimentary implementation of a management command which can be used to
sync datasets and associated user permissions to AWS QuickSight so that
we can test out whether the tool might be a good fit for us. This is
purposefully a bit shonky because it's only for prototyping.

Some infrastructure changes are needed in order to allow QuickSight to
connect to the Data Workspace datasets DB - these have been/will be
created manually in AWS console for now.

Ticket: https://trello.com/c/gRDNW4YS/1064

The plan is to run this manually via AWS Console for each environment via `django-admin` using existing container overrides to change the start command from `./start.sh` to something like `django-admin sync_sources_and_permissions_to_quicksight`.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/management/commands/sync_sources_and_permissions_to_quicksight.py
+++ b/dataworkspace/dataworkspace/apps/applications/management/commands/sync_sources_and_permissions_to_quicksight.py
@@ -1,0 +1,328 @@
+from collections import defaultdict
+from typing import List, Dict
+
+import boto3
+import psycopg2
+from django.core.management.base import BaseCommand
+from django.conf import settings
+
+from dataworkspace.apps.core.utils import database_dsn
+from dataworkspace.apps.datasets.models import MasterDataset
+
+
+QUICKSIGHT_COLUMN_TYPES_MAP = defaultdict(
+    lambda: 'STRING',
+    **{
+        "date": 'DATETIME',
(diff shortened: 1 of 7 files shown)


dataworkspace/dataworkspace/apps/applications/management/commands/sync_sources_and_permissions_to_quicksight.py
```diff
@@ -0,0 +1,328 @@
...
+        super().__init__(*args, **kwargs)
+
+    def _make_dataset_id(self, dataset):
+        return f"{settings.ENVIRONMENT.upper()}-{str(dataset.id)}"
+
+    def _get_dataset_columns(self, connection, source_table):
+        with connection.cursor() as cursor:
+            cursor.execute(
+                "SELECT column_name, data_type FROM information_schema.columns WHERE table_name = %s",
+                [source_table.table],
```
michalc, line 77: This may not return the right thing when the same table name is used in multiple schemas? (At least historically, a lot of table names were just "L1")

samuelhwilliams: Yes you're right - I need to add the schema 👍 Doing now.

PR #718, opened 2020-05-14
(dev) Ability to load up Metabase locally

michalc:
### Description of change

Not connected to any database other than the ephemeral one it would load
up in its container, but enough to test a bit of code in the proxy.

### Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,6 +12,7 @@ services:
     links:
       - "data-workspace-postgres"
       - "data-workspace-redis"
+      - "data-workspace-metabase"
 
   data-workspace-postgres:
     build:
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -31,5 +32,10 @@ services:
     ports:
       - "6379:6379"
 
+  data-workspace-metabase:
```
allait, line 35: Shall we link it to the DB?

docker-compose-dev.yml
```diff
@@ -48,6 +49,10 @@ services:
     image: data-workspace-redis
     ports:
       - "6379:6379"
+  data-workspace-metabase:
```
allait, line 52: Is this a bit heavy for the dev compose? I think it makes sense for the main one since it's close to live, but we don't need metabase for most of the day to day dev

michalc: Will remove

michalc: Removed

docker-compose.yml
michalc: Yes... probably will at some point, but I think will leave that until later

PR #731, opened 2020-05-19
Add a management command to sync Metabase user permissions

allait:
When run using `django-admin sync_metabase_permissions` in one of
the environments the command will create a Metabase group for each
of the Data Workspace master datasets, assign permission for dataset
tables to that group and add any users that have dataset access as
members.

This requires a configured database connection and initial schema sync
to have finished in Metabase. The Metabase database name should match the
`Database.memorable_name` in Data Workspace (an alternative matching
based on host and DB name wouldn't allow us to use read-only replica in
Metabase for example). These could be created using the API, but since
we only have 1 DB at the moment that requires manual security group
configuration and the initial schema discovery is asynchronous it's much
easier to do this manually.

This also doesn't create users in Metabase at the moment, so only users
who have logged in at least once before will get dataset permissions.

The script pre-fetches datasets, tables and permissions so any changes
done in Data Workspace during the run shouldn't result in inconsistent
Metabase configuration.

It will also clean up any user groups (excluding the built-in ones)
and hide any tables that don't match current Data Workspace datasets.

The script is structured to use Metabase bulk API actions where possible
(like fetching all users and updating all group permissions with one
request), so it's not well suited for doing incremental updates after
individual user or dataset changes, but it could be restructured later
to enable this.

### Checklist


dataworkspace/dataworkspace/apps/applications/management/commands/sync_metabase_permissions.py
```diff
@@ -0,0 +1,269 @@
...
+            self.stdout.write(f"   Hiding Metabase table {table}")
+            metabase.update_table(table, visible=False)
+
+    def handle(self, *args, **options):
+        metabase = MetabaseClient()
+
+        datasets = (
+            MasterDataset.objects.live()
+            .filter(published=True)
+            .prefetch_related(
```
samuelhwilliams, line 236: 👍

PR #754, opened 2020-06-03
Add an endpoint to display embeded Metabase visualisation

allait:
### Description of change

This is a proof-of-concept endpoint that displays a full page
embed of a Metabase dashboard.

There are no permissions checks here at the moment as it's not
connected to a visualisation template, but a dashboard needs to
be first enabled for embedding from Metabase, so this shouldn't
expose any sensitive data.

It's using the existing visualisation wrapper template, so can be
switched to static header and footer in the future.

### Checklist


dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -1187,6 +1189,25 @@ def visualisation_publish_html_view(request, gitlab_project_id):
...
+        "exp": round(time.time()) + 600,
+    }
+    token = jwt.encode(payload, settings.METABASE_SECRET_KEY, algorithm="HS256")
+
+    return render(
+        request,
+        "running.html",
+        {
+            "visualisation_src": f"//{settings.METABASE_SITE_URL.rstrip('/')}/embed/dashboard/{token.decode('utf8')}#bordered=false&titled=false",
+            "wrap": "FULL_HEIGHT_IFRAME",
```
samuelhwilliams, line 1206: Should be `IFRAME_WITH_VISUALISATIONS_HEADER` I think? PoC ticket says to include the header.

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -1187,6 +1189,25 @@ def visualisation_publish_html_view(request, gitlab_project_id):
     return HttpResponse(status=405)
 
 
+@csp_update(FRAME_SRC=settings.METABASE_SITE_URL)
+def metabase_visualisation_embed_view(request, dashboard_id):
+    payload = {
+        "resource": {"dashboard": dashboard_id},
+        "params": {},
+        "exp": round(time.time()) + 600,
```
michalc, line 1197: If this is in seconds, then this feels a touch long? It only needs to be valid long enough for the response to make it back to the user's browser + and for it subsequent request for the iframe src to make it back to metabase?

[I guess just makes me nervous that it seems possible for a user to send out the iframe src to anyone with DW access, and they would be able to see the dashboard... but not sure what the plan is, auth-wise]

allait: That's what the sample snippet from Metabase is using, I'm not sure if there's any specific reason for it. I agree though, we could look into shortening that in the future.

I was a bit concerned about this as well, but thinking about it - at the point where the user is digging out the iframe src from the page source they could've shared a screenshot just as easily.

michalc: >[...] they could've shared a screenshot just as easily.

While this is always true... my impression of the vibe from above, is that while we can't ever stop screenshots/downloading + sharing them via another channel, we should prevent direct access to resources for those that shouldn't have it.

[Admittedly, not sure if this means that any action should be taken in this PR... but if there is some way to make sure that the iframe src only works for the current user, I think that would be good]

PR #755, opened 2020-06-03
Add endpoint for embedding a QuickSight dashboard

samuelhwilliams:
### Description of change
This is a proof-of-concept that shows an AWS QuickSight dashboard
embedded in our standard visualisation template. There are no
permissions checks here - this is just proving that the thing can be
done. I would expect us to mostly rip this code out at a later date.

### Show it
screencapture-dataworkspace-test-8000-datasets-b084809b-48fc-4830-987d-24810edef2f4-2020-06-04-11_31_36

screencapture-dataworkspace-test-8000-visualisations-quicksight-66795c77-ec51-4d3b-8175-347002aab7ee-2020-06-04-11_31_49

### Checklist


samuelhwilliams:
Hmm.. I chose a purposefully-obscure path somewhat on purpose to force us to come back to this and do it properly later. But it's probably not really much of a problem ... I can update it.

dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -360,6 +366,14 @@ def get_context_data(self, **kwargs):
             for data_link in data_links
         ]
 
+        quicksight_dashboard_id = self.request.GET.get("quicksight_dashboard_id", None)
```
allait, line 369: I'm not sure about this part, this feels like it's going beyond PoC and unlike the full page wrapper it's not really how we'd implement the actual integration where embed URL/ID would be stored on a model

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -92,3 +93,15 @@ def get_code_snippets(dataset: DataSet):
     sql_snippet = f"""SELECT * FROM "{schema}"."{table_name}" LIMIT 50"""
 
     return {"python": python_snippet, "r": r_snippet, "sql": sql_snippet}
+
+
+def get_quicksight_dashboard_name_url(dashboard_id):
```
allait, line 98: Could we move the view to `applications` to match the URL?

samuelhwilliams: Sure...

dataworkspace/dataworkspace/apps/datasets/views.py
samuelhwilliams: Mm it's definitely a hack. I'm not sure it matters to me at the moment that it's different from how we'd actually implement it. It just makes it temporarily possible to see how various dashboards might look on that page (FYI, IMHO, bad).

PR #800, opened 2020-06-15
(feat) Override the server header

michalc:
### Description of change

 The `server_tokens off;` turned off the nginx version from the server header, but it was still sent as `nginx`, and it comes up in ITHCs that we shouldn't be exposing the underlying server in headers.

### Checklist


samuelhwilliams:
Maybe we should just closed-source all of our code as well then.

michalc:
@samuelhwilliams Yes... I know...

PR #811, opened 2020-06-25
Allow access to Metabase embeded dashboards without VPN

allait:
### Description of change

This removes any /embed/ Metabase URLs from the VPN check, allowing
all Data Workspace users to access them

### Checklist

### DIFF
--- a/dataworkspace/proxy.py
+++ b/dataworkspace/proxy.py
@@ -194,6 +194,9 @@ async def async_main():
             and request.url.path == '/app/assets/img/site.webmanifest'
         )
 
+    def is_metabase_embed_requested(request):
+        return is_metabase_requested(request) and request.url.path.startswith('/embed/')
+
     def is_app_requested(request):
         return (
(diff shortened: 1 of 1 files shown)


dataworkspace/proxy.py
```diff
@@ -194,6 +194,9 @@ def is_metabase_manifest_requested(request):
             and request.url.path == '/app/assets/img/site.webmanifest'
         )
 
+    def is_metabase_embed_requested(request):
+        return is_metabase_requested(request) and request.url.path.startswith('/embed/')
```
michalc, line 198: Not sure if it's important, but in all cases, `is_metabase_requested(request)` will be `True` if `is_metabase_embed_requested` is called(?), so there there is an argument that its redundant.

allait: I parenthesized it in one place I use it, but if it's otherwise just chained to existing conditions as `and not ...` the it might also catch any `/embed` URLs that might exist on other apps

PR #819, opened 2020-06-29
Instrument sentry release tracking

samuelhwilliams:
### Description of change
Add support for sentry to track our releases so that it can
automatically attribute errors to commits/releases, which may help track
down issues over time. Our releases here are tagged by the commit ID
that is live, which should match up with the docker image tag as well.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/settings/base.py
+++ b/dataworkspace/dataworkspace/settings/base.py
@@ -290,7 +290,10 @@ S3_ROLE_PREFIX = env['S3_ROLE_PREFIX']
 YOUR_FILES_ENABLED = env.get('YOUR_FILES_ENABLED', 'False') == 'True'
 
 if env.get('SENTRY_DSN') is not None:
-    sentry_sdk.init(env['SENTRY_DSN'], integrations=[DjangoIntegration()])
+    sentry_kwargs = {"integrations": [DjangoIntegration()]}
+    if env.get('CONTAINER_TAG'):
+        sentry_kwargs['release'] = env.get('CONTAINER_TAG')
+    sentry_sdk.init(env['SENTRY_DSN'], **sentry_kwargs)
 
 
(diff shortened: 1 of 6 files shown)


dataworkspace/dataworkspace/settings/base.py
```diff
@@ -284,7 +284,10 @@ def aws_fargate_private_ip():
 YOUR_FILES_ENABLED = env.get('YOUR_FILES_ENABLED', 'False') == 'True'
 
 if env.get('SENTRY_DSN') is not None:
-    sentry_sdk.init(env['SENTRY_DSN'], integrations=[DjangoIntegration()])
+    sentry_kwargs = {"integrations": [DjangoIntegration()]}
+    if env.get('CONTAINER_TAG'):
+        sentry_kwargs = {"release": env.get('CONTAINER_TAG')}
```
niross, line 289: This will remove `integrations` from the kwargs key is that correct?

samuelhwilliams: Oof, no 😅

samuelhwilliams:
Just need to wait for @michalc who needs to run TF and update Jenkins env before we can merge this

samuelhwilliams:
@michalc 

I think it does need the env vars added to Jenkins before deploying, no?

I'll add the line to s3 - but yeah if it's unused it probably makes sense to rip it out at some point.

samuelhwilliams:
This is waiting on a fix from webops to allow dependanbot emails through sentry's (weirdly strict, in our case) email validation: https://uktrade.atlassian.net/browse/SRE-681

samuelhwilliams:
Webops have closed off their ticket to support this. It's a tangle of dependencies that goes through Sentry and back to Django's email validator, which rejects dependabot's email address for dependency bumps.

No support for this, for now.

PR #842, opened 2020-07-15
Auto sync quicksight perms

samuelhwilliams:
### Description of change
Now that we're rolling QuickSight out to more users (i.e. leaving the
prototype/discovery stage), the overhead and inconvenience of a dev
manually syncing Data Workspace dataset permissions is going to become
bothersome. This patch adds two automated sync steps:

1) Whenever a master dataset permission is changed (either
granting/revoking individual access, or changing the auth type from
authenticated to authorized), immediately kick off a background celery
job to sync the relevant permissions.
2) Run a nightly celery job to update all user permissions, in case any
of the real-time tasks fail or something hasn't been caught.

### Checklist


michalc:
I don't think it's changed in this PR, but to check, the users whose passwords last 31 days. What's the mechanism for removing them? I have two concerns

- They will be removed quickly because there is no tools/visualisation associated with them
- Or... they will never be removed (yes: they will expire, but never removed)

samuelhwilliams:
So there is a specific exception in the user cleanup tasks for quicksight DB users - so they don't get cleaned up at all right now. But unlike the tools/vis users, they have a static name, so it's not the case that we will have lots of them created - each quicksight author will have a single static DB user - though the password be changed frequently (whenever permissions are synced).

We could do with something to clean up DB users for people who no longer have QuickSight access though there's less urgency on that one.

PR #843, opened 2020-07-16
Implement sentry on the proxy

samuelhwilliams:
### Description of change
Ticket: https://trello.com/c/sYyEn1Sl/499

### Checklist


michalc:
I wonder if it's worth an integration test to make sure that the proxy still starts up with SENTRY_DSN set: if there is an exception thrown then nothing will work?

PR #853, opened 2020-07-22
Use immutable DB users for quicksight

samuelhwilliams:
At the moment we have a static DB user for quicksight and whenever we
grant new permissions we re-generate credentials for the same username
and push those to QuickSight. However, we didn't have an implementation
for revoking permissions. After a little time looking at adding more
thorough grant/revoke to the single mutable user, it felt easier to
instead go for immutable users that grant all the right permissions as a
one-off and then we simply discard that user and generate a new one if
we need a different set of permissions.

Our immutable quicksight users last 7 days rather than 31 for tool
users. This is for a few reasons:

1) We re-sync permissions for all QuickSight users every night, so
really they are unlikely to be used after 24 hours. 7 days gives us some
grace if this nightly job fails for some reason.
2) We can't tell when QuickSight credentials are in use, unlike with
tools - so rather than checking if the application is in use, we will
instead delete old quicksight users when their password expires. 7 days
means that there (hopefully) won't be too many old quicksight users
hanging around for too long.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -35,7 +35,7 @@ from dataworkspace.apps.core.utils import (
     stable_identification_suffix,
     source_tables_for_user,
     new_private_database_credentials,
-    persistent_postgres_user,
+    postgres_user,
 )
 from dataworkspace.apps.applications.gitlab import gitlab_has_developer_access
 from dataworkspace.apps.datasets.models import VisualisationCatalogueItem
@@ -459,7 +459,10 @@ def _do_delete_unused_datasets_users():
             cur.execute(
                 """
                 SELECT usename FROM pg_catalog.pg_user
-                WHERE valuntil != 'infinity' AND usename LIKE 'user_%' AND usename NOT LIKE '%_quicksight'
+                WHERE
+                (valuntil != 'infinity' AND usename LIKE 'user_%' AND usename NOT LIKE '%_qs')
+                OR
+                (valuntil != 'infinity' AND valuntil < now() AND usename LIKE 'user_%' AND usename LIKE '%_qs')
                 ORDER BY usename;
             """
             )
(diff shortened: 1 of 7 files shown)


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -459,12 +459,22 @@ def _do_delete_unused_datasets_users():
...
+            # Also delete expired quicksight credentials
+            cur.execute(
+                """
+                SELECT usename, valuntil FROM pg_catalog.pg_user
+                WHERE valuntil != 'infinity' AND valuntil < now() AND usename LIKE 'user_%' AND usename LIKE '%_qs'
+                ORDER BY usename, valuntil;
+            """
+            )
+            usenames.extend([result[0] for result in cur.fetchall()])
+
```
michalc, line 477: Extremely minor/style thing, but this could be part of the same query above, with an `OR` between the clauses?

dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -54,7 +55,10 @@ def db_role_schema_suffix_for_app(application_template):
 
 
 def new_private_database_credentials(
-    db_role_and_schema_suffix, source_tables, db_user, allow_existing_user=False
+    db_role_and_schema_suffix,
+    source_tables,
+    db_user,
+    valid_for=lambda: datetime.timedelta(days=31),
```
michalc, line 61: A minor style thing, but I would be tempted to not have a default value, and just make it required+explicit at each point of use

dataworkspace/dataworkspace/apps/applications/utils.py
samuelhwilliams: I didn't mind them split out too much, but it's not particularly complex so I'll combine them.

dataworkspace/dataworkspace/apps/core/utils.py
samuelhwilliams: Yeah that sounds sensible to be honest. Will change.

PR #856, opened 2020-07-24
Launch QuickSight button

samuelhwilliams:
### Description of change
Add a "Launch QuickSight" button to the tools page. This button has
three states:

1) Hidden - if the Django-Waffle switch is disabled (the default)
2) Request Access - if the user doesn't have a specific
'access_quicksight' permission. The user will have to submit an access
request, whereupon we can grant the SSO profile permission and toggle
the user flag in django-admin. Gives us a light touch point to
understand who's asking for access and why.
3) Launch QuickSight - when the user has the access_quicksight
permission.

Ticket: https://trello.com/c/5fLTpMx8/603

### Checklist


dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -173,9 +173,11 @@ def link(application_template):
...
         return f'{request.scheme}://{app}-{sso_id_hex_short}.{settings.APPLICATION_ROOT_DOMAIN}/'
 
-    has_any_tool_perms = request.user.has_perm(
-        'applications.start_all_applications'
-    ) or request.user.has_perm('applications.access_appstream')
+    has_any_tool_perms = (
+        request.user.has_perm('applications.start_all_applications')
+        or request.user.has_perm('applications.access_appstream')
+        or request.user.has_perm('applications.access_quicksight')
+    )
```
michalc, line 180: Maybe not for now, but will the above reveal all the other tools if the use _just_ has QuickSight access and not the tools in general, but they then won't be able to start them?

michalc: Ignore me: I'm wrong

PR #862, opened 2020-07-30
Loop over fields via reference dataset rather than record.

niross:
### Description of change

Trello card: https://trello.com/c/Uk5Shu5p/480-improve-loading-time-on-reference-dataset-pages

Get reference dataset fields via the reference dataset model rather than through the record itself when looping through records. This was causing extra queries on each record (thousands in total) when looping through a recordset.

Before:
ui-ref-datasets-before

After:
ui-ref-datasets-after

### Checklist


samuelhwilliams:
That is a solid improvement 👍  Nice work

PR #863, opened 2020-07-30
(feat) Use EFS access point as home directory if one exists

michalc:
### Description of change

The aim of this is to increase the available space in user's home directories, which are current synced to and from S3 using the "s3sync" sidecar container. This syncing still takes place, but now instead of syncing with a local filesystem, it syncs with an EFS filesystem.

It is very deliberate to only do this if a user has an access point configured, to have a slow rollout, with the possibility of going back for any given user.

Output of `df` on JupyterLab Python

Screenshot 2020-08-24 at 12 27 19

### Checklist


dataworkspace/dataworkspace/apps/accounts/admin.py
```diff
@@ -86,6 +92,13 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         instance = kwargs['instance']
 
+        self.fields[
+            'home_directory_efs_access_point_id'
+        ].initial = instance.profile.home_directory_efs_access_point_id
+        self.fields['home_directory_efs_access_point_id'].widget.attrs[
+            'class'
+        ] = 'vTextField'
```
samuelhwilliams, line 100: Was looking into why some text fields get this automatically and this is the only place that we have to add this class explicitly... Based on [this](https://stackoverflow.com/questions/2038292/django-why-does-my-charfield-not-get-given-the-class-vtextfield), we can probably drop this snippet and instead set the `widget` on the field to `AdminTextInputWidget`?

E.g.
```python
    home_directory_efs_access_point_id = forms.CharField(
        label='Home directory ID',
        help_text='EFS Access Point ID',
        max_length=128,
        required=False,
        widget=AdminTextInputWidget,
    )
```

samuelhwilliams (approved):

dataworkspace/dataworkspace/apps/accounts/admin.py
michalc: Done

PR #865, opened 2020-07-31
Centralise sentry config

samuelhwilliams:
### Description of change
Move the sentry setup code from Django settings and the proxy to a
separate file so that we can share code/config.

Group CancelledErrors together so that we only have a single event for
them in Sentry, making them less noisy and easier to handle. Same for
ConnectionResetError, which is also fairly prolific in the proxy and
pops up in various places with the same outcome.

### Checklist

### DIFF
--- a/dataworkspace/sentry.py
+++ b/dataworkspace/sentry.py
@@ -1,0 +1,28 @@
+import os
+from asyncio import CancelledError
+
+import sentry_sdk
+
+
+def before_send(event, hint):
+    if 'exc_info' not in hint:
+        return event
+
+    exception = hint['exc_info'][1]
+
+    # Group together all asyncio/aiohttp cancellederrors
+    if isinstance(exception, CancelledError):
+        event['fingerprint'] = ['cancelled-error', '{{ module }}']
+    elif isinstance(exception, ConnectionResetError):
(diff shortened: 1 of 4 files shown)


dataworkspace/sentry.py
```diff
@@ -0,0 +1,28 @@
...
+from asyncio import CancelledError
+
+import sentry_sdk
+
+
+def before_send(event, hint):
+    if 'exc_info' not in hint:
+        return event
+
+    exception = hint['exc_info'][1]
```
sekharpanja, line 11: Wondering it might a good idea to check `hint['exc_info']` has items or use `.get()`

samuelhwilliams: This is a copy/paste from the Sentry recommended implementation here: https://docs.sentry.io/data-management/event-grouping/sdk-fingerprinting/

```python
class DatabaseConnectionError(Exception):
    pass

def before_send(event, hint):
    if 'exc_info' not in hint:
        return event

    exception = hint['exc_info'][1]
    if isinstance(exception, DatabaseConnectionError):
        event['fingerprint'] = ['database-connection-error']

    return event

sentry_sdk.init(..., before_send=before_send)
```

It feels like we should be able to trust that this exists... but let's bullet-proof it anyway as you recommend.

PR #880, opened 2020-08-03
Sleep when attempting to create new IAM credentials for running tools

abbas123456:
### Description of change

When a user accesses tools for the first time, IAM credentials need to be created but as the various IAM operations are eventually consistent, this may fail. This is currently attempted 3 times but it will be useful to add a sleep for 5 seconds in between attempts to reduce the chance of this failing.

Sentry error where this was picked up: https://sentry.ci.uktrade.digital/dit/data-workspace/issues/28536/

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/api_v1/views.py
+++ b/dataworkspace/dataworkspace/apps/api_v1/views.py
@@ -2,6 +2,7 @@ import datetime
 import json
 import logging
 import re
+import time
 
 import boto3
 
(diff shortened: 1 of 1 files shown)


dataworkspace/dataworkspace/apps/api_v1/views.py
```diff
@@ -304,6 +305,8 @@ def aws_credentials_api_GET(request):
             if i == max_attempts - 1:
                 raise
         else:
+            # To try compensating for eventual consistency
+            time.sleep(5)
```
michalc, line 309: I'm not sure if gevent monkey patches time.sleep... if it doesn't, this could block the event loop for 5 seconds. Can this be gevent.sleep just in case?

(Also: 5 seconds feels a bit long during a request. Can it be 1 second?)

PR #883, opened 2020-08-05
Feature/handle non existent application

abbas123456:
### Description of change

At the moment, accessing a tool or visualisation that doesn't exist raises an exception that isn't handled and this is logged to Sentry. This PR catches the exception and returns a 403 to the user if this happens. 

This PR also changes the logging behaviour in the proxy to only log UserExceptions to sentry if they are 500's

### Checklist

### DIFF
--- a/dataworkspace/proxy.py
+++ b/dataworkspace/proxy.py
@@ -314,23 +314,21 @@ async def async_main():
                 return await handle_metabase(downstream_request, method, path, query)
             return await handle_admin(downstream_request, method, path, query)
         except Exception as exception:
-            logger.exception(
-                'Exception during %s %s %s',
-                downstream_request.method,
-                downstream_request.url,
-                type(exception),
-            )
+            user_exception = isinstance(exception, UserException)
+            if not user_exception or (user_exception and exception.args[1] == 500):
+                logger.exception(
+                    'Exception during %s %s %s',
+                    downstream_request.method,
+                    downstream_request.url,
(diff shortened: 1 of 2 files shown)


dataworkspace/proxy.py
```diff
@@ -314,23 +314,21 @@ async def handle(downstream_request):
...
-                type(exception),
-            )
+            user_exception = isinstance(exception, UserException)
+            if not user_exception or (user_exception and exception.args[1] == 500):
+                logger.exception(
+                    'Exception during %s %s %s',
+                    downstream_request.method,
+                    downstream_request.url,
+                    type(exception),
+                )
```
samuelhwilliams, line 324: I think this only affects logging behaviour, not reporting to Sentry, so I'm not sure this accomplished the goal of "This PR also changes the logging behaviour in the proxy to only log UserExceptions to sentry if they are 500's". Unless I misunderstand?

michalc: Ah this may be my fault: apologies if this is the case. For some reason I thought Sentry integrates with logger.exception and adds "send to Sentry" behaviour, but perhaps it doesn't...

michalc: Although I do see things like `Exception during GET ...` in the Sentry exception reports, so I now think it does...

samuelhwilliams: Huh, looks like you're right! TIL.

PR #884, opened 2020-08-06
Use <td> for empty table cell instead of <th>

abbas123456:
### Description of change

The wave tool has picked up an empty table header cell as an accessibility error, this PR replaces the `<th>` with a `<td>`

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/templates/files.html
+++ b/dataworkspace/dataworkspace/templates/files.html
@@ -224,7 +224,7 @@ Modified by the Department for International Trade
                 <table class="govuk-table" style="table-layout: fixed;">
                     <thead>
                         <tr class="govuk-table__row">
-                            <td scope="col" class="govuk-table__header govuk-table__header--checkbox"></td>
+                            <td class="govuk-table__header govuk-table__header--checkbox"></td>
                             <th scope="col" class="govuk-table__header">Name</th>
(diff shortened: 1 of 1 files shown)


dataworkspace/dataworkspace/templates/files.html
```diff
@@ -224,7 +224,7 @@ <h1 class="govuk-heading-l">Your files</h1>
                 <table class="govuk-table" style="table-layout: fixed;">
                     <thead>
                         <tr class="govuk-table__row">
-                            <th scope="col" class="govuk-table__header govuk-table__header--checkbox"></th>
+                            <td scope="col" class="govuk-table__header govuk-table__header--checkbox"></td>
```
michalc, line 227: Small thing: should the scope tag be removed if it's no longer a header? (Since it was empty perhaps it never really made sense anyway?)

abbas123456: Yes should definitely be removed as its not a valid attribute for `<td>`, good spot.

abbas123456: Will also change the class, should be `govuk-table__cell govuk-table__cell--checkbox"` i guess

abbas123456: Actually that breaks the page completely, will keep the class as it is

PR #889, opened 2020-08-12
Keep Your Files modal HTML out of the DOM until it's needed

michalc:
### Description of change

This is to make the Your Files page more reasonable when CSS is disabled. Note that the modal HTML is there is Javascript is disabled, since the ng-if will remove elements using Javascript, but in this case, the whole interface won't work anyway.

There is some duplication here, but given the number of modals, the current uncertainty regarding the page, and the fact that to address this I suspect will have to do a lot of faffing with templates and/or transclusion that can get a bit complicated, suspect not worth it.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Specified sizes on SVGs in the Your Files page to make it more reasonable without CSS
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Just trying this out with CSS disabled, I can't seem to dismiss the "deleted file" modal (though I can use it just fine).

michalc:
For posterity, from offline discussion, it looks like if CSS is disabled after page load, then the modal contents won't disappear from the DOM, since the check for CSS being enabled happens on page load.

Leaving as is, since for now at least, this is acceptable.

PR #892, opened 2020-08-13
fix: stop using env vars for tool visibility

samuelhwilliams:


dataworkspace/dataworkspace/apps/applications/management/commands/ensure_application_template_models.py
```diff
@@ -39,7 +40,6 @@ def handle(self, *args, **options):
             )
             try:
                 ApplicationTemplate.objects.create(
-                    visible=desired_application_template['VISIBLE'] == 'True',
```
michalc, line 42: I think the default visibility in the model is `True`: so I think either that should be changed or this line kept?

samuelhwilliams: Ah yep - have set it to false. 👍

PR #894, opened 2020-08-14
fix: Drop objects owned by a db user before dropping user

niross:
Fixes issue where dropping a user with existing privileges was failing with the error "role cannot be dropped because some objects depend on it"

Closes: https://trello.com/c/4KN0gY7O/717-unable-to-delete-users-that-have-db-privileges-set

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -568,6 +568,11 @@ def _do_delete_unused_datasets_users():
                             'delete_unused_datasets_users: dropping user %s', usename
                         )
                         cur.execute(
+                            sql.SQL('REASSIGN OWNED BY {} TO current_user;').format(
+                                sql.Identifier(usename)
+                            )
+                        )
+                        cur.execute(
                             sql.SQL('DROP OWNED BY {};').format(sql.Identifier(usename))
                         )
                         cur.execute(
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -568,6 +568,9 @@ def _do_delete_unused_datasets_users():
                             'delete_unused_datasets_users: dropping user %s', usename
                         )
                         cur.execute(
+                            sql.SQL('DROP OWNED BY {};').format(sql.Identifier(usename))
+                        )
+                        cur.execute(
                             sql.SQL('DROP USER {};').format(sql.Identifier(usename))
                         )
                     except redis.exceptions.LockError:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 2020-08-14
+
+### Changed
+
+- Before dropping a db user drop any objects that are owned by them
+
 ## 2020-08-13
 
 ### Changed


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -567,6 +567,9 @@ def _do_delete_unused_datasets_users():
                         logger.info(
                             'delete_unused_datasets_users: dropping user %s', usename
                         )
+                        cur.execute(
+                            sql.SQL('DROP OWNED BY {};').format(sql.Identifier(usename))
```
samuelhwilliams, line 571: Do we instead want to transfer these resources to the persistent user role instead of dropping them? I'm not clear on what these might be - created tables I guess? I think I'm a bit uncomfortable just deleting them.

niross: Yeah understandable, I can look into it, my thinking was as these are "unused" (https://github.com/uktrade/data-workspace/blob/319fd3cc1d6e82ed31a222c06f02ec638009fd15/dataworkspace/dataworkspace/apps/applications/utils.py#L489-L491) users they should have no tables. This should just drop any remaining privileges granted when the app opens.

samuelhwilliams: Ah that comment confuses me a bit. I think it means currently-unused users, not never-used users. So it's a combination of users that never got used and ones from applications that were running/used but have now closed.

niross: Ok yeah I think I was confused by the wording. As discussed it now reassigns objects before dropping.

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -567,6 +567,14 @@ def _do_delete_unused_datasets_users():
                         logger.info(
                             'delete_unused_datasets_users: dropping user %s', usename
                         )
+                        cur.execute(
+                            sql.SQL('REASSIGN OWNED BY {} TO current_user;').format(
```
samuelhwilliams, line 571: I think this will reassign the user's resources to the data workspace admin user/role right? We probably want to reassign the user's resources to their permanent role.

For context - users get given a single permanent database role when they start up any tool, which is reused across all tools. But each tool startup also creates a temporary db user, which is granted that role, for that tool session. So to persist any created resources I think we want to transfer ownership away from their temporary user and to their permanent role. Michal might be able to correct me if I'm wrong about anything there...

niross: Got it, thanks. I'll update

michalc: > So to persist any created resources I think we want to transfer ownership away from their temporary user and to their permanent role.

Yes... _ideally_ though, we would do this in real time. There is a trigger that does this (/ _should_ do this) https://github.com/uktrade/data-workspace/blob/4c82bde924d4bde881d567a77fa93190c165e8d3/dataworkspace/dataworkspace/apps/core/utils.py#L137

Say, if someone creates a table in pgAdmin, it should be owned by the permanent role to be used in RStudio immediately. I would probably be pro investigating sorting out the trigger to cover more object types. I think I might even pro _not_ silently re-assigning the objects in the cleanup task, exactly for us to be prodded that something with the trigger isn't working.

dataworkspace/dataworkspace/apps/applications/utils.py
michalc: > [...] before dropping.

To confirm: the cleanup shouldn't drop objects at all? At most, it should re-assign? (Although as per the comment above, ideally, done in real time beforehand)

michalc: > I think it means currently-unused users, not never-used users. So it's a combination of users that never got used and ones from applications that were running/used but have now closed.

Oh... yes, can see it's confusing. The comment was more in response to "Why not just use the users in ApplicationInstanceDbUsers?" where the answer would be "We _also_ want to be able to delete the users that never made it into ApplicationInstanceDbUsers"

dataworkspace/dataworkspace/apps/applications/utils.py
niross: Ok thanks for clearing that up :+1:  I will close this and create a ticket to investigate possible updates to the trigger

PR #900, opened 2020-08-19
Feature/add state filter to applicationinstances admin

abbas123456:
### Description of change

Sentry errors related to tools/visualisations sometimes have the application url in the stacktrace but no user email address and sometimes it's necessary to contact the user to understand what they did to trigger the error.

The user of a **running** application can be found by going to /admin/applications/applicationinstance and looking at the `PUBLIC HOST` field, but as a sentry error usually relates to a tool that has crashed, the row won't be visible.

This PR adds the state filter to the Applicationinstances admin page to allow an easy search for the user's email address when the Applicationinstance is in a `STOPPED` state.

There is also a small refactor as I noticed that the application/visualisation URL is generated using a hex digest of the user's sso id but this is repeated in loads of different places.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/admin.py
+++ b/dataworkspace/dataworkspace/apps/applications/admin.py
@@ -36,8 +36,9 @@ from dataworkspace.apps.eventlog.utils import log_permission_change
 
 @admin.register(ApplicationInstance)
 class ApplicationInstanceAdmin(admin.ModelAdmin):
-
-    list_display = ('owner', 'public_host', 'created_date')
+    list_display = ('owner', 'public_host', 'created_date', 'state')
+    list_filter = ('state',)
+    ordering = ('-created_date',)
     fieldsets = [
         (
             None,
@@ -64,10 +65,6 @@ class ApplicationInstanceAdmin(admin.ModelAdmin):
 
     def has_add_permission(self, request):
         return False
-
-    def get_queryset(self, request):
-        qs = super().get_queryset(request)
-        return qs.filter(state__in=['SPAWNING', 'RUNNING'])
 
     def max_cpu(self, obj):
         try:
--- a/dataworkspace/dataworkspace/apps/core/utils.py
+++ b/dataworkspace/dataworkspace/apps/core/utils.py
@@ -46,7 +46,7 @@ def postgres_user(stem, suffix=''):
 
 
 def db_role_schema_suffix_for_user(user):
-    return hashlib.sha256(str(user.profile.sso_id).encode('utf-8')).hexdigest()[:8]
+    return stable_identification_suffix(str(user.profile.sso_id), short=True)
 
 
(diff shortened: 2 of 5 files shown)


dataworkspace/dataworkspace/apps/applications/admin.py
```diff
@@ -36,8 +36,9 @@
 
 @admin.register(ApplicationInstance)
 class ApplicationInstanceAdmin(admin.ModelAdmin):
-
-    list_display = ('owner', 'public_host', 'created_date')
+    list_display = ('owner', 'public_host', 'created_date', 'state')
+    list_filter = ('state',)
```
abbas123456, line 40: The new page will show all ApplicationInstances by default so I'm not sure if this will have an impact on whoever uses this page normally. They would need to filter by "Running" or "Spawning" to get the view they used to have.

samuelhwilliams: Is there a way to set the default filter as "Running"?

abbas123456: There is, but it involves having to write a custom `SimpleListFilter` as shown here: https://stackoverflow.com/questions/851636/default-filter-in-django-admin. I wasn't sure if the problem I'm addressing in the PR merits the extra 20-30 lines and confusion :)

abbas123456: Although, having looked at the SO link again, the accepted answer is much shorter than the first answer so it might not be too bad.

samuelhwilliams: Probably not :)

dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -652,5 +650,8 @@ def __init__(self, *args, **kwargs):
             connection.close_if_unusable_or_obsolete()
 
 
-def stable_identification_suffix(identifier):
-    return hashlib.sha256(identifier.encode('utf-8')).hexdigest()[:8]
+def stable_identification_suffix(identifier, short=True):
+    digest = hashlib.sha256(identifier.encode('utf-8')).hexdigest()
+    if short:
+        return digest[:8]
+    return digest
```
samuelhwilliams, line 657: Just a minor (optional) nit .. It might be nice to remove the default from `short` here so that it's explicit in each case what we're going for?

samuelhwilliams: Agree with this refactor in general though 👍

samuelhwilliams (approved):

dataworkspace/dataworkspace/apps/core/utils.py
abbas123456: Yep, I can do that

PR #903, opened 2020-08-20
feature: add preview page for datacut datasets

kristof-bb:
On the datacut detail page, there is the option to get a preview of the data. This contains:
- the datacut fields
- 10 random rows of sample data
- a download button to download the data as a csv

### Description of change

image

### Checklist


dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -465,6 +465,9 @@ def get_google_data_studio_link(self):
     def can_show_link_for_user(self, user):
         return False
 
+    def type(self):
+        return DataLinkType.SOURCE_TABLE.value
```
samuelhwilliams, line 469: Would these be better as with a `@property`' decorator?

dataworkspace/dataworkspace/settings/base.py
```diff
@@ -327,6 +327,7 @@ def aws_fargate_private_ip():
 SEARCH_RESULTS_DATASETS_PER_PAGE = 15
 
 REFERENCE_DATASET_PREVIEW_NUM_OF_ROWS = 1000
+DATACUT_DATASET_PREVIEW_NUM_OF_ROWS = 10
```
samuelhwilliams, line 330: Is there a reason these are so different? Not sure I understand why we'd want to show 1000 reference rows but only 10 datacut rows...

dataworkspace/dataworkspace/templates/datasets/data_cut_dataset.html
```diff
@@ -122,6 +125,18 @@ <h2 class="govuk-heading-l govuk-!-margin-top-8">Data Links</h2>
                             {{ data_link.frequency }}
                           {% endif %}
                         </td>
+                        {% if has_access %}
+                        <td class="govuk-table__cell">
+                            {% if can_show_link and data_link.type == 3 %}
```
samuelhwilliams, line 130: Is it worth making the `== 3` check a bit more explicit somehow? Maybe add the data link type enum to template context?

dataworkspace/dataworkspace/settings/base.py
niross: Maybe we could make these settable via env vars so we can test against them?

dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -534,6 +536,33 @@ def write(self, value):
     return response
 
 
+def get_random_data_sample(database, query, sample_size=10, batch_size=1000):
+    query_timeout = 300 * 1000
+
+    with connect(database_dsn(settings.DATABASES_DATA[database])) as conn, conn.cursor(
+        name='data_preview'
+    ) as cur:  # Named cursor => server-side cursor
```
michalc, line 544: If we will always have a fairly small sample size, and we're always putting all the results in memory anyway, I wonder if no need for a cursor? We can just do a regular query?

I think also, if you use this function with sample_size > batch_size, then it won't work as expected?

(I might also lean to not have defaults for the parameters to this function if we're not using them... but that is a bit of a style thing)

michalc: Oh! Ignore me, I see the queries don't have limits.

michalc: (And in that case... it's a cunning way to make what appears to be a random preview of the results of an arbitrary query...)

PR #914, opened 2020-08-24
fix: address/ignore pylint errors

samuelhwilliams:
### Description of change
Either fix or ignore all our existing pylint errors, move pylint config to a dedicated file, and add pylint checks to CI tests.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/forms.py
+++ b/dataworkspace/dataworkspace/apps/datasets/forms.py
@@ -1,13 +1,20 @@
 from django import forms
-from .models import SourceTag
+
 from dataworkspace.apps.datasets.constants import DataSetType
+from .models import SourceTag
 
 
 class FilterWidget(forms.widgets.CheckboxSelectMultiple):
     template_name = 'datasets/filter.html'
     option_template_name = "datasets/filter_option.html"
 
-    def __init__(self, group_label, hint_text=None, *args, **kwargs):
+    def __init__(
+        self,
+        group_label,
+        hint_text=None,
+        *args,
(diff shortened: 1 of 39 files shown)


dataworkspace/dataworkspace/apps/datasets/forms.py
```diff
@@ -1,13 +1,14 @@
...
 from dataworkspace.apps.datasets.constants import DataSetType
+from .models import SourceTag
 
 
 class FilterWidget(forms.widgets.CheckboxSelectMultiple):
     template_name = 'datasets/filter.html'
     option_template_name = "datasets/filter_option.html"
 
-    def __init__(self, group_label, hint_text=None, *args, **kwargs):
+    def __init__(self, group_label, *args, hint_text=None, **kwargs):
```
michalc, line 11: I get a bit scared of changes like this (and similar argument/keyword arguments changes elsewhere in this PR). For me: they often they never quite behave as I expect... But perhaps my mental model of how arguments work in Python isn't quite right

I wouldn't be against splitting these out and deploying one-ish at a time to keep the risk of multiple things going wrong in production at the same time down.

michalc (approved):

dataworkspace/dataworkspace/apps/datasets/forms.py
samuelhwilliams: Yeah it could have broken things, although all of our usages for `FilterWidget` were fine as it happens.

samuelhwilliams: Have removed the signature changes for now.

PR #916, opened 2020-08-24
footer is updated to match the new design

sekharpanja:
### Description of change

Footer is updated to match the new design as per https://trello.com/c/00IJSmmr

image

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/static/data-workspace.css
+++ b/dataworkspace/dataworkspace/static/data-workspace.css
@@ -1,6 +1,9 @@
 @import 'app-tabs.css';
 @import 'app-copy.css';
 
+.govuk-footer__inline-list--spaced{
+  padding-bottom: 70px;
+}
 .govuk-panel--confirmation--left {
   text-align:left;
 }
--- a/dataworkspace/dataworkspace/static/data-workspace.css
+++ b/dataworkspace/dataworkspace/static/data-workspace.css
@@ -1,9 +1,6 @@
 @import 'app-tabs.css';
 @import 'app-copy.css';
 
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/static/data-workspace.css
```diff
@@ -1,6 +1,9 @@
 @import 'app-tabs.css';
 @import 'app-copy.css';
 
+.govuk-footer__inline-list--spaced{
+  padding-bottom: 70px;
+}
```
samuelhwilliams, line 6: Do we need this custom styling? I'm not sure I understand what purpose it's fills. I would think our footer is just a standard GOV.UK Design System footer - but happy to be proven wrong.

If we do need custom styling, I think we should rename this class to `app-footer__inline-list--spaced` to make it clear that it's not part of the Design System styles ("Small modifications to components" at https://design-system.service.gov.uk/get-started/extending-and-modifying-components/).

sekharpanja: I will get this across with Bo, if he is happy without that spacing. In that case, we don't need to make any changes to gov uk footer styling.

michalc: In case the space remains: typically it would be changed (and typically, lower) on mobile view, so may need a media query for that.

sekharpanja: Clarified with Bo and removed that custom styling.

kristof-bb (approved):

samuelhwilliams (approved):

PR #917, opened 2020-08-24
fix: preview design updates

kristof-bb:
design changes:
- back button on preview page instead of breadcrumbs
- no vertical scrollbar for 10 or less preview records
- truncate preview results if more then 100 chars

### Description of change

image

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add weighting to dataset search
 - Allow for searching of datasets by source tag name
 
+### Changed
+
+- Datacut preview page design updates
+
 ## 2020-08-20
 
 ### Added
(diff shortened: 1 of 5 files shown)


samuelhwilliams:
Hmm, this is maybe more of a design challenge, but I'm not sure about back links on this page ... Reading the [design system back links page](https://design-system.service.gov.uk/components/back-link/), it does mention that services should either use breadcrumbs or back links. I asked on the X-GOV Design System channel and someone suggested that back links are more for linear transactional services, and breadcrumbs are for more hierarchical websites (like ours) ... so I'd be a bit tempted to push back on this. Particularly as I think this is the only place we'd have a back link and everywhere else has breadcrumbs, it feels like it reduces navigation consistency which would/could confuse users more?

samuelhwilliams (approved):
Dev changes look good otherwise. Would be interested to hear any outcome of above discussion before this is merged though.

PR #921, opened 2020-08-27
fix: force a cache bust of site CSS to avoid strange behaviour on front page

michalc:
### Description of change

Suspect that the changes at the bottom of the CSS file, `.app-js-hidden` and/or `.hidden` are not visible in browsers due to caching

### Checklist


michalc:
Changed back to draft since no longer sure it's the cause of the issue

samuelhwilliams (approved):
This works as a quick fix. A slightly easier fix might be to just throw a query param on the end. An automatic option might be to hash the file and throw that on the end of the URL as a query param?

Django also has some static files hash manifest thing but I haven't used that before.

Thanks for fixing.

michalc:
> A slightly easier fix might be to just throw a query param on the end.

Yes... although (and I might be out of date) I though some browsers then never cache results of HTTP requests with query string. Which would then fix the issue, but then the CSS won't ever be cached, and so re-requested on every page load

> Django also has some static files hash manifest thing but I haven't used that before.

I suspect something along these lines would be good

PR #924, opened 2020-08-28
fix: sys.exit call, healthcheck failure and slugishness

michalc:
### Description of change

During multiple parallel downloads there are three issues that this change addresses, at least partially.

- sys.exit is called, interrupting all connections on that process, which is 50% of all connections to the admin applicaiton, since we have two processes in total.
- The ALB healthcheck fails, removing the instance from the load balancer entirely.
- Sluggishness when browsing the site when there are concurrent downloads

#### Regarding sys.exit
The gunicorn arbiter process appears to think the worker process is unresponsive during parallel downloads, and then kills the worker, interrupting all current connections. The exact cause is unknown. However, testing does suggest the event loop isn't being blocked by anything in Django. Possible it's related to the filesystem blocking the heartbeat process, as describe at https://docs.gunicorn.org/en/stable/faq.html#how-do-i-avoid-gunicorn-excessively-blocking-in-os-fchmod, but admittedly this is just a theory.

However, we only ever have a single worker process, so gunicorn's system of starting and monitoring processes, in addition to possibly being a bit broken in our case, seems unnecessary. It's crossed my mind that killing and re-starting the worker process could contribute to the database connection leak that sometimes happens, although no direct evidence of this has been seen.

#### Regarding the failing ALB healthcheck

From logs, it looks like the ALB healthcheck times out during the parallel downloads when CPU hit 100%. While this change I suspect won't stop the issue 100%, when testing this PR, so far parallel downloads have not caused the instance to be removed due to a failing healthcheck.

#### Regarding slugishness

The most notable change happened when increasing the chunk size from the default of 1kb to 64kb.

#### Performance

The gevent pywgi server is used in the DIT S3Proxy, where large (and small) files are streamed, and no performance issues have been reported. Also, the gevent server does not kill idle persistant connections (from the proxy), but the aiohttp client in the proxy (by default) kills such connections after 15 seconds, which I think _could_ address closing transport errors occasionally seen from the proxy, since the client would be in control of closing connections.

It does look like the server is a bit slower to start accepting connections, hence the `sentry_requests.clear()` to clear exceptions while the test is repeatedly attempting to call the healthcheck. This is a bit unfortunate, however, I suspect this wouldn't cause real-world impact.

### Checklist


test/test_application.py
```diff
@@ -1148,6 +1148,9 @@ async def test_application_shows_forbidden_if_not_auth_ip(self):
 
         await until_succeeds('http://dataworkspace.test:8000/healthcheck')
 
+        await asyncio.sleep(1)
```
samuelhwilliams, line 1151: Can you explain a bit more what this is for? Why do we wait for after the healthcheck has passed, and then a bit more, to clear sentry requests?

michalc: Done

samuelhwilliams (approved):

niross (approved):

PR #935, opened 2020-09-01
Preview filter numbers

samuelhwilliams:
### Description of change
Appends search result numbers to the filters on the homepage, showing users how many results will be available if that filter option is toggled on/off.

Adds a functional test using `pyppeteer` to launch chromium and interact with the site. This is the first browser-based test we have, so would be interesting to get some feedback on implementation/choice as this could be expanded/reused for any smoke tests we want to run on the live service. The test itself is quite specific in order to test variations of filters against datasets - open to ways to increase the flexibility and range of the test.

Includes a minor HTML semantic fix to make sure that filter hints have unique IDs rather than a strangely-hardcoded one, and another fix to properly filter unpublished datasets for admin users.

Ticket: https://trello.com/c/ZEyoaZXo/792

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/views.py
+++ b/dataworkspace/dataworkspace/apps/datasets/views.py
@@ -118,10 +118,10 @@ def filter_datasets(
                 datacut_type
             )
 
-            if user.has_perm(master_perm) and (not use or str(master_type) in use):
+            if user.has_perm(master_perm) and (not use or master_type in use):
                 dataset_filter |= Q(published=False, type=master_type)
 
-            if user.has_perm(datacut_perm) and (not use or str(datacut_type) in use):
+            if user.has_perm(datacut_perm) and (not use or datacut_type in use):
                 dataset_filter |= Q(published=False, type=datacut_type)
 
     datasets = datasets.filter(dataset_filter).annotate(
--- a/dataworkspace/dataworkspace/apps/datasets/views.py
+++ b/dataworkspace/dataworkspace/apps/datasets/views.py
@@ -5,7 +5,7 @@ import csv
 import io
 from itertools import chain
 import json
-from typing import Union
+from typing import Set
 
 import boto3
 from botocore.exceptions import ClientError
@@ -13,12 +13,23 @@ from csp.decorators import csp_update
 from psycopg2 import sql
 from django.conf import settings
 from django.contrib.postgres.aggregates.general import ArrayAgg
+from django.contrib.postgres.fields import ArrayField
 from django.contrib.postgres.search import SearchQuery, SearchRank, SearchVector
 from django.core import serializers
 from django.core.exceptions import PermissionDenied
 from django.core.paginator import Paginator
 from django.core.serializers.json import DjangoJSONEncoder
-from django.db.models import CharField, F, IntegerField, Q, Value
+from django.db.models import (
+    F,
+    IntegerField,
+    Q,
+    Value,
+    Case,
+    When,
+    BooleanField,
+    UUIDField,
+    QuerySet,
+)
 from django.http import (
     Http404,
     HttpResponse,
@@ -76,25 +87,21 @@ from dataworkspace.zendesk import (
     create_support_request,
     get_people_url,
 )
-from .models import SourceTag
 
 
-def filter_datasets(
-    datasets: Union[ReferenceDataset, DataSet],
-    query,
-    access,
-    source,
-    use=None,
-    user=None,
+def preprocess_datasets(
+    datasets: QuerySet, query, use=None, user=None,
 ):
-    search = (
-        SearchVector('name', weight='A', config='english')
-        + SearchVector('short_description', weight='B', config='english')
-        + SearchVector('source_tags__name', weight='B', config='english')
-    )
-    search_query = SearchQuery(query, config='english')
+    """
+    Filters the dataset queryset for:
+        1) visibility (whether the user can know if the dataset exists)
+        2) matches the search terms
 
-    dataset_filter = Q(published=True)
+    Annotates the dataset queryset with:
+        1) `has_access`, if the user can use the dataset's data.
+    """
+    # Filter out datasets that the user is not allowed to even know about.
+    visibility_filter = Q(published=True)
 
     if user:
         if datasets.model is ReferenceDataset:
@@ -104,7 +111,7 @@ def filter_datasets(
             )
 
             if user.has_perm(reference_perm):
-                dataset_filter |= Q(published=False)
+                visibility_filter |= Q(published=False)
 
         if datasets.model is DataSet:
             master_type, datacut_type = (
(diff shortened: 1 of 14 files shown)


dataworkspace/dataworkspace/tests/datasets/test_views.py
```diff
@@ -372,12 +381,14 @@ def test_find_datasets_filters_visualisations_by_use(client):
             'short_description': vis.short_description,
             'source_tag_ids': mock.ANY,
             'purpose': DataSetType.VISUALISATION.value,
+            'has_access': True,
         },
     ]
 
 
 def test_find_datasets_filters_by_source(client):
```
michalc, line 389: Is it worth an assertion on the number of queries, e.g. using assertNumQueries / django_assert_num_queries or similar? The main thing I think would be to somehow assert that there _isn't_ a query per result(?)

samuelhwilliams: Sounds like a good idea, yes.

samuelhwilliams: Added a test in 35193a8f23f36f44d5a98f017b6efab3335c576b

test/pages.py
```diff
@@ -0,0 +1,49 @@
...
+from pyppeteer.browser import Browser
+
+
+async def get_browser() -> Browser:
+    browser = await launch(
+        headless=True,
+        executablePath='/usr/bin/chromium-browser',
+        args=['--no-sandbox', '--disable-gpu'],
+    )
+    return browser
```
kristof-bb, line 11: I like the browser based testing. Very readable.

test/test_application.py
```diff
@@ -2007,6 +2009,110 @@ async def test_sentry_dsn_does_not_stop_proxy_from_becoming_healthy(self):
...
+        assert "Analyse in tools (2)" in with_no_filters
+        assert "Use as reference data (1)" in with_no_filters
+        assert "View data visualisation (1)" in with_no_filters
+
+        # Toggle the master dataset filter on (filters: master)
+        await home_page.toggle_filter("Analyse in tools")
+
+        with_master_filter = await home_page.get_html()
+        assert "You have access (1)" in with_master_filter
+        assert "Download (3)" in with_master_filter
```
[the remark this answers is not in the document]
samuelhwilliams: Good call out. I just spoke with Bo and went over the design again and we caught some edge cases, so have changed the implementation slightly.

The filter numbers will remain more static now, so shouldn't increase like you've called out here. They'll now show, of the results available, how many match each filter.

test/pages.py
michalc: I now wonder if we should replace some of code in the integration tests that interacts with Django models via the Django shell, with the browser taking actions in Django admin...

michalc: (To be clear: not in this PR)

dataworkspace/dataworkspace/static/app-filter-show-more.js
```diff
@@ -27,9 +27,16 @@ FilterShowMore.prototype.init = function () {
           $(choices[i]).show();
       }
   });
+  window.addEventListener('popstate', function () {
+      that.init();
```
michalc, line 31: Why the call to init on popstate?

Also, it looks like that on popstate, init would be called, which adds a listener to popstate, so next popstate, it fires twice, so then two more listeners are added...

samuelhwilliams: Yeah, this isn't necessary, you're right. I think it's a fragment from a previous alternative implementation.

dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -76,25 +87,21 @@
...
+        2) matches the search terms
 
-    dataset_filter = Q(published=True)
+    Annotates the dataset queryset with:
+        1) `has_access`, if the user can use the dataset's data.
+    """
+    # Filter out datasets that the user is not allowed to even know about.
+    visibility_filter = Q(published=True)
 
     if user:
```
michalc, line 106: Although it looks like it was doing it before, so feel free to not change it now, but I suspect there isn't a need to check if there is a user. There really should be a user(?), to the point where I would be fairly happy exploding if somehow this function has been reached without a user.

dataworkspace/dataworkspace/static/search.js
```diff
@@ -151,6 +151,11 @@
     }
 
     this.$wrapper.css('opacity', '1')
+
+    if (window.installFilterShowMore !== undefined) {
+        // Hook into app-filter-show-more.js to un-hide the "show more" button on search filters.
+        window.installFilterShowMore();
```
michalc, line 157: Just thinking about this... I'm not 100% on how the JS works, but if you click "Show more", and then click the checkbox next to one of the revealed sources, will that source disappear from the list? If so, should instead the "Show more" state be preserved between searches?

michalc: Ah also: why the `window.installFilterShowMore !== undefined` check?

dataworkspace/dataworkspace/apps/datasets/views.py
samuelhwilliams: Yeah, I am a bit curious why this check exists ... We could rip it out separately but I think this PR is large enough as-is.

dataworkspace/dataworkspace/static/search.js
samuelhwilliams: It would be nice for the "Show more" state to be preserved between searches, yes. I thought about it but left it for now as MVP... But could revisit that if you think it's worth doing now.

michalc: I think prod already preserves the state of show more, so it feels like it would be a regression, so I think worth doing now.

samuelhwilliams: Mm, prod preserves show more because it doesn't update the search filters at all. So yes it's a regression, but it's technically still better than prod. Prod is actually broken right now because if you filter on some source tags, then refresh the page, a load of the source tags disappear. (This was semi intentional but have agreed with Sanna to partially undo that as part of this).

test/pages.py
samuelhwilliams: Could do that, although (our) django-admin is pretty slow so I think that would just make the tests take even longer, which is not ideal and perhaps an argument to not do that (though we could still have some tests that assert behaviour of our admin panel).

dataworkspace/dataworkspace/static/search.js
samuelhwilliams: Now that I've tested it myself and see better what you're talking about, I agree ... so I'll make a patch to preserve show more state.

samuelhwilliams: Ok, done.

michalc (approved):

PR #937, opened 2020-09-02
fix: add query param to support embedding in ie11

samuelhwilliams:
### Description of change
Tested via manual deploy to analysisworkspace using an IE11 VM.

Dashboard embedding is broken in IE11. The latest update from support
suggesed using their AWS QuickSight JS embedding SDK, which we tried,
but ironically that doesn't support IE11 right now (syntax error:
awslabs/amazon-quicksight-embedding-sdk#50).
The "non-recommended" alternative was to append this query parameter
manually, which we'll now do in order to bring IE11 support back
immediately.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/views.py
+++ b/dataworkspace/dataworkspace/apps/applications/views.py
@@ -4,7 +4,7 @@ import datetime
 import itertools
 import random
 import re
-from urllib.parse import urlsplit
+from urllib.parse import urlsplit, urlencode
 
 from csp.decorators import csp_exempt, csp_update
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


niross (approved):

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -228,8 +228,12 @@ def _get_embedded_quicksight_dashboard(request, dashboard_id):
         dashboard_id, request.user
     )
 
+    extra_params = urlencode(
+        [("punyCodeEmbedOrigin", f"{request.scheme}://{request.get_host()}/-")]
```
michalc, line 232: Out of curiosity, why the "-" at the end? (It's crossed my mind it's a typo, but if not, maybe worth a comment since, to me at least, it is quite unexpected?)

samuelhwilliams: Ah, not a typo, probably worth a comment as you say - though I don't know what it does. :) The AWS JS SDK formatted the query param this way so I just copied it. May not actually be necessary...

PR #938, opened 2020-09-02
feature: parse datacut sql queries and display queried tables on admin screen

abbas123456:
### Description of change

This PR uses the https://github.com/alculquicondor/psqlparse library to parse the sql query on the CustomDatasetQuery model. 

The parser uses the Postgres source code to parse the query so it doesn't need to connect to the database or run any additional queries. 

The parser generates a list of database tables derived from a query's FROM clause and these are exposed on the CustomDatasetQuery model as a property called `parsed_query_tables`. 

For now this is only displayed on the admin screen as a readonly field as shown below but will be shown on the front end eventually:

Screenshot 2020-09-02 at 16 50 40

It was decided to do this parsing dynamically rather than on save to avoid having to migrate all of the existing queries on production. We should probably re-evaluate this once we are happy with the quality and accuracy of the parsing and once we know exactly what we will be doing with the table names.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/tests/datasets/test_models.py
+++ b/dataworkspace/dataworkspace/tests/datasets/test_models.py
@@ -1,3 +1,4 @@
+import psqlparse
 import pytest
 
 from dataworkspace.apps.datasets.models import SourceLink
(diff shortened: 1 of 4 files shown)


dataworkspace/dataworkspace/tests/datasets/test_models.py
```diff
@@ -137,3 +138,29 @@ def test_source_link_filename(db):
...
+
+    join_query = factories.CustomDatasetQueryFactory(
+        dataset=ds, query='select * from foo join bar on foo.id = bar.id'
+    )
+    assert sorted(join_query.parsed_query_tables) == ['bar', 'foo']
+
+    with_query = factories.CustomDatasetQueryFactory(
+        dataset=ds, query='with test as (select * from foo) select * from test'
+    )
+    assert sorted(with_query.parsed_query_tables) == ['foo', 'test']
```
samuelhwilliams, line 162: Do we actually want it to report `test` here? I suspect it would be more useful to only report actual underlying tables... is that possible?

abbas123456: Ideally it should only return actual tables but this is out of our control. We may be able to filter them out by extending the parser but I haven't tried and I don't think its worth the effort. I think postgres treats them similar to temporary tables so I guess they end up in the query plan but I'm not on expert on with statements.

samuelhwilliams: I guess we could check them against the DB in the future and filter out any that don't have corresponding tables ... but fair enough to leave that for a future iteration.

samuelhwilliams (approved):

sekharpanja (approved):

PR #946, opened 2020-09-04
feature: use temporary views to extract tables from datacut queries

abbas123456:
### Description of change

Extracting table names from the data cut queries was previously done using the psqlparse library (see #938) but this had some limitations as it treats WITH clause statements as table names.

The change introduced in this PR involves subclassing the admin form's save method. It connects to the database and runs some raw SQL to extract the table names. ~~and doubles up as a way of validating the data cut queries at the same time.~~

As this method of parsing involves running database queries, the table names are now saved in a new model called `CustomDatasetQueryTable` rather than generated dynamically.

The PR also includes a data migration to migrate the existing queries.

### Checklist


dataworkspace/dataworkspace/apps/dw_admin/forms.py
```diff
@@ -429,6 +431,34 @@ def clean(self):
...
+            self.cleaned_data['database'].memorable_name
+        ].cursor() as cursor:
+            try:
+                cursor.execute(
+                    f"create temporary view get_tables as (select 1 from ({self.cleaned_data['query']}) sq limit 1)"
+                )
+            except DatabaseError as e:
+                raise forms.ValidationError(str(e))
+            cursor.execute(
+                "select table_name from information_schema.view_table_usage where view_name = 'get_tables'"
```
michalc, line 451: Should the schema name also be fetched?

(Also just to confirm, is the creation of the view going to behave with multiple concurrent requests creating views of the same name? I'm not familiar with exactly how temporary views work...)

kristof-bb: temporary view are only visible within the same session and get dropped afterwards. they use some unique generated schema like pg_temp_x something so should be ok concurrently i think.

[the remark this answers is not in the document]
abbas123456: See above response from @kristof-bb

abbas123456: Have prefixed the table names with their schema name

dataworkspace/dataworkspace/apps/dw_admin/forms.py
```diff
@@ -429,6 +431,34 @@ def clean(self):
...
+            cursor.execute(
+                "select table_schema, table_name from information_schema.view_table_usage where view_name = 'get_tables'"
+            )
+            tables = cursor.fetchall()
+            cursor.execute("drop view get_tables")
+
+        # Save the extracted tables against the model for later use. This has to be set
+        # against the instance rather than cleaned_data as 'tables' is a readonly admin
+        # field and therefore the admin save method will not persist any changes to it.
+        self.instance.tables = [f'{t[0]}.{t[1]}' for t in tables]
```
michalc, line 459: Can/should the table and schema names be kept separate? Both can have dots in, and I think would need to be escaped separately in queries. (And I think some of our table and/or schema names _do_ have dots in).

abbas123456: I think this field is mainly for display purposes rather than for copying / pasting into queries but I might be wrong. How would you suggest storing them seperately in the model? Using a dict and storing as a JSONField?

abbas123456: A simple fix would be to add quote marks around the schema and table name like this:

self.instance.tables = [f'"{t[0]}"."{t[1]}"' for t in tables]

michalc: Oh I actually assumed they would be used to query the metadata for any upstream master datasets... 

It's borderline overkill, but another model? It would make the link between master datasets and datacuts right there in the database, and be fairly future friendly to a reasonable range of possible futures, since a master dataset is specified by schema and table name (in separate fields).

michalc: > A simple fix would be to add quote marks around the schema and table name like this:

I would probably still lean to separate model, or at least somehow separate without the quotes. It's a bit of a crazy edge case, but technically I think table and schema names can have quotes in them, and then used as identifiers as long as they are all correctly wrapped in quotes and escaped.

But also, in some queries, you need the table and schema names not as identifiers, but as literals, and so have single quotes, and different escaping rules. In fact, I think when querying `information_schema` as here, they would be single-quoted.

abbas123456: Ok thanks for the suggestion, I will have a chat with Anthony and see if what you are proposing aligns with the original request and then I'll update accordingly.

abbas123456: We agreed to use a foreign key model to store the table and schema names for each CustomDatasetQuery object

abbas123456:
Please note that query validation has now been removed. Now that the logic invoves creating a foreign key object it made sense to do this once the admin inline form has been validated and saved to the database. Otherwise we would have to delete any foreign key objects created in the clean_query method if the object fails to save for whatever reason and this complexity isn't worth the effort.

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -615,12 +614,22 @@ def can_show_link_for_user(self, user):
...
-            return None
-        statements = psqlparse.parse(self.query)
-        return list(statements[0].tables())
+
+class CustomDatasetQueryTable(models.Model):
+    query = models.ForeignKey(
+        CustomDatasetQuery, on_delete=models.CASCADE, related_name='tables'
+    )
+    table = models.CharField(
+        max_length=1024,
```
samuelhwilliams, line 623: Very minor point, but I think the table/schema identifiers can only be 64 (63?) characters long right now, so maybe would be nice if these were the same. But doesn't make a big difference.

dataworkspace/dataworkspace/apps/datasets/migrations/0053_set_custom_dataset_query_tables.py
```diff
@@ -0,0 +1,39 @@
...
+
+    for query in CustomDatasetQuery.objects.all():
+        if query.tables.all():
+            continue
+
+        with connections[query.database.memorable_name].cursor() as cursor:
+            try:
+                cursor.execute(
+                    f"create temporary view get_tables as (select 1 from ({query.query}) sq limit 1)"
+                )
```
samuelhwilliams, line 18: I'm not sure how postgres might optimise this query, but will doing `select 1 from (query)` actually run the entire query? Would it be a problem if it's a fairly slow?

I know we have CSV pipelines in data-flow to handle the use-case where datacut queries become fairly expensive, but not sure exactly where that cut-off lies. If it does execute the query, should we put a timeout on this step?

dataworkspace/dataworkspace/apps/datasets/models.py
abbas123456: I based the max_length on the value used in the master dataset's SourceTable which should correspond to the tables being queried in the datacut:

https://github.com/uktrade/data-workspace/blob/master/dataworkspace/dataworkspace/apps/datasets/models.py#L446-L450

dataworkspace/dataworkspace/apps/datasets/migrations/0053_set_custom_dataset_query_tables.py
michalc: It would surprise me if creating a view would run the query... https://www.postgresql.org/docs/9.2/sql-createview.html#REFSECT1 I think strongly implies this by

> CREATE VIEW defines a view of a query. The view is not physically materialized. Instead, the query is run every time the view is referenced in a query.

dataworkspace/dataworkspace/apps/dw_admin/forms.py
```diff
@@ -429,6 +431,32 @@ def clean(self):
                 # be updated for users who have `reviewed` as a read-only field (i.e. "Subject Matter Experts").
                 self.instance.reviewed = False
 
+        return self.cleaned_data
+
+    def save(self, commit=True):
```
michalc, line 436: Should this function/code be in a transaction (unless it is already)? (Say, to avoid the case of two people changing the same query, with the query from one can be saved, but then the tables from the other?)

dataworkspace/dataworkspace/apps/datasets/models.py
samuelhwilliams: Ah fair enough :)

dataworkspace/dataworkspace/apps/datasets/migrations/0053_set_custom_dataset_query_tables.py
samuelhwilliams: Ah right - great. 👍

michalc (approved):

dataworkspace/dataworkspace/apps/datasets/migrations/0053_set_custom_dataset_query_tables.py
abbas123456: Yes that's correct Michal, what I will do is remove the `limit 1` as it doesn't actually make a difference here because the underlying query never gets executed.

kristof-bb: you can test with:
create temporary view get_tables as (
	select 1 from (select pg_sleep(10)) sq
);
-> executes immediately

select * from get_tables
-> takes 10 sec

dataworkspace/dataworkspace/apps/dw_admin/forms.py
abbas123456: Yeah that's a good point, but to guarantee that we would need to lock the row using select_for_update. Will update the PR.

abbas123456: I will also add a database lock on the CustomDatasetQuery row being updated to prevent other race conditions

PR #954, opened 2020-09-07
feat: add last updated time to data cut and master dataset pages

niross:
Closes: https://trello.com/c/U2f069TH/779-spike-show-last-updated-date-and-time-against-data-cuts-and-master-datasets

- Query the new data flow metadata table for master dataset and query last update date
- Query s3 for data cut source link last updated date
 
#### Data cut page before
before

#### Data cut page after
after

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/models.py
+++ b/dataworkspace/dataworkspace/apps/datasets/models.py
@@ -472,7 +472,7 @@ class SourceTable(BaseSource):
 
     def get_data_last_updated_date(self):
         return get_tables_last_updated_date(
-            self.database.memorable_name, ['.'.join([self.schema, self.table])]
+            self.database.memorable_name, ((self.schema, self.table),)
         )
 
 
@@ -631,6 +631,15 @@ class CustomDatasetQuery(ReferenceNumberedDatasetSource):
     def type(self):
         return DataLinkType.CUSTOM_QUERY.value
 
+    def get_data_last_updated_date(self):
+        tables = CustomDatasetQueryTable.objects.filter(query=self)
+        if tables:
+            return get_tables_last_updated_date(
+                self.database.memorable_name,
+                tuple((table.schema, table.table) for table in tables),
+            )
+        return None
+
 
 class CustomDatasetQueryTable(models.Model):
     query = models.ForeignKey(
@@ -647,14 +656,6 @@ class CustomDatasetQueryTable(models.Model):
         validators=[RegexValidator(regex=r'^[a-zA-Z][a-zA-Z0-9_\.]*$')],
         default='public',
     )
-
-    def get_data_last_updated_date(self):
-        # Ensure all tables have a schema as `public` is not always specified in queries
-        tables = [
-            f'{"public." if len(table_name.split(".")) == 1 else ""}{table_name}'
-            for table_name in self.parsed_query_tables
-        ]
-        return get_tables_last_updated_date(self.database.memorable_name, tables)
 
 
 class ReferenceDataset(DeletableTimestampedUserModel):
--- a/dataworkspace/dataworkspace/apps/datasets/models.py
+++ b/dataworkspace/dataworkspace/apps/datasets/models.py
@@ -43,6 +43,7 @@ from dataworkspace.apps.datasets.model_utils import (
     get_linked_field_display_name,
     get_linked_field_identifier_name,
 )
+from dataworkspace.datasets_db import get_tables_last_updated_date
 
 
 class DataGroupingManager(DeletableQuerySet):
@@ -470,10 +471,6 @@ class SourceTable(BaseSource):
         return DataLinkType.SOURCE_TABLE.value
 
     def get_data_last_updated_date(self):
-        from dataworkspace.datasets_db import (
-            get_tables_last_updated_date,
-        )  # pylint: disable=import-outside-toplevel
-
         return get_tables_last_updated_date(
             self.database.memorable_name, ['.'.join([self.schema, self.table])]
         )
@@ -652,10 +649,6 @@ class CustomDatasetQueryTable(models.Model):
     )
 
     def get_data_last_updated_date(self):
-        from dataworkspace.datasets_db import (
-            get_tables_last_updated_date,
-        )  # pylint: disable=import-outside-toplevel
-
         # Ensure all tables have a schema as `public` is not always specified in queries
         tables = [
             f'{"public." if len(table_name.split(".")) == 1 else ""}{table_name}'
--- a/dataworkspace/dataworkspace/apps/datasets/models.py
+++ b/dataworkspace/dataworkspace/apps/datasets/models.py
@@ -469,6 +469,15 @@ class SourceTable(BaseSource):
     def type(self):
         return DataLinkType.SOURCE_TABLE.value
 
+    def get_data_last_updated_date(self):
+        from dataworkspace.datasets_db import (
+            get_tables_last_updated_date,
+        )  # pylint: disable=import-outside-toplevel
+
+        return get_tables_last_updated_date(
+            self.database.memorable_name, ['.'.join([self.schema, self.table])]
+        )
+
 
 class SourceView(BaseSource):
     view = models.CharField(
@@ -572,6 +581,17 @@ class SourceLink(ReferenceNumberedDatasetSource):
     def type(self):
         return DataLinkType.SOURCE_LINK.value
 
+    def get_data_last_updated_date(self):
+        if self.link_type == self.TYPE_LOCAL:
+            try:
+                metadata = boto3.client('s3').head_object(
+                    Bucket=settings.AWS_UPLOADS_BUCKET, Key=self.url
+                )
+                return metadata.get('LastModified')
+            except ClientError:
+                pass
+        return None
+
 
 class CustomDatasetQuery(ReferenceNumberedDatasetSource):
     FREQ_DAILY = 1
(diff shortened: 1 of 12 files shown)


dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -470,6 +470,15 @@ def can_show_link_for_user(self, user):
     def type(self):
         return DataLinkType.SOURCE_TABLE.value
 
+    def get_data_last_updated_date(self):
+        from dataworkspace.datasets_db import (
+            get_tables_last_updated_date,
+        )  # pylint: disable=import-outside-toplevel
```
samuelhwilliams, line 476: Do we need to do this import here rather than just at the top level for some reason?

niross: I'm (still) having a bit of a nightmare with cyclic imports it turns out. I may have to just move this to the utils file with the rest of the code that accesses the db.

dataworkspace/dataworkspace/datasets_db.py
```diff
@@ -33,3 +34,32 @@ def query_columns(connection, source):
...
+            FROM (
+                SELECT
+                    CONCAT(metadata.table_schema, '.', metadata.table_name),
+                    MAX(dataflow_swapped_tables_utc) run_date
+                FROM dataflow.metadata
+                WHERE CONCAT(metadata.table_schema, '.', metadata.table_name) = ANY(%s)
+                GROUP BY 1
+            ) a
+            ''',
+            [tables],
```
samuelhwilliams, line 63: I don't think this query is quite right. I can guess why you're reading `dataflow_swapped_tables_utc`, because the `source_data_modified_utc` column is often NULL - but that's sort of intentional. `dataflow_swapped_tables_utc` doesn't describe when the source data is from - it just says when we made it available on Data Workspace, so it isn't semantically the correct thing to pull/display. We should definitely use `source_data_modified_utc` if it exists. This query as it stands will lie about the ONS datasets in particular - we might run the pipeline every day but the data is only updated once a month/quarter. This would report the ONS data as being updated daily.

We only have values for `source_data_modified_utc` in ONS data right now, I think, because they're the only datasets where we've defined what that value should be. For the remaining pipelines/master datasets we need to make decisions on what that should be and how we get it. For real-time APIs it's not going to be very wrong to use `dataflow_swapped_tables_utc`, but if we have any other pipelines that aren't real time then we aren't going to be showing the right thing - and it may vary on just how wrong it is.

dataworkspace/dataworkspace/apps/datasets/models.py
samuelhwilliams: Ahhh gotcha, ok. 👍

dataworkspace/dataworkspace/datasets_db.py
niross: Got it, i've swapped the table name out

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Ok sorted now. I've fixed my local pylint issue as well :+1:

dataworkspace/dataworkspace/datasets_db.py
```diff
@@ -18,18 +17,39 @@ def get_columns(database_name, schema=None, table=None, query=None):
...
+                    CONCAT(metadata.table_schema, '.', metadata.table_name),
+                    MAX(source_data_modified_utc) run_date
+                FROM dataflow.metadata
+                WHERE CONCAT(metadata.table_schema, '.', metadata.table_name) = ANY(%s)
+                GROUP BY 1
+            ) a
+            ''',
+            [tables],
+        )
+        return cursor.fetchone()[0]
```
samuelhwilliams, line 55: Would this error out if no tables match? I think we might need to support the case where there isn't a known last updated date for a set of tables.

niross: No, it just returns None if no tables match. I've added another test for null last updated date as well.

samuelhwilliams (approved):
Nice 👍 

I wonder if we might prefer to not have a column for "Data last updated" on master datasets when it's not available, but I'm fairly open to being transparent about not having that info yet either.

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -622,6 +639,14 @@ def parsed_query_tables(self):
         statements = psqlparse.parse(self.query)
         return list(statements[0].tables())
 
+    def get_data_last_updated_date(self):
+        # Ensure all tables have a schema as `public` is not always specified in queries
+        tables = [
+            f'{"public." if len(table_name.split(".")) == 1 else ""}{table_name}'
+            for table_name in self.parsed_query_tables
```
samuelhwilliams, line 646: I wonder a bit if this might break if the schema/table names have dots or quotes in them (some schemas are definitely dotted). Is this definitely safe? Should we quote/escape them to be extra sure?

niross:
> Nice 
> 
> I wonder if we might prefer to not have a column for "Data last updated" on master datasets when it's not available, but I'm fairly open to being transparent about not having that info yet either.

Hmm yeah, there is a couple of master datasets out there with more than one table though (export wins is the one that comes to mind), so there is a very small chance that one table on the page could have a last updated date and another one doesn't.

niross:
Also... I am going to hold off on merging this temporarily until @abbas123456's query table changes are in.

dataworkspace/dataworkspace/apps/datasets/models.py
niross: Aha yeah just added another comment. I think this will be unnecessary once https://github.com/uktrade/data-workspace/pull/946 is merged - so I'll hold off until then.

dataworkspace/dataworkspace/datasets_db.py
```diff
@@ -18,18 +17,39 @@ def get_columns(database_name, schema=None, table=None, query=None):
...
+
+        cursor.execute(
+            '''
+            SELECT MIN(run_date)
+            FROM (
+                SELECT
+                    CONCAT(metadata.table_schema, '.', metadata.table_name),
+                    MAX(source_data_modified_utc) run_date
+                FROM dataflow.metadata
+                WHERE CONCAT(metadata.table_schema, '.', metadata.table_name) = ANY(%s)
```
michalc, line 49: Would it be better/clearer to keep the table and schema names separate in the query, not concatenating them?

- Any indexes on metadata won't be used: this is always a full table scan?
- There are some edge cases when this would return the wrong thing? table,schema of ("one.two","three") would be the same as ("one","two.three")

niross: Sure, yes, this would require some work on the data flow side so happy to create a ticket to cover it (and include updating this in it). Currently we don't have any schema names with dots configured in data flow.

niross: Actually scratch that I got confused sorry. We can split the schema and table out but I it would mean running the query for every table separately. I went this route because we'd need to run this query for each table, so if there are queries with lots of tables and data cuts with lots of queries we would be running this quite a lot per page view.

niross: Unless... I guess I could build an or query. Let me have a think

michalc: How about something like...
```sql
SELECT
  DISTINCT ON (table_schema, table_name)
  table_schema,
  table_name,
  MAX(dataflow_swapped_tables_utc)
OVER
  (PARTITION BY table_schema, table_name)
FROM
  dataflow.metadata
WHERE
  (table_schema, table_name) IN (
    ('companieshouse', 'companies_match_ids'),
    ('dit', 'consent_service__current_consents')
  );
```

niross: Cheers, I am not sure if the partition is necessary here is it? Could I not achieve the same result with a group by? I've not used over/partition before so not sure what the benefit is.

```
SELECT
    table_schema,
    table_name,
    MAX(dataflow_swapped_tables_utc) AS run_date
FROM dataflow.metadata
WHERE (table_schema, table_name) IN (
    ('companieshouse', 'companies_match_ids'),
    ('dit', 'consent_service__current_consents')
)
GROUP BY (1, 2)
```

michalc: You're absolutely right: I was needlessly overcomplicating it

(Also: I didn't now you could group by by position in the select list...)

dataworkspace/dataworkspace/datasets_db.py
```diff
@@ -18,18 +17,40 @@ def get_columns(database_name, schema=None, table=None, query=None):
...
-    sql = psycopg2.sql.SQL('SELECT * from {} WHERE false').format(source)
-    with connection.cursor() as cursor:
-        cursor.execute(sql)
-        return [c[0] for c in cursor.description]
+def get_tables_last_updated_date(database_name: str, tables: Tuple[Tuple[str, str]]):
+    """
+    Return the earliest of the last updated dates for a list of tables.
+    """
+    with connections[database_name].cursor() as cursor:
+        # Check the metadata table exists before we query it
```
michalc, line 36: Just realised... this does exist in production. Can we assume it exists? (together with creating it in any other environments?)

niross: Yeah was wavering on that myself. Happy to remove it

PR #974, opened 2020-09-24
feat: hide other users queries

samuelhwilliams:
### Description of change
The tool-based Data Explorer is a single-tenant solution, and we want to
maintain the existing design/implementation as we move across to the
integrated version.

This patch accordingly hides other users queries and makes them
inaccessible.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/explorer/views.py
+++ b/dataworkspace/dataworkspace/apps/explorer/views.py
@@ -9,7 +9,12 @@ from django.contrib.auth.views import LoginView
 from django.db import DatabaseError
 from django.db.models import Count
 from django.forms.models import model_to_dict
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/apps/explorer/views.py
```diff
@@ -419,6 +432,7 @@ def get(self, request, query_id):
 
     def post(self, request, query_id):
         action = request.POST.get("action", "")
+        print(action)
```
michalc, line 435: Left in from debugging?

michalc (approved):

dataworkspace/dataworkspace/apps/explorer/views.py
samuelhwilliams: Yes :)

PR #994, opened 2020-10-01
feat: launch dataset in data explorer

samuelhwilliams:
### Description of change
Add a link on master dataset pages to launch a basic query of that one
of that dataset's tables in Data Explorer. For now, this is behind the
same flag that controls whether or not a user can use the integrated
Data Explorer, so that people with the tools-based version don't see
this and think it doesn't work.

### Checklist


niross (approved):
Nice! We've been wanting something like this for a year :-)

PR #1008, opened 2020-10-06
feature: add R script to build cran mirror with binary packages

abbas123456:
### Description of change

The existing CRAN mirror serves source packages and these can be very slow to compile. This PR introduces a new mirroring script written in R that installs a small subset of packages and uploads the built artifacts to a seperate folder in S3 containing only binary packages.

The rstudio and visualisation-base-r Docker images have also been changed to have two CRAN mirrors defined. This means that binary packages are installed if they exist in the new mirror and if not, it will fall back to installing source packages from the second mirror.

### Checklist


michalc (approved):

michalc:
Looks good!

PR #1015, opened 2020-10-09
feat: don't cache pip packages

michalc:
### Description of change

This would result in both smaller Docker images, and so a faster start, and use less disk space. And, I suspect less likely things would go wrong down the line if some of the packages do something environment-specific like compiling.

From https://pip.pypa.io/en/stable/user_guide/, to _enable_ the boolean option `--no-cache-dir` you have to specify a falsey value.

### Checklist


samuelhwilliams:
>From https://pip.pypa.io/en/stable/user_guide/, to enable the boolean option --no-cache-dir you have to specify a falsey value.

Well that's super intuitive ...

samuelhwilliams (approved):

PR #1017, opened 2020-10-09
Data explorer cleanup

samuelhwilliams:
### Description of change
* Remove unused parts of Data Explorer
* Remove the webpack/sass preprocessors and the npm/node supporting stack - replaced with plain JS served directly. This does mean we're serving more, smaller files, but at this time it does not feel justified to have the tech burden just to handle a couple of pages of the codebase.

### Checklist


samuelhwilliams:
Question to reviewer: we could merge all of the ACE JS fragments into a single file to minimally-optimise clients downloading 1 file instead fo 4. Do we think that's worth doing right now? It probably is.

kristof-bb (approved):
looks good!

kristof-bb:
> Question to reviewer: we could merge all of the ACE JS fragments into a single file to minimally-optimise clients downloading 1 file instead fo 4. Do we think that's worth doing right now? It probably is.

The less http requests to the server the better. since it's any easy change worth a shot i guess to see if it improves performance a bit.

PR #1023, opened 2020-10-13
fix: use POST not GET for dialog/saving queries

samuelhwilliams:
### Description of change
At the moment, in Data Explorer, we pass SQL queries and related data
through query params until the user does a final POST to create a query
object. This presents a couple of problems:

1) Long queries (e.g. over 2000 chars) may be truncated or corrupted by
   some browsers which have low limits on the length of URLs/query
   params.
2) There could potentially be data protection issues with sending
   queries in the URL, as it may contain sensitive information that
   would end up being logged/retained for longer than necessary.

Let's use POST to move SQL query data between pages. This addresses the
above concerns but does introduce a few others:

* Back/forward browser navigation is not as clean - users are likely to
see 'resubmit form'-style warnings if they navigate back, because we're
showing a page directly from a POST rather than redirecting them to a
GET page (because we don't create a query object until they finally
save, so there's nothing for users to GET).
* Slightly more complex form actions, as certain buttons will override
  the main form action and be handled by a different view. This is
  obviously a minor thing.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/explorer/views.py
+++ b/dataworkspace/dataworkspace/apps/explorer/views.py
@@ -170,6 +170,23 @@ class CreateQueryView(CreateView):
     def post(self, request, *args, **kwargs):
         action = request.POST.get('action', '')
 
+        if action == 'dialog':
+            # This represents the 'dialog' box for creating a query when clicking 'save' from the playground.
+            context_data = {}
+
+            sql = self.request.POST.get("sql")
+            if sql:
+                context_data['disable_sql'] = True
+
+            query_params = (('sql', request.POST.get('sql')),)
+            context_data['backlink'] = (
+                reverse("explorer:index") + f"?{urlencode(query_params)}"
+            )
+            context_data['form'] = self.get_form()
+            context_data['form'].errors.clear()
+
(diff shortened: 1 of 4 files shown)


dataworkspace/dataworkspace/apps/explorer/views.py
```diff
@@ -344,32 +385,43 @@ def render_with_sql(self, request, query, run_query=True, log=False):
...
+            form.errors.clear()
+            vm = query_viewmodel(
+                request.user,
+                query,
+                form=form,
+                run_query=show,
+                rows=url_get_rows(request),
+                page=url_get_page(request),
+            )
+            vm['backlink'] = reverse("explorer:index") + f"?query_id={query.id}"
```
samuelhwilliams, line 420: This backlink will potentially lose data for users where:
* they open a saved query to edit the SQL
* they click save and come to this dialog page, then click the back link
* changes to the saved query will be lost, because the SQL query box will be re-populated with the current SQL from the saved query.

It's not clear how to fix this with the current design.

niross (approved):
As discussed it's not ideal from a user perspective if you get stuck in a "resubmit confirmation loop" but the benefits of going down this route far outweigh that.

PR #1027, opened 2020-10-19
feature: update gitlab and quicksight visualisation last updated dates

abbas123456:
https://trello.com/c/NgNgdFFl/945-spike-data-visualisation-last-updated-information

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -34,7 +34,10 @@ from dataworkspace.apps.core.utils import (
     new_private_database_credentials,
     postgres_user,
 )
-from dataworkspace.apps.applications.gitlab import gitlab_has_developer_access
+from dataworkspace.apps.applications.gitlab import (
+    gitlab_api_v4,
+    gitlab_has_developer_access
+)
 from dataworkspace.apps.datasets.models import VisualisationCatalogueItem
 from dataworkspace.cel import celery_app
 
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -338,7 +340,7 @@ def sorted_datasets_and_visualisations_matching_query_for_user(query, use, user)
     all_datasets = (
         master_and_datacut_datasets.union(reference_datasets)
         .union(visualisations)
-        .order_by('-search_rank', 'name')
```
samuelhwilliams, line 341: Has this just slipped in? Doesn't feel like part of this change.

abbas123456: I have changed the order_by clause to be '-modified_date' first so I can confirm that the catalogue items are getting updated in the celery task without having to check the sql database

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -907,3 +910,47 @@ def sync_quicksight_permissions(
...
+        latest_commit = master_branch['commit']
+        latest_commit_date = datetime.datetime.strptime(
+            latest_commit['committed_date'], '%Y-%m-%dT%H:%M:%S.%f%z'
+        )
+
+        visualisation.modified_date = latest_commit_date
+        visualisation.save()
+
+    # AWS QuickSight visualisations
+    for visualisation in VisualisationCatalogueItem.objects.filter(visualisation_template__isnull=True):
```
samuelhwilliams, line 935: I think it's technically possible for a catalogue item to have a gitlab visualisation and a quicksight visualisation, so this is probably not quite the right query.

abbas123456: From Alex:

The data visualisation add/edit page on Django admin supports both types but I'm not aware of a use case for including both and don't think there are any instances of this.

samuelhwilliams: Yeah I agree, there's not much of a use case ... but it's technically possible, and presumably also fairly trivial to update the query to support it, so I guess my question would be "why not?"

dataworkspace/dataworkspace/apps/datasets/views.py
samuelhwilliams: Yep, I understand why you've done it of course - just saying that it's probably not a change we'd want to merge as part of this?

abbas123456: Oh ok sorry, yeah the plan was to remove it after I've finished testing

dataworkspace/dataworkspace/apps/applications/utils.py
abbas123456: Yeah fair enough, will update it

abbas123456:
Will close the PR for now as have decided to focuson making a catalogue item's "Published date" available as an option for sorting first.

PR #1028, opened 2020-10-19
Move signature check to config

abbas123456:
### Description of change

As we are using a custom debian mirrror we currently disable the check which verifies expired digital signatures.

This PR moves this option to the apt config rather than setting it every time we call apt-get commands.

The docker images affected were all built successfully on my local machine after making the change.

### Checklist


michalc (approved):
Looks good. Can the Dockerfile in visualisation-base also be changed? (I realise it doesn't even have use current pattern, but I think it should)

PR #1031, opened 2020-10-20
feat: updated development environment

niross:
### Description of change

- Update dev docker config to run nginx -> proxy -> django app to match production
- Add hot loading to the proxy and django apps when running locally
- Add libraries and docs for debugging remotely with docker


abbas123456 (approved):

dataworkspace/start-dev.sh
```diff
@@ -0,0 +1,28 @@
...
+    # it manages to read its config file. To avoid errors on startup,
+    # we configure its prefix to be a writable location
+    mkdir -p /home/django/logs
+
+    # Start nginx, proxy and application
+    echo "Starting nginx, proxy and django application..."
+    parallel --will-cite --line-buffer --jobs 5 --halt now,done=1 ::: \
+        "python3 -m start" \
+        "PROXY_PORT='8001' UPSTREAM_ROOT='http://localhost:8002' nodemon --watch proxy.py --exec 'python3 -m proxy'" \
+        "nginx -p /home/django"
```
samuelhwilliams, line 27: Would it make sense to also start the celery beat/worker here?

abbas123456:
I think some changes need to be made to the docker-compose files after https://github.com/uktrade/data-workspace/pull/1029 was merged:

I'm getting the following error when running locally:

```
$ docker-compose -f docker-compose-dev.yml up --build
ERROR: The Compose file './docker-compose-dev.yml' is invalid because:
networks.default value Additional properties are not allowed ('name' was unexpected)
```

dataworkspace/start-dev.sh
niross: Yeah can do, i don't think we've ever run celery locally (by default anyway) so didn't occur to me.

Dockerfile
```diff
@@ -55,10 +55,25 @@ USER django
...
+RUN \
+    apk add --no-cache \
+    nodejs=10.19.0-r0 \
+    npm=10.19.0-r0
+
+RUN npm install --global --unsafe-perm nodemon
+
+USER django
+
+CMD ["/dataworkspace/start-dev.sh"]
```
samuelhwilliams, line 71: Wonder how much worse this will make our problem with docker using up loads of disk space... 🤔

postgres/init-databases.sql
```diff
@@ -9,6 +9,15 @@ CREATE TABLE test_dataset (id int primary key, data text);
...
 FROM generate_series(1, 20000) AS t(i);
+CREATE SCHEMA dataflow;
+CREATE TABLE dataflow.metadata
+(
+    id SERIAL NOT NULL CONSTRAINT metadata_pkey PRIMARY KEY,
+    table_schema TEXT NOT NULL,
+    table_name TEXT NOT NULL,
+    source_data_modified_utc TIMESTAMP,
+    dataflow_swapped_tables_utc TIMESTAMP NOT NULL
+);
```
samuelhwilliams, line 20: Oof, this is tricky, yeah ... Can't think of anything better to do here immediately.

dataworkspace/start.py
```diff
@@ -10,16 +13,26 @@
...
 
-print('Stopping WSGIServer. Waiting for all requests to complete...', flush=True)
-gevent.get_hub().join()
 
-print('Requests completed. Exiting gracefully.', flush=True)
+if os.environ.get('DEBUG') == 'True':
+    print('Running in DEBUG mode with hot reloader')
+    run_with_reloader(run_server)
+else:
+    run_server()
```
samuelhwilliams, line 38: For some reason this makes me feel a little iffy. I can't imagine any of us setting DEBUG to true in prod, but it's only one condition away...

niross: I know what you're saying, setting debug to true on prod would also enable debug toolbar and a few other bits as well I think.

postgres/init-databases.sql
niross: yep, only added as a new dev on firebreak was unable to spin the project up due to this table being missing.

samuelhwilliams (approved):
Commits probably need squashing but this looks good 🎉

PR #1032, opened 2020-10-20
feat: use POST rather than GET for Data Explorer

samuelhwilliams:
### Description of change
Use POST to send SQL queries from the playground to the "save a query"
form. Our proxy, and some older browser (e.g. IE11) can't handle URLs
that are more than ~2000 chars, which some of our users' queries are.

This patch adds an intermediate query model used to save the SQL
temporarily so that we can use clean post-redirect-get processes to
prevent users constantly hitting "resubmit forms" pages if they use
browser navigation back/forth. This temporary model is only used to
populate the SQL fields in the playground and the query saving page, and
a clean-up job will semi-regularly delete old rows from this model as
they should be considered transient.

### Checklist


michalc (approved):

abbas123456 (approved):

abbas123456:
Good work

PR #1034, opened 2020-10-21
Add published date for master and datacut sets

abbas123456:
### Description of change

This PR adds `published_at` to Master and Datacut datasets in the same way its currently used for Reference datasets and Visualisation Catalogue Items. This will enable us to sort by published date on the catalogue page.

The PR also adds the "published_at" date to all catalogue item templates.

A data migration has also been added to set the published_at dates for all existing Master and Datacut datasets.

### Checklist


samuelhwilliams (dismissed):

dataworkspace/dataworkspace/templates/datasets/referencedataset_detail.html
```diff
@@ -167,8 +167,12 @@ <h2 class="govuk-heading-l govuk-!-margin-top-8">
 
     <div class="govuk-grid-row">
       <div class="govuk-grid-column-two-thirds">
-        <h2 class="govuk-heading-l govuk-!-margin-top-8">Additional information</h2>
+        <h2 class="govuk-heading-l govuk-!-margin-top-8">Additional Information</h2>
```
samuelhwilliams, line 170: Could be overruled but generally sentence case is preferred in headings to title case for GDS styles, so I'd say the other headings should be updated instead probably.

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -200,6 +201,21 @@ def __init__(self, *args, **kwargs):
     def __str__(self):
         return self.name
 
+    def save(
+        self, force_insert=False, force_update=False, using=None, update_fields=None
+    ):
+        self.update_published_timestamp()
```
samuelhwilliams, line 207: This feels like it might end up inserting strange published timestamps for all our existing datasets. Things that have been published for ages will start saying they have only just been published.

I wonder if we could (and if it's worth) scraping the event log for the first event on a given dataset and using the date from that as the published date? Or there might even be a specific event for publishing datasets that we could use. Would that be better?

abbas123456: I've just added a data migration that sets all existing Datasets to have published_at = created_date. Obviously this isn't necessarily true as some are created and then published at a later stage but it does avoid the scenario of giving old datasets a recent published date. I haven't looked into event log model but not sure if its worth the effort.

samuelhwilliams: oh yeah that's way easier 👍

dataworkspace/dataworkspace/apps/datasets/migrations/0057_auto_20201021_1454.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 3.0.8 on 2020-10-21 14:54
+
+from django.db import migrations
+
+
+def set_published_at_dates(apps, schema_editor):
+    Dataset = apps.get_model('datasets', 'dataset')
+    for dataset in Dataset.objects.filter(published_at__isnull=True):
```
samuelhwilliams, line 8: I think this filter should also check that the dataset is published (i.e. so we don't set published_at on unpublished datasets), right?

abbas123456: oh yeah, thanks for spotting that

samuelhwilliams (approved):
Looks good; maybe squash the commits down a bit?

PR #1039, opened 2020-10-26
fix: pgAdmin not properly starting when in a new tab

michalc:
### Description of change

Suspect it's due to https://github.com/postgres/pgadmin4/blob/REL-4_8/web/pgadmin/browser/static/js/layout.js#L77 which raises an exception when window.opener is present, but from a different domain.

> DOMException: Blocked a frame with origin [...] from accessing a cross-origin frame

It looks like it's fixed in more recent versions of pgAdmin, https://github.com/postgres/pgadmin4/blob/80ab5969920b1b258c4bb52020badbc67cf805ab/web/pgadmin/static/js/window.js#L14, but I think better for that to be a last resort since there could be UI changes.

It looks like IE11 doesn't support `rel="noopener"` https://caniuse.com/?search=noopener . However, it looks like window.opener is not set in IE for cross domains, so it doesn't need this fix https://stackoverflow.com/q/18625733/1319998

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/templates/partials/tool_section.html
+++ b/dataworkspace/dataworkspace/templates/partials/tool_section.html
@@ -36,7 +36,7 @@
           </form>
         </div>
         <div class="govuk-grid-column-one-half">
-          <form action="{{ application_url }}" method="GET" style="display: inline-block" class="govuk-!-width-full" target="_blank">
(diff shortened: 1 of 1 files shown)


michalc:
This would have been good... if it worked. It doesn't look like `rel="noopener"` is respected on `form` elements...

PR #1047, opened 2020-10-26
fix: semantic separation between debug and local

samuelhwilliams:
### Description of change
We use the `debug` Django setting to configure Django for running
locally, but it should be possible to run the app not in debug mode when
running it locally. This patch separates out the setting for running in
debug mode from a setting that confirms we're running locally, so that
various settings can be tied to the correct semantic check.

### DIFF
--- a/.envs/test.env
+++ b/.envs/test.env
@@ -1,4 +1,5 @@
 # Static: as in Dockerfile
+DEBUG=False
 PYTHONPATH=/dataworkspace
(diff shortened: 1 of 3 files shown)


niross (approved):
Do we want to use `LOCAL` in `start.py` as well?

samuelhwilliams:
Good point - happy with this change? Makes me feel a little bit safer about that as well now that there's two conditional checks ...

PR #1051, opened 2020-10-28
feat: expand sql query box in data explorer

samuelhwilliams:
### Description of change
Allow the SQL query box in Data Explorer to be expanded to (near) full
page width, as some users have reported that the box is too narrow to
use effectively.

### Checklist


dataworkspace/dataworkspace/static/explorer_query.js
```diff
@@ -95,3 +95,31 @@ if (sqlTextarea !== null) {
       );
   }
 }
+
+
+
+let tabsContainer = document.getElementById('playground-tabs');
+let fullWidthToggle = document.getElementById('full-width');
+let normalWidthToggle = document.getElementById('normal-width');
+console.log(tabsContainer, fullWidthToggle, normalWidthToggle);
```
michalc, line 104: Debug left in

dataworkspace/dataworkspace/static/explorer_custom.css
```diff
@@ -100,3 +100,9 @@
 .app-readonly-sql {
     background-color: #F0F0F0
 }
+
+@media only screen and (max-width: 1020px) {
```
michalc, line 104: For at least some of the other media queries:
- We use min-width rather than max-width
- The unit is em
Should it be consistent?

test/selenium/test_selenium.py
```diff
@@ -271,3 +272,33 @@ def test_data_explorer_cached_credentials_can_be_reset_using_admin_action(
...
+        self.driver.implicitly_wait(1)
+        try:
+            home_page.click_full_width()
+        except ElementNotInteractableException:
+            pass
+        finally:
+            self.driver.implicitly_wait(5)
+
+        # And that the "normal width" toggle is now click-able.
+        home_page.click_normal_width()
```
michalc, line 304: Should it also test that the button makes something wider?

test/selenium/test_selenium.py
```diff
@@ -271,3 +272,33 @@ def test_data_explorer_cached_credentials_can_be_reset_using_admin_action(
...
+        home_page.click_full_width()
+
+        # Check the "full width" toggle has been hidden
+        self.driver.implicitly_wait(1)
+        try:
+            home_page.click_full_width()
+        except ElementNotInteractableException:
+            pass
+        finally:
+            self.driver.implicitly_wait(5)
```
michalc, line 301: Would the above all pass even if the ElementNotInteractableException exceptions are _not_ thrown?

dataworkspace/dataworkspace/static/explorer_custom.css
samuelhwilliams: We're looking to hide a link when the screen is small - it feels right to use `max-width` to apply a style in that case? I suppose I could change it to min-width to apply an "unhide" style to desktop, but that feels contrived.

I'll switch it to `em`, that feels probably right.

samuelhwilliams: GOV.UK Design System uses 960px as their page width (the width of the main container), plus 30px margins either side, which is where the 1020px comes from. It feels right to use this rather than try to convert it to `em` - which I suspect in certain cases probably isn't a reliable conversion.

michalc: > I'll look at em ... GOV.UK Design System uses px for their screen sizes, which is what I cribbed from.

Actually, looking now, I see both `em` and `px` in media queries... and specifically, I see `@media (min-width: 1020px)`, so all good with `px`

However...

> I suppose I could change it to min-width to apply an "unhide" style to desktop, but that feels contrived.

Maybe a bit... But I guess I've got into too-complex layouts before with both mins and maxes floating about, and it tends to be easier to just pick one. And...

... I think "desktop" in GDS is `@media (min-width: 1020px)`, which I think is "1020px or above"(?) WIth the `max-width`, at exactly 1020px, the link won't be displayed, even though it's at "desktop" size. It is perhaps an almost trivial example, and doesn't matter too much right now, but it sort of introduces a unique layout at exactly 1020px, which is easy to miss when testing(?)

michalc (approved):

dataworkspace/dataworkspace/static/explorer_custom.css
samuelhwilliams: > Maybe a bit... But I guess I've got into too-complex layouts before with both mins and maxes floating about, and it tends to be easier to just pick one. And...

The 'best' thing would be if we used sass and then used the Design System variables directly, I suppose. I'll switch the media query up to min-width 1020px because, while I agree that the example is trivial, it still feels like a valid point re: unique layout at exactly 1020px.

PR #1052, opened 2020-10-28
Feature/sync activity stream sso users

abbas123456:
### Description of change

Tools access for new users can fail sometimes due to the IAM role creation that happens in the request/response cycle. The failure can either be due to rate limiting or the fact that IAM operations are eventually consistent.

This PR adds a celery task that runs every minute and polls the activity stream for SSO users and creates a django user for them if necessary.

If the user already exists it checks if they have permissions to run tools and if so, it checks whether they need to have an IAM role created. This check is determined by a new field called `tools_access_role_arn` that has been added to the profile model.

Another celery task has been added and this gets called in the following circumstances:

1) In the `sync_activity_stream_sso_users` celery task if the user has permissions to runs tools but has `tools_access_role_arn` set to None
2) During the login flow if the user has permissions to runs tools but has `tools_access_role_arn` set to None
3) When a user has the "Can start local tools" permission added in the admin screen and has `tools_access_role_arn` set to None

The PR also changes the current behaviour when a tool is spawned: Instead of calling AWS to check if the user has an IAM role, it now checks the new `tools_access_role_arn` field on the profile model and returns this if its set. If its not set, the function calls AWS to create the role and associated policy and then saves the arn against the profile model.

The functionality was tested on the dev environment using the following test script:
https://docs.google.com/document/d/1eLn0OPO0SWg0q1gcJtNxMwH4JoEbkrsGOZlSxThNTVM/edit

### Checklist


dataworkspace/dataworkspace/apps/accounts/models.py
```diff
@@ -9,6 +9,7 @@
 class Profile(models.Model):
     user = models.OneToOneField(get_user_model(), on_delete=models.CASCADE)
     sso_id = models.UUIDField(unique=True, default=uuid.uuid4)
+    tools_access_role_arn = models.TextField(null=True, blank=True)
```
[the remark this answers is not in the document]
samuelhwilliams: This feels like a slightly contentious general suggestion (albeit perhaps not in this case) ...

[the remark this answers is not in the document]
abbas123456: Yep

abbas123456:
The celery task will be changed to cache the most recent activity's "published" date and pass this to the query date range filter. This is to ensure activities are not missed during downtime.

dataworkspace/dataworkspace/apps/accounts/models.py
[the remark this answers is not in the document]
richtier: agreed, the suggestions are food for though rather than dictatorial "musts" :)

sometimes they hit the mark, other times not. In this case it seems to have the mark :fireworks:

.envs/test.env
```diff
@@ -92,3 +92,5 @@ EFS_ID=some-id
 
 EXPLORER_CONNECTIONS={"Postgres": "my_database"}
 EXPLORER_DEFAULT_CONNECTION=my_database
+
+DISABLE_CELERY=True
```
samuelhwilliams, line 96: I think this should probably be `DISABLE_CELERY_BEAT`, as we're still keeping celery+workers itself active?

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
...
+## 2020-10-28
+
+### Added
+- Field to cache the IAM role arn for tools access.
+- Celery task to poll activity stream and create django users.
+
+### Changed
+- Renamed core.utils.create_file_access_role and changed to check the user profile for tools_access_role_arn, returning this if its set instead of calling AWS.
+- Show tools_access_role_arn field in user admin screen.
+
```
samuelhwilliams, line 16: No need for this now (changelog is about to disappear)

dataworkspace/dataworkspace/tests/applications/test_utils.py
```diff
@@ -407,3 +418,388 @@ def test_poll_until_user_created(self, mock_cache, mock_boto3_client, mock_creds
...
         assert len(mock_data_client.create_data_source.call_args_list) == 1
         assert len(mock_data_client.update_data_source.call_args_list) == 0
+
+
+class TestSyncActivityStreamSSOUsers:
+    @pytest.mark.django_db
+    @mock.patch('dataworkspace.apps.applications.utils.hawk_request')
+    @override_settings(ACTIVITY_STREAM_BASE_URL='http://activity.stream')
+    def test_sync_calls_activity_stream(self, mock_hawk_request):
+        cache.delete('activity_stream_sync_last_published')
```
samuelhwilliams, line 428: This feels like test state is leaking between tests. Would it be better to disable the caching backend for these tests (e.g. using override_settings and the DummyCache)?

dataworkspace/dataworkspace/settings/base.py
```diff
@@ -253,33 +253,40 @@ def aws_fargate_private_ip():
...
-        'args': (),
-    },
-    'clean-up-old-data-explorer-playground-sql-queries': {
-        'task': 'dataworkspace.apps.explorer.tasks.cleanup_playground_sql_table',
-        'schedule': 60 * 60 * 6,
-        'args': (),
-    },
-}
+
+if not env.get('DISABLE_CELERY_BEAT_SCHEDULE', False):
```
samuelhwilliams, line 257: This is a slightly weird one, because `DISABLE_CELERY_BEAT_SCHEDULE=False` or `DISABLE_CELERY_BEAT_SCHEDULE=0` would both end up still disabling the schedule ... feels like one we'd fall for later.

Can we change this to do some type conversion, e.g. to a bool (strtobool is a fairly reasonable candidate, and I think used elsewhere here).

dataworkspace/dataworkspace/tests/applications/test_utils.py
abbas123456: Yep fair enough, the logic around this cached value is only required for the `test_sync_first_time` and `test_sync_recently_published_users` tests so I will override it for the rest

dataworkspace/dataworkspace/settings/base.py
abbas123456: Will do

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -907,3 +913,149 @@ def sync_quicksight_permissions(
...
+
+        logger.info('Fetched %d from activity stream', len(records))
+
+        for record in records:
+            obj = record['_source']['object']
+
+            user_id = obj['dit:StaffSSO:User:userId']
+            email = obj['dit:emailAddress'][0]
+
+            User = get_user_model()
```
samuelhwilliams, line 1016: Minor thing, but this only needs to be retrieved once, right? So we can do it all the way outside of the loops.

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -907,3 +913,149 @@ def sync_quicksight_permissions(
...
+            try:
+                user = User.objects.get(username=email)
+            except User.DoesNotExist:
+                logger.info('User with SSO id %s does not exist, creating...', user_id)
```
samuelhwilliams, line 1020: This isn't quite accurate ... 
```suggestion
                logger.info('User with email address %s does not exist, creating...', email)
```

dataworkspace/dataworkspace/tests/applications/test_utils.py
```diff
@@ -407,3 +418,388 @@ def test_poll_until_user_created(self, mock_cache, mock_boto3_client, mock_creds
...
+
+        _do_sync_activity_stream_sso_users()
+
+        User = get_user_model()
+        all_users = User.objects.all()
+
+        assert len(all_users) == 1
+        assert (
+            str(all_users[0].profile.sso_id) == '00000000-0000-0000-0000-000000000000'
+        )
```
samuelhwilliams, line 640: I think this test should/could also assert that `publish_to_iam_role_creation_channel` is called, maybe? If that's our interface into getting that piece of work done.

samuelhwilliams (commented):
Quite a chunky ticket - but looks good after you explained the context and scope to me. Ping me for a final review after addressing comments and then it should be good for 👍  from me.

dataworkspace/dataworkspace/tests/applications/test_utils.py
abbas123456: Because the celery task is doing two distinct things: 1) creating users and 2) calling `publish_to_iam_role_creation_channel`, I thought it was best to these seperately to account for the different edge cases. I prefer this pattern of testing when writing unit tests to make it clear how each code path should behave. The `test_sync_creates_role_if_user_can_access_tools`, `test_sync_doesnt_create_role_if_user_cant_access_tools`, and `test_sync_doesnt_create_role_if_user_already_has_role` make assertions related to the `publish_to_iam_role_creation_channel` method

abbas123456:
> Quite a chunky ticket - but looks good after you explained the context and scope to me. Ping me for a final review after addressing comments and then it should be good for 👍 from me.

Have updated the description and added some code initally planned for part 2 to give the PR a bit more context

dataworkspace/dataworkspace/apps/accounts/backends.py
```diff
@@ -46,6 +50,15 @@ def authenticate(self, request, username=None, password=None, **kwargs):
...
 
+            if (
+                user.user_permissions.filter(
+                    codename='start_all_applications',
+                    content_type=ContentType.objects.get_for_model(ApplicationInstance),
+                ).exists()
+                and not user.profile.tools_access_role_arn
+            ):
+                publish_to_iam_role_creation_channel(user)
+
```
samuelhwilliams, line 61: I think this will add a second DB request to every incoming HTTP request ... can we bundle it into line 34 somehow to prevent that?

samuelhwilliams (approved):
Good work 👍

dataworkspace/dataworkspace/apps/accounts/backends.py
samuelhwilliams: Ignore me :)

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -907,3 +914,172 @@ def sync_quicksight_permissions(
...
+
+
+@celery_app.task()
+def create_tools_access_iam_role_task(user_id):
+    with cache.lock(
+        f"create_tools_access_iam_role_task-{user_id}",
+        blocking_timeout=60,
+        timeout=360,
+    ):
+        _do_create_tools_access_iam_role(user_id)
```
samuelhwilliams, line 951: I'd be tempted to have zero blocking timeout and just set the task to retry after x seconds (i.e. re-queuing itself) if it can't immediately grab the lock ... should allow for a lot more load/waiting before anything breaks.

samuelhwilliams (approved):

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -907,3 +914,165 @@ def sync_quicksight_permissions(
...
+        data=body,
+        headers={'Authorization': header, 'Content-Type': content_type},
+    )
+    return response.status_code, response.content
+
+
+@celery_app.task(autoretry_for=(redis.exceptions.LockError,))
+def create_tools_access_iam_role_task(user_id):
+    with cache.lock(
+        f"create_tools_access_iam_role_task-{user_id}", blocking_timeout=0, timeout=360,
```
michalc, line 947: Should this be a the same lock for all users, to try to avoid the rate limiting issues? And then a small sleep after creating the role? (I don't know really, but it crossed my mind...)

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -907,3 +914,165 @@ def sync_quicksight_permissions(
...
+        logger.info('Fetched %d from activity stream', len(records))
+
+        for record in records:
+            obj = record['_source']['object']
+
+            user_id = obj['dit:StaffSSO:User:userId']
+            email = obj['dit:emailAddress'][0]
+
+            try:
+                user = User.objects.get(username=email)
```
michalc, line 1033: I think it should query for any of the email addresses in SSO, as it does currently https://github.com/uktrade/data-workspace/blob/master/dataworkspace/dataworkspace/apps/accounts/backends.py#L33

The first email in the list in SSO seems quite often to not be the one they use, and may well not be the one in in DW. For example, just after their contact email in SSO has been changed

michalc: Actually... I think should check if there is a user with a matching SSO ID (the UUID) first, and only then check if there is a user with a matching email address. (Essentially, the logic at https://github.com/uktrade/data-workspace/blob/master/dataworkspace/dataworkspace/apps/accounts/backends.py#L11 unless there is a reason for it to be different?)

abbas123456: As the logic in the authentication backend uses `HTTP_SSO_PROFILE_EMAIL` if `HTTP_SSO_PROFILE_CONTACT_EMAIL` is None, the same logic can't be replicated as Activity Stream returns a list of emails. The agreed logic is therefore:

1) It should search for a user with any of the emails in the list
2) It should create a user taking the first email in the list
3) And then update the user with the first email in the list

michalc (approved):

PR #1054, opened 2020-10-30
feat: allow the default memory and CPU to be set per tool and visualisation

michalc:
### Description of change

This is for one of the Innovation Lab's visualisations. It loads datasets into memory.

A nice bonus is that there is a bit less logic when firing up a Fargate task. CPU and memory is _always_ specified.

### Checklist


dataworkspace/dataworkspace/apps/applications/models.py
```diff
@@ -71,6 +71,16 @@ class ApplicationTemplate(TimeStampedModel):
...
+    default_memory = models.CharField(
+        default='8192',
+        max_length=16,
+        help_text='The default amount of memory allocated in MBs',
+    )
+    default_cpu = models.CharField(
+        default='1024',
+        max_length=16,
+        help_text='The default amount of CPU allocated, where 1024 is 1 CPU',
+    )
```
samuelhwilliams, line 83: Hmm, why CharField over IntegerField? Does it support short-hand like 4G?

michalc: As far I know, it doesn't support short-hand...

But ,there are some details as to why strings at https://github.com/uktrade/data-workspace/blob/master/dataworkspace/dataworkspace/apps/applications/models.py#L225 . When setting it up, it looked like boto3 expected strings for CPU and memory in the override, so we used strings everywhere

I'm not sure this is still the case to be honest. But... I would lean to consistency for now and keep "strings everywhere", and leave any migration until "later"

samuelhwilliams (approved):

PR #1055, opened 2020-10-30
fix: paginate through all users for quicksight sync

samuelhwilliams:
### Description of change

At the moment, the full overnight QuickSight sync is only running
against the first page of user results, and we now have well over 100
users (albeit mostly readers that this sync task will ignore). This
update adds pagination to the task so that all users are properly
synced. This is not overly urgent, because when users launch QuickSight
via Data Workspace we kick off an individual sync just for them, but
still feels like something we should fix.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -768,63 +768,7 @@ def create_update_delete_quicksight_user_data_sources(
                 raise e
 
 
-@celery_app.task()
-def sync_quicksight_permissions(
-    user_sso_ids_to_update=tuple(), poll_for_user_creation=False
-):
-    logger.info(
-        'sync_quicksight_user_datasources(%s, poll_for_user_creation=%s) started',
-        user_sso_ids_to_update,
-        poll_for_user_creation,
-    )
-
-    # QuickSight manages users in a single specific regions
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -902,6 +846,88 @@ def sync_quicksight_permissions(
...
+        'quicksight', region_name=settings.QUICKSIGHT_USER_REGION
+    )
+    # Data sources can be in other regions - so here we use the Data Workspace default from its env vars.
+    data_client = boto3.client('quicksight')
+
+    account_id = boto3.client('sts').get_caller_identity().get('Account')
+
+    quicksight_user_list: List[Dict[str, str]]
+    if len(user_sso_ids_to_update) > 0:
+        next_token = None
```
abbas123456, line 871: Does this need to be defined here? It's not being used in this condition.

abbas123456 (approved):
Looks good to me apart from one minor comment

dataworkspace/dataworkspace/apps/applications/utils.py
samuelhwilliams: Ah yes, you're right - it's not used. I think I refactored a bit and made this redundant, but didn't remove it. Good spot.

PR #1062, opened 2020-11-03
feat: log when users view visualisations

samuelhwilliams:
### Description of change
Add entries to our event log when users view (quicksight and datastudio)
visualisations so that we can understand how users are interacting with
our catalogue and which visualisations are popular/not. The eventlog is
automatically scraped by data-flow and returned to workspace as a master
dataset, so no further action needed for this.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/views.py
+++ b/dataworkspace/dataworkspace/apps/applications/views.py
@@ -45,6 +45,7 @@ from dataworkspace.apps.applications.utils import (
     application_options,
     get_quicksight_dashboard_name_url,
     sync_quicksight_permissions,
+    log_visualisation_view,
 )
 from dataworkspace.apps.applications.spawner import get_spawner
 from dataworkspace.apps.applications.utils import stop_spawner_and_application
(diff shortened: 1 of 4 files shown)


dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -274,8 +275,28 @@ def visualisation_link_html_view(request, link_id):
...
+        EventLog.objects.create(
+            user=request.user,
+            event_type=EventLog.TYPE_VIEW_DATASTUDIO_VISUALISATION,
+            content_type=ContentType.objects.get_for_model(VisualisationLink),
+            object_id=visualisation_link.id,
+            extra={
+                "name": visualisation_link.name,
+                "visualisation_identifier": visualisation_link.identifier,
+            },
+        ).save()
```
abbas123456, line 299: Could you factor this out into a seperate function as its duplicated code at the moment other than the `event_type`

samuelhwilliams: Yes - I didn't bother because it's very minor duplication (<3 instances) .. but will do anyway.

abbas123456 (approved):
Looks good, just one small refactoring suggestion

sekharpanja (approved):
Looks good to me, apart from what Mo mentioned.

PR #1064, opened 2020-11-05
fix: match uuids in url paths for catalogue views

niross:
This fix ensures a 404 is returned if an invalid uuid is provided in a url. Without this fix a validation error is thrown.

E.g. https://sentry.ci.uktrade.digital/dit/data-workspace/issues/27049/activity/


samuelhwilliams (approved):
Maybe just a very quick test for one/some of these routes?

PR #1066, opened 2020-11-05
hide unpublished master datasets in related data

sekharpanja:
### Description of change
This PR hides unpublished master datasets in related data section of datacuts.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/views.py
+++ b/dataworkspace/dataworkspace/apps/datasets/views.py
@@ -531,9 +531,9 @@ class DatasetDetailView(DetailView):
         for query in custom_queries:
             query_tables.extend([qt.table for qt in query.tables.all()])
 
-        ds_tables = SourceTable.objects.filter(table__in=query_tables).prefetch_related(
-            'dataset'
-        )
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -534,7 +534,9 @@ def get_context_data(self, **kwargs):
         ds_tables = SourceTable.objects.filter(table__in=query_tables).prefetch_related(
             'dataset'
         )
-        related_masters = [ds_table.dataset for ds_table in ds_tables]
+        related_masters = [
+            ds_table.dataset for ds_table in ds_tables if ds_table.dataset.published
+        ]
```
samuelhwilliams, line 539: Can this filter just be added to the query above? e.g. `SourceTable.objects.filter(dataset__published=True, table__in=query_tables)`

samuelhwilliams: Probably not if we want superusers to still see unpublished datasets here ... (although code will need updating still if that's the case, as this won't do it, I think?)

samuelhwilliams (approved):

PR #1071, opened 2020-11-10
feature: choose tools cpu and memory in separate page

abbas123456:
### Description of change

Currently the amount of cpu and memory when launching a tool is chosen using a dropdown list. This makes it difficult to show extra content explaining the different values.

This PR replaces the dropdown with a separate page with radio buttons that gives more details about the different options and allows the user to save their preference for a particular tool by utilising a new model called UserToolSizeChoice.

https://trello.com/c/iGqImZA8/1015-choose-tools-cpu-and-memory-on-separate-page
### Checklist


dataworkspace/dataworkspace/apps/api_v1/views.py
```diff
@@ -214,8 +219,8 @@ def application_api_PUT(request, public_host):
             public_host=public_host,
             state='SPAWNING',
             single_running_or_spawning_integrity=public_host,
-            cpu=cpu,
-            memory=memory,
+            cpu=int(cpu * 1024),
+            memory=int(memory * 1024),
```
michalc, line 223: Perhaps a minor thing if this always works... but I always fear something odd depending on floating point arithmetic + truncation to ints. Is it guaranteed that we won't end up one less than we expect? Essentially, I would lean on the side of paranoia, and use the integer values everywhere, e.g. where 1024 is 1 CPU, as now.

abbas123456: Yeah I did it this way to avoid dividing the megabytes by 1024 in the templates or having a separate mapping but I have now noticed the `spawner_memory` and `spawner_cpu` template filters for this exact purpose so I will use that :)

dataworkspace/dataworkspace/apps/applications/models.py
```diff
@@ -287,3 +288,71 @@ class Meta:
         proxy = True
         verbose_name = 'Application report'
         verbose_name_plural = 'Application report'
+
+
+SizeConfig = namedtuple('SizeConfig', ['name', 'memory', 'cpu', 'description'])
+
+
+class UserToolSizeChoice(models.Model):
```
samuelhwilliams, line 296: I wonder if `UserToolConfiguration` or `UserToolOptions` is a more generic name, in case there are other options that we might users to be able to set in the future (e.g. with or without internet access).

dataworkspace/dataworkspace/apps/applications/models.py
```diff
@@ -287,3 +288,71 @@ class Meta:
...
+        ),
+        SIZE_EXTRA_LARGE: SizeConfig(
+            dict(_SIZES)[SIZE_EXTRA_LARGE],
+            30720,
+            4096,
+            'Allows up to 4 parallel processes for faster analysis, and supports datasets of up to 30 gigabytes in memory.',
+        ),
+    }
+
+    user = models.ForeignKey(get_user_model(), on_delete=models.PROTECT)
```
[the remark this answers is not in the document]
abbas123456: There's no plan on getting all choices for a particular user so this is fine

dataworkspace/dataworkspace/apps/applications/models.py
abbas123456: Ok, `UserToolConfiguration` sounds good

dataworkspace/dataworkspace/apps/applications/urls.py
```diff
@@ -17,4 +18,14 @@
         login_required(quicksight_start_polling_sync_and_redirect),
         name='quicksight_redirect',
     ),
+    path(
+        'quicksight/redirect',
+        login_required(quicksight_start_polling_sync_and_redirect),
+        name='quicksight_redirect',
+    ),
+    path(
+        'choose-size/<str:tool_name>/',
```
samuelhwilliams, line 27: Maybe this should likewise be `configure/<str:tool_name>/`?

abbas123456: Well it depends on whether all future configuration such as internet access will be managed on the same page. Maybe changing this to `configure-size` for now will be better?

dataworkspace/dataworkspace/templates/user_tool_size_configuration_form.html
```diff
@@ -0,0 +1,38 @@
...
+                            {% with form.instance.SIZE_CONFIGS|get_key:value as config %}
+                                <div class="govuk-radios">
+                                    <div class="govuk-radios__item">
+                                        <input class="govuk-radios__input" id="id_size_{{ value }}" name="size" {% if value == form.size.value %}checked {% endif %} type="radio" value="{{ value}}" aria-describedby="id_size_{{ value }}_hint">
+                                        <label class="govuk-label govuk-radios__label" for="id_size_{{ value }}">
+                                            {{ config.name }} {% if config.name == form.instance.default_config.size_config.name %}(default){% endif %}
+                                        </label>
+                                        <div id="id_size_{{ value }}_hint" class="govuk-hint govuk-radios__hint">
+                                            <p class="govuk-hint">{{ config.cpu|spawner_cpu }} CPU, {{ config.memory|spawner_memory }} memory</p><p class="govuk-hint">{{ config.description}}</p>
+                                        </div>
```
samuelhwilliams, line 26: Discussed on slack - this doesn't look quite right, so will need to tweak it to visually match the prototype rather than copy the prototype's html directly here.

dataworkspace/dataworkspace/tests/applications/test_views.py
```diff
@@ -290,3 +291,119 @@ def test_view_starts_celery_polling_job(self):
                 poll_for_user_creation=True,
             )
         ]
+
+
+class TestToolsPage:
+    @pytest.mark.django_db
+    def test_user_with_no_size_config_shows_default_config(self):
+        factories.ApplicationTemplateFactory()
+        user = get_user_model().objects.create()
```
samuelhwilliams, line 300: Ah ... what's the difference between this and `UserFactory.create()`?

abbas123456: Probably nothing, `test_views.py` currently uses a mix of both so I just went for that one

samuelhwilliams (approved):
LGTM 👍

PR #1073, opened 2020-11-10
Adding sort feature to dataset browser

sekharpanja:
### Description of change

This PR adds datasets search/filter page, the capability to sort by the following:
            - `Relevance`: '-search_rank,name', 
            - `Newest first`: '-published_at',
            - `Oldest first`: 'published_at',
            - `Alphabetical (A to Z)`: 'name'

image

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/views.py
+++ b/dataworkspace/dataworkspace/apps/datasets/views.py
@@ -196,6 +196,7 @@ def get_datasets_data_for_user_matching_query(
         'source_tag_ids',
         'purpose',
         'published',
+        'published_at',
     ).annotate(has_access=BoolOr('_has_access'))
 
     return datasets.values(
@@ -208,6 +209,7 @@ def get_datasets_data_for_user_matching_query(
         'source_tag_ids',
         'purpose',
         'published',
+        'published_at',
         'has_access',
     )
 
@@ -297,6 +299,7 @@ def get_visualisations_data_for_user_matching_query(
         'source_tag_ids',
         'purpose',
         'published',
+        'published_at',
     ).annotate(has_access=BoolOr('_has_access'))
 
     return visualisations.values(
@@ -309,6 +312,7 @@ def get_visualisations_data_for_user_matching_query(
         'source_tag_ids',
         'purpose',
         'published',
+        'published_at',
         'has_access',
     )
 
@@ -321,10 +325,12 @@ def _matches_filters(data, access: bool, use: Set, source_ids: Set):
     )
 
 
-def sorted_datasets_and_visualisations_matching_query_for_user(query, use, user):
+def sorted_datasets_and_visualisations_matching_query_for_user(
+    query, use, user, sort_by
+):
     """
     Retrieves all master datasets, datacuts, reference datasets and visualisations (i.e. searchable items)
-    and returns them, sorted by desc(search_rank), asc(name).
+    and returns them, sorted by incoming sort field, default is desc(search_rank).
     """
     master_and_datacut_datasets = get_datasets_data_for_user_matching_query(
         DataSet.objects.live(), query, use, user=user, id_field='id'
@@ -339,10 +345,13 @@ def sorted_datasets_and_visualisations_matching_query_for_user(query, use, user)
     )
 
     # Combine all datasets and visualisations and order them.
+
+    sort_fields = sort_by.split(',')
+
     all_datasets = (
         master_and_datacut_datasets.union(reference_datasets)
         .union(visualisations)
-        .order_by('-search_rank', 'name')
+        .order_by(*sort_fields)
     )
 
     return all_datasets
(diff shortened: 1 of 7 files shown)


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -335,10 +341,16 @@ def sorted_datasets_and_visualisations_matching_query_for_user(query, use, user)
     )
 
     # Combine all datasets and visualisations and order them.
+
+    if sort_by == '':
+        sort_by = '-search_rank,name'
```
samuelhwilliams, line 346: I think this doesn't feel ideal - it could get out of sync if we change the option in the form. Can we use a default/initial value in the form field to prevent the need for this?

dataworkspace/dataworkspace/templates/datasets/select.html
```diff
@@ -0,0 +1,19 @@
+{% with id=widget.attrs.id %}
+<div class="govuk-form-group">
+  <span class="govuk-label">
+    Sort by
```
samuelhwilliams, line 4: I think it would be nice if this text can come from the field rather than be hard-coded, otherwise this isn't really a re-usable template. Are you able to set this on the field instance and then pull it through in the template?

dataworkspace/dataworkspace/templates/datasets/index.html
```diff
@@ -72,6 +72,7 @@ <h2 class="govuk-heading-l">Filters</h2>
   <div class="govuk-grid-column-two-thirds">
     <p class="govuk-body"><span id="search-results-count">{{ datasets.paginator.count }}</span> results</p>
     <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+    {{ form.sort }}
```
samuelhwilliams, line 75: It's not easy to tell from the design on the Trello ticket, but I think this field should be inline with the results text and right-aligned. Can we make that change?

dataworkspace/dataworkspace/templates/datasets/select.html
```diff
@@ -0,0 +1,19 @@
...
+        <optgroup label="{{ group_name }}">
+        {% endif %}
+        {% for option in group_choices %}
+        {% include option.template_name with widget=option %}
+        {% endfor %}
+        {% if group_name %}
+        </optgroup>
+        {% endif %}
+      {% endfor %}
+    </select>  </span>
```
samuelhwilliams, line 17: Very minor nit, but could we clean up the indenting here a bit so it's easier to read at a glance? e.g. maybe indenting inside the `if`/`for` and dropping the `span` onto the new line.

dataworkspace/dataworkspace/templates/datasets/index.html
samuelhwilliams: Nevermind - @sekharpanja showed me the design and this implementation is correct.

dataworkspace/dataworkspace/templates/datasets/select.html
sekharpanja: I ended up overriding widgets's `__init__`. Please suggest if there is a much better and straightforward way to achieve this.

samuelhwilliams: Most fields have a `label` attribute that provides this kind of text - can we use that?

dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -355,16 +364,23 @@ def find_datasets(request):
...
+
     if form.is_valid():
         query = form.cleaned_data.get("q")
         access = form.cleaned_data.get("access")
         use = set(form.cleaned_data.get("use"))
+        sort = (
+            form.cleaned_data.get("sort")
+            if form.cleaned_data.get("sort")
+            else sort_default[0]
+        )
```
samuelhwilliams, line 377: I think we can get rid of this (and `sort_default` var) if we add a `clean_sort` method to the `DatasetSearchForm` that sets the cleaned data to our desired default value if none is provided by the user, see https://docs.djangoproject.com/en/3.1/ref/forms/validation/#cleaning-a-specific-field-attribute

dataworkspace/dataworkspace/templates/datasets/index.html
samuelhwilliams, line 75: I would love to see this tucked up above the first result - e.g. floated right from the "x results" phrase ... feels like it fits a bit better there. But I know this isn't part of the current design, so take it or leave it ...

samuelhwilliams (commented):
A few comments, and I know you're looking into the selenium test already which is good. Looks like it's almost there though.

dataworkspace/dataworkspace/templates/datasets/index.html
sekharpanja: Before I saw the design, I went with the same idea. Data Hub has it the same way. Will make the change and discuss with Sana.

sekharpanja: image

samuelhwilliams: I think given data hub has it the same we could "re-use" their design. I think the margin/padding below is excessive though - can we reduce it? Adding `govuk-!-margin-bottom-0` to the form group for the select seems to work.

Screen Shot 2020-11-13 at 09 06 05

samuelhwilliams (approved):
Have agreed to add one test here around validating `sort` data in form submissions, but otherwise looks good! 👍

PR #1074, opened 2020-11-11
fix: catch IntegrityError when trying to create users from sso

abbas123456:
### Description of change

The sync sso task is currently failing to complete due to bad SSO data causing the creation of DW users to fail. This is now caught and logged to sentry instead.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -1130,14 +1130,17 @@ def _do_sync_activity_stream_sso_users():
             emails = obj['dit:emailAddress']
             primary_email = emails[0]
 
-            create_user_from_sso(
-                user_id,
-                primary_email,
-                emails,
-                obj['dit:firstName'],
-                obj['dit:lastName'],
-                check_tools_access_if_user_exists=True,
-            )
+            try:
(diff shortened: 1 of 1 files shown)


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1130,14 +1130,17 @@ def _do_sync_activity_stream_sso_users():
...
+                create_user_from_sso(
+                    user_id,
+                    primary_email,
+                    emails,
+                    obj['dit:firstName'],
+                    obj['dit:lastName'],
+                    check_tools_access_if_user_exists=True,
+                )
+            except IntegrityError:
+                logger.exception('Failed to create user record')
```
samuelhwilliams, line 1143: We'll probably want to "ignore" these in Sentry so we don't get slack alerts every 30 minutes? But yes seems sensible even so.

samuelhwilliams (approved):

dataworkspace/dataworkspace/apps/applications/utils.py
abbas123456: Once the SSO records are fixed we will stop seeing these errors, but if we don't log them we won't have any visibility of the bad data

PR #1078, opened 2020-11-12
Remove conda

samuelhwilliams:
### Description of change
Remove Conda from JupyterLab-Python in order to have a simpler and faster build that is more consistent with our other tools.

Leaving JupyterLab-R for now, as we intend to retire that shortly given the limited uptake and the availability of an alternative tool, RStudio.

### Checklist


abbas123456 (approved):
Good stuff

PR #1080, opened 2020-11-13
feature: rename SourceTag->Tag and add type field

abbas123456:
Currently only "source" tags are supported for datasets and reference datasets, this PR introduces a new "topic" tag but reuses the same model for both types, allowing us to add more types in the future without complicating existing search and filtering functionality.

This PR just changes the admin screen, the front end still only shows "Source" tag filters on the search results page. However, any "Topic" tags added to a dataset will be picked up by search which I don't necessarily think is a bad thing.

### Checklist


samuelhwilliams (approved):
Looks good, nice 👍

PR #1087, opened 2020-11-17
fix: move psycopg2 to requirements file

samuelhwilliams:
### Description of change
Rather than singling out psycopg2 for installation via the Dockerfile,
add it to our standard requirements.in tracking file for Python
requirements.

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,8 +26,7 @@ RUN apt-get update && \
         python3=3.7.3-1 \
         python3-dev=3.7.3-1 \
         python3-pip=18.1-5 \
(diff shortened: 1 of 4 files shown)


michalc (approved):
I have some vague memory that the two are not exactly equivalent. something like the OS-package version is pre-compiled, by the pypi one might need compilation, and so it's a bit slower to install? But if it's more or less the same, then all good

samuelhwilliams:
@michalc the pypi one does need compilation, that's true. There's a `psycopg2-binary` package that doesn't need compiling, but that does explicitly recommend compiling from source for production use-cases.

PR #1088, opened 2020-11-17
fix: unify theia/jupyterlab-python packages

samuelhwilliams:
## Summary
Probably best to view this one commit-by-commit.

## Commit 1
Standardises the set of Python packages installed into our JupyterLab-Python and Theia tools.

## Commit 2 (optional)
Slightly consolidates the directory structure for tools, putting all tools-related directories under `tools/` parent directory. We then pull out the duplicated `requirements.in/txt` files from jupyterlab-python and theia, moving them to the parent `tools/` dir so they can be shared easily. The idea here is that it should be easier to ensure these two tools stay in sync and always expose the same set of python libs.

We'll probably need to update quay.io build triggers after this is merged. Would we need to update anywhere else? @michalc 

## Commit 3
Functional tests have been a little flaky lately, I think since moving the Docker image to a debian base. It seems like celery beat sometimes doesn't terminate/clean up properly under test conditions when it gets sent a SIGTERM, leaving behind a pidfile that prevents a future celery beat from staying alive. When the future celery beat dies, our use of `parallel` also brings down the new server, so the tests aren't able to hit our application and fail. This slightly hacky workaround, manually removing the celery pidfile, seems to have stabilised things ... but open to better alternatives.


tools/python-setup.sh
```diff
@@ -0,0 +1,4 @@
+python3 -m pip install --upgrade pip setuptools wheel
+python3 -m pip install -r /root/requirements.txt
+python3 -m spacy download en
+python -m nltk.downloader -d /usr/local/share/nltk_data wordnet stopwords gutenberg
```
michalc, line 4: Just for consistency (since I suspect python and python3 are likely the same), should this be python3?

michalc (approved):

michalc:
> We'll probably need to update quay.io build triggers after this is merged. Would we need to update anywhere else? 

And to answer this, nothing leaps to mind: I think quay.io is it.

tools/python-setup.sh
samuelhwilliams: Ah good spot - thanks 👍

tools/jupyterlab-python/Dockerfile
```diff
@@ -104,10 +77,13 @@ RUN \
...
-COPY jupyterlab_database_access /jupyterlab_database_access
-COPY jupyterlab_template_notebooks /jupyterlab_template_notebooks
+COPY jupyterlab-python/jupyterlab_database_access /jupyterlab_database_access
+COPY jupyterlab-python/jupyterlab_template_notebooks /jupyterlab_template_notebooks
+COPY requirements.txt python-setup.sh /root/
 
 RUN \
+    chown -R jovyan:jovyan /usr/local && \
+    /root/python-setup.sh && \
     python -m pip install jupyterlab_template_notebooks/server/ && \
```
sekharpanja, line 87: This is not changed in the PR, but should this be python3 too?

sekharpanja (approved):
Looks good to me

PR #1089, opened 2020-11-17
fix: remove duplicates when query matches both source & catalogue item

abbas123456:
If a catalogue item has multiple sources and any of the source names are also contained in the catalogue item's description or name etc, searching for the source name returns duplicate results. This happens when a SearchVector is used for related models and can be fixed by using a string aggregate SearchVector

https://trello.com/c/4f4H9JZ0/1142-data-workspace-duplicate-results-in-some-cases

Example on dev to be checked once reviewed and deployed to dev:
https://analysisworkspace.dev.uktrade.io/datasets/?q=advisers&sort=-search_rank%2Cname

Screenshot 2020-11-17 at 15 04 29

### Checklist


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -136,7 +137,9 @@ def get_datasets_data_for_user_matching_query(
     search = (
         SearchVector('name', weight='A', config='english')
         + SearchVector('short_description', weight='B', config='english')
-        + SearchVector('source_tags__name', weight='B', config='english')
+        + SearchVector(
+            StringAgg('source_tags__name', delimiter=' '), weight='B', config='english'
```
samuelhwilliams, line 141: Maybe a `\n` delimiter is slightly better? I think it's possibly we could have source tags that have spaces in them?

samuelhwilliams (approved):

PR #1090, opened 2020-11-18
move celery workers into separate docker container

sekharpanja:
### Description of change
This PR is the first step in moving Data Workspace celery workers to a separate service. It sets up a docker container, based on data-workspace image, for celery workers.

### Checklist

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -14,6 +14,16 @@ services:
       - "data-workspace-redis"
     volumes:
       - ./dataworkspace:/dataworkspace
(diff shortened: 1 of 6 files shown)


docker-compose-dev.yml
```diff
@@ -12,8 +12,19 @@ services:
     links:
       - "data-workspace-postgres"
       - "data-workspace-redis"
+      - "data-workspace-celery"
```
samuelhwilliams, line 15: Do we need this? It's only in this file

samuelhwilliams (approved):

PR #1093, opened 2020-11-23
feat(visualisations): download vis cloudwatch logs

niross:
Allow visualisation developers to download the cloudwatch log of the latest run of a commit.
 
- Adds the ability to download a github visualisation commit once it has been previewed (or is live in prod).
- If no events exist a file with an error message is downloaded instead. This is to avoid getting a 404 in the browser window the link was clicked on in.

This branch is currently deployed to catalogue-dev for testing

Note: Will need a new env var, `VISUALISATION_CLOUDWATCH_LOG_GROUP`, added to each environment before release.

Closes: https://trello.com/c/knxIRoKv/397-epic-visualisation-logs-more-visible-to-help-visualisation-creators-debug-needs-a-kick-off-session-to-size-and-understand-resour


michalc:
Ah: you should probably be able to set the env var in `ecs_main_admin_container_definitions.json` (assuming the log group is available in the terraform, but I think it should be)

sekharpanja (approved):

PR #1096, opened 2020-11-24
feat add pgaudit

niross:
Add pgaudit to docker and rds dataset dbs

- Logging is turned off by default
- New init script for docker to install the pgaudit library
- New terraform resource to create the necessary parameter group for postgres10


infra/rds_datasets.tf
```diff
@@ -0,0 +1,21 @@
...
+  parameter {
+    name  = "shared_preload_libraries"
+    value = "pgaudit"
+    apply_method = "pending-reboot"
+  }
+
+  parameter {
+    name  = "pgaudit.role"
+    value = "rds_pgaudit"
+  }
```
samuelhwilliams, line 15: Should we also set `pgaudit.log` here or is that not necessary?

samuelhwilliams: I'm actually thinking now that we don't need `pgaudit.role`, just `pgaudit.log`. `role` is for per-object logging, and I think we can rely on session logging ... so we could remove this?

niross: > Should we also set `pgaudit.log` here or is that not necessary?

We can. By default it's turned off which I thought is what we wanted?

samuelhwilliams: Ah ok - I just meant to explicitly set it to `none`. If it defaults to that it's not too important, just more explicit maybe.

postgres/entrypoint.sh
```diff
@@ -11,4 +11,4 @@ chmod 0600 ssl.key ssl.crt
 export POSTGRES_USER=postgres
 export POSTGRES_PASSWORD=postgres
 
-docker-entrypoint.sh "$@" -c ssl=on -c ssl_cert_file=/ssl.crt -c ssl_key_file=/ssl.key
+docker-entrypoint.sh "$@" -c shared_preload_libraries=pgaudit -c ssl=on -c ssl_cert_file=/ssl.crt -c ssl_key_file=/ssl.key
```
samuelhwilliams, line 14: Do we also need to set it to emit via csvlog?

niross: we do and I have

samuelhwilliams (approved):

PR #1100, opened 2020-11-25
Celery ECS service

samuelhwilliams:
### Description of change
Adds a separate ECS service to run celery workers for Data Workspace. **Commit-by-commit review probably easiest.**

This re-uses a lot of resources from the admin service, as it runs the same code and in theory needs almost all of the same access to external resources (e.g. redis, database, gitlab, etc). The notable exception is that the admin service SG allows HTTP/S in, which we theoretically would not want for celery. The celery service receives no public IP and so should not be routable from the Internet. I feel like adding a new security group with separate controls might increase the risk of not keeping the celery workers in sync with the main application in terms of access to external resources, potentially breaking a future release if the security group is only updated for the admin app and not the celery service.

The first commit is a refactor to prevent duplication of the main block of the template variables. There are some minor differences between the two de-duplicated vars for the template, but I think these were inconsequential. It bears close inspection during code review though.

### DIFF
--- a/infra/ecs_main_admin.tf
+++ b/infra/ecs_main_admin.tf
@@ -150,7 +150,7 @@ resource "aws_ecs_task_definition" "admin" {
 data "template_file" "admin_container_definitions" {
   template = "${file("${path.module}/ecs_main_admin_container_definitions.json")}"
 
-  vars = "${merge(local.admin_container_vars, map("container_command", "/dataworkspace/start.sh"))}"
+  vars = "${merge(local.admin_container_vars, map("container_command", "[\"/dataworkspace/start.sh\"]"))}"
 }
 
 data "external" "admin_current_tag" {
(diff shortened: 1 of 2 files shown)


infra/ecs_main_admin.tf
```diff
@@ -90,8 +23,8 @@ data "template_file" "admin_container_definitions" {
     environment = "${var.admin_environment}"
 
     #notebooks_bucket = "${var.appstream_bucket}"
```
sekharpanja, line 25: Tiny one.. looks like this can be removed.

infra/ecs_main_admin.tf
```diff
@@ -177,92 +147,80 @@ resource "aws_ecs_task_definition" "admin_store_db_creds_in_s3" {
...
+    container_name = "jupyterhub-admin"
+  }
+}
 
-    sentry_dsn = "${var.sentry_dsn}"
+resource "aws_ecs_service" "admin_celery" {
+  name            = "${var.prefix}-admin-celery"
+  cluster         = "${aws_ecs_cluster.main_cluster.id}"
+  task_definition = "${aws_ecs_task_definition.admin_celery.arn}"
+  desired_count   = 2
```
sekharpanja, line 170: Do we need 2 instances or 1 would be enough?

sekharpanja (approved):
Looks good to me, couple of minor comments.

infra/ecs_main_admin.tf
samuelhwilliams: At the moment we have 2. I think 1 is generally enough, but it's nice to have 2 so that we've got some redundancy.

sekharpanja (approved):
Looks good to me

PR #1114, opened 2020-12-03
reset data explorer creds when changing master dataset permissions

sekharpanja:
### Description of change
This PR is to reset user credentials that are cached when master dataset permissions are changed.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/accounts/admin.py
+++ b/dataworkspace/dataworkspace/apps/accounts/admin.py
@@ -29,6 +29,7 @@ from dataworkspace.apps.applications.utils import (
 )
 from dataworkspace.apps.eventlog.models import EventLog
 from dataworkspace.apps.eventlog.utils import log_permission_change
+from dataworkspace.apps.explorer.schema import clear_schema_info_cache_for_user
 
 
 class AppUserCreationForm(forms.ModelForm):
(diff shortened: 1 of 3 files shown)


niross (approved):

dataworkspace/dataworkspace/apps/accounts/admin.py
```diff
@@ -430,6 +432,7 @@ def log_change(event_type, permission, message):
             )
             if dataset.type == DataSetType.MASTER.value:
                 update_quicksight_permissions = True
+            clear_schema_info_cache = True
```
samuelhwilliams, line 435: This probably could be indented? We only need to reset for master dataset changes.

dataworkspace/dataworkspace/apps/datasets/admin.py
```diff
@@ -307,6 +310,10 @@ def save_model(self, request, obj, form, change):
                 f"Removed dataset {obj} permission",
             )
             changed_user_sso_ids.add(str(user.profile.sso_id))
+            clear_schema_info_cache = True
+
+        if clear_schema_info_cache:
+            clear_schema_info_cache_for_user(user)
```
samuelhwilliams, line 316: This doesn't feel right ... this is just an arbitrary one of the affected users?

PR #1124, opened 2020-12-10
feat: Sync pgaudit logs from rds to admin db

niross:
### Description of change

- Sync logs via the rds api when not running locally.
- Sync via a shared log directory in tests or when running locally.
- Sync job runs every 5 minutes, but could be made more frequent if deemed necessary.
- Integration tests were very fiddly and I wasn't able to log in and sync logs as a new user so we will have to suffice with using the postgres user for now

### Checklist


.envs/test.env
```diff
@@ -95,4 +95,5 @@ EXPLORER_DEFAULT_CONNECTION=my_database
 
 DISABLE_CELERY_BEAT_SCHEDULE=True
 
-PGAUDIT_LOG_SCOPES='ALL'
+PGAUDIT_LOG_SCOPES='ALL, -MISC'
+PGAUDIT_LOG_TYPE = 'docker'
```
samuelhwilliams, line 99:
```suggestion
PGAUDIT_LOG_TYPE=docker
```

Should this be the same as the sample? I'm not sure if this needs the spaces/quotes.

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1184,3 +1188,153 @@ def fetch_visualisation_log_events(log_group, log_stream):
...
+
+    logger.info('Syncing pgaudit logs between %s and %s', from_date, to_date)
+
+    db_name_map = {v['NAME']: k for k, v in settings.DATABASES.items()}
+    databases = Database.objects.all()
+
+    if settings.PGAUDIT_LOG_TYPE == 'docker':
+        logs = _fetch_docker_pgaudit_logs(from_date, to_date)
+    else:
+        logs = _fetch_rds_pgaudit_logs(from_date, to_date)
```
samuelhwilliams, line 1294: Nice separation of handling for these two options 👍

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1184,3 +1188,153 @@ def fetch_visualisation_log_events(log_group, log_stream):
...
+
+        valid_log_lines.append(log_line)
+
+    return valid_log_lines
+
+
+def _fetch_docker_pgaudit_logs(from_date, to_date):
+    log_dir = '/var/log/postgres'
+    logs = []
+    for filename in os.listdir(log_dir):
```
samuelhwilliams, line 1233: How does this fetch the logs from the postgres container? I think I'm missing something...

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1184,3 +1188,153 @@ def fetch_visualisation_log_events(log_group, log_stream):
...
+        logger.info(
+            'Created log record for user %s in db %s',
+            log['user_name'],
+            log['database_name'],
+        )
+
+    cache.set('query_tool_logs_last_run', to_date)
+
+
+@celery_app.task(autoretry_for=(redis.exceptions.LockError,))
```
samuelhwilliams, line 1337: Would it be better to just drop these if they don't acquire the lock? Since they're re-scheduled every 5 minutes anyway. I wonder if there might be some strange case where the task takes longer than 5 minutes, and there's always just some that keeping hitting lock errors and getting re-queued, so our backlog builds up continuously over time. Presumably this task doesn't take 5 minutes ... but if we started reducing the interval significantly (E.g. 1 minute, or 30 seconds), then maybe that's a more realistic case?

docker-compose-test.yml
```diff
@@ -26,17 +26,25 @@ services:
       - "testvisualisation-a.dataworkspace.test:127.0.0.1"
       - "testvisualisation-b.dataworkspace.test:127.0.0.1"
       - "api.ecr.my-region-1.amazonaws.com:127.0.0.1"
+    volumes:
+      - db-logs-dev:/var/log/postgres
```
samuelhwilliams, line 30: Ah this is how... 👍  Neat

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1184,3 +1188,153 @@ def fetch_visualisation_log_events(log_group, log_stream):
...
+        try:
+            db_user = ApplicationInstanceDbUsers.objects.get(
+                db=database, db_username=log['user_name']
+            )
+        except ApplicationInstanceDbUsers.DoesNotExist:
+            logger.info(
+                'Skipping log entry for user %s on db %s as no matching user could be found',
+                log['user_name'],
+                log['database_name'],
+            )
```
samuelhwilliams, line 1318: It looks like we need to do something about Data Explorer - queries are just getting skipped right now as they don't fill info into ApplicationInstanceDbUsers ... 🤔

dataworkspace/dataworkspace/apps/applications/utils.py
niross: Yes makes sense. I've removed the retry.

dataworkspace/dataworkspace/apps/applications/utils.py
niross: Ah yeah good catch, i'll see what I can do

dataworkspace/dataworkspace/apps/core/models.py
```diff
@@ -102,3 +102,10 @@ class Meta:
 
     def __str__(self):
         return f'{self.memorable_name}'
+
+
+class DatabaseUser(TimeStampedModel):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    owner = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    username = models.CharField(max_length=256)
+    deleted = models.DateTimeField(null=True)
```
samuelhwilliams, line 111: Possibly `deleted_date` to more closely echo the `created_date` `modified_date` fields on TimeStampedModel?

dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -55,14 +55,19 @@ def postgres_user(stem, suffix=''):
...
+    username = (
         'user_'
         + re.sub('[^a-z0-9]', '_', stem.lower())[:max_email_length]
         + '_'
         + unique_enough
         + suffix
     )
 
+    if user is not None:
+        DatabaseUser.objects.create(owner=user, username=username)
```
samuelhwilliams, line 67: I'm not sure this is the right place - feels better to do this closer to where the role is created? I think there are a few places this could fail in between, which would maybe be bad/misleading. This is just generating a possible name, rather actual saying that name is a valid credential for use.

samuelhwilliams: I think probably right at/near the end of `new_private_database_credentials` is better? Once all the roles have been creating, granted, and collected?

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -520,6 +525,10 @@ def _do_delete_unused_datasets_users():
             usename for usename in usenames if usename not in in_use_usenames
         ]
 
+        DatabaseUser.objects.filter(username__in=not_in_use_usernames).update(
+            deleted_date=datetime.datetime.now()
+        )
+
```
samuelhwilliams, line 531: Happy to be challenged on this, but again I wonder if it would be better to do this individually after the user is actually dropped (line 600), so that we are sure the specific credential has been fully destroyed? I suppose it's possible that we could mark 10 users as deleted here, but the first one fails and therefore none of them are actually dropped? Possibly relevant as we do seem to get quite a few errors around not being able to drop users because of dependent objects.

niross: Ok yeah makes sense. one thing to note is I don't think the application db users are ever dropped weirdly.

samuelhwilliams (approved):

samuelhwilliams:
🎉

PR #1127, opened 2020-12-14
feature: store query results temporarily in a table when executing and fetch from there

abbas123456:
### Description of change

When a query is executed, the results are now stored in a table and fetched from there rather than returning the results directly by executing the query. 

This also deletes `DownloadQueryView` and its associated tests/templates etc

https://trello.com/c/jkfQNB6U/1085-use-tables-views-for-data-explorer-queries

### Checklist


dataworkspace/dataworkspace/tests/explorer/test_views.py
```diff
@@ -218,66 +218,27 @@ def test_cannot_post_to_another_users_query(self, staff_user, staff_client):
...
-    def test_empty_playground_renders(self, staff_user, staff_client):
+    @pytest.fixture(scope='function', autouse=True)
+    def setUp(self, staff_user_data):
+        suffix = stable_identification_suffix(
+            staff_user_data["HTTP_SSO_PROFILE_USER_ID"], short=True
+        )
+        schema_and_user_name = f'{USER_SCHEMA_STEM}{suffix}'
+        for alias in self.databases:
+            with connections[alias].cursor() as cursor:
+                cursor.execute(f'CREATE SCHEMA IF NOT EXISTS {schema_and_user_name}')
```
abbas123456, line 232: Not really sure why this is suddenly needed, the `with user_explorer_connection(user_connection_settings) as conn:` in utils.py execute_query should handle this. Without this a 500 error is raised because the _user_xxxx schema doesn't exist. Any pointers would be much appreciated

samuelhwilliams: I wonder if adding `pytest.mark.django_db(transaction=True)` might stop this error being thrown. I don't fully understand what that does in tests, but it _does_ tend to resolve weird things and it _might_ help here?

I should understand it better ...

dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -250,45 +251,91 @@ def __str__(self):
...
+    1082: 'date',
+    1083: 'time',
+    1114: 'timestamp with time zone',
+    1184: 'timestamp with time zone',
+    1186: 'timestamp with time zone',
+    1266: 'time',
+    1700: 'numeric',
+    2950: 'uuid',
+    3802: 'jsonb',
+}
```
samuelhwilliams, line 280: I think we're better off using `psycopg2.extensions.string_types` here. It's a dictionary that already contains these mappings. Eg:

```
In [10]: psycopg2.extensions.string_types[3802]
Out[10]: <psycopg2._psycopg.type 'JSONB' at 0x110b456b0>

In [11]: psycopg2.extensions.string_types[3802].name
Out[11]: 'JSONB'
```

dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -250,45 +251,91 @@ def __str__(self):
...
+            # This is required to handle multiple select columns with the same name.
+            # It adds a prefix of col_x_ to each column returned from the query and
+            # these prefixed column names are used to create a table containing the
+            # query results. The prefixes are removed when the results are returned.
+            cursor.execute(f'SELECT * FROM ({sql}) sq limit 0')
+
+            prefixed_sql_columns = [
+                f'"col_{i}_{col[0]}" {TYPE_CODES_REVERSED[col[1]]}'
+                for i, col in enumerate(cursor.description, 1)
+            ]
```
samuelhwilliams, line 313: I'm not sure what's better here. It feels like we're possibly introducing a class of errors wherein the user provides column aliases that are long, but not too long, then we make them too long by adding the prefix and break their query. At the moment this would probably throw an error that the user wouldn't be able to understand/resolve.

I think it's probably better for us to only provide names for columns that don't have one explicitly set, i.e. `?column?`?

dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -250,45 +251,91 @@ def __str__(self):
         return self.title
 
 
-def execute_query(query, user, page, limit, timeout, log_query):
+def tempory_query_table_name(user, query_log_id):
+    schema_name = f'{USER_SCHEMA_STEM}{db_role_schema_suffix_for_user(user)}'
+    return f'{schema_name}.tmp_query_{query_log_id}'
```
samuelhwilliams, line 256: Do you think it might be better to call the tables something like "_data_explorer_tmp_query_{query_log_id}"? These are going to show up in pgadmin, I think, and so making it clearer to users where these have come from might save a few questions down the line.

dataworkspace/dataworkspace/settings/base.py
```diff
@@ -461,6 +466,7 @@ def sort_database_config(database_list):
     'django',
     'dynamic_models',
     'data_explorer',
+    'tmp_query_',
```
samuelhwilliams, line 469: If we update the temp query table name don't forget to come here as well :)

dataworkspace/dataworkspace/apps/explorer/utils.py
abbas123456: yep makes sense

dataworkspace/dataworkspace/apps/explorer/utils.py
abbas123456: Well the prefixes are necessary for duplicate column names, not for anonymous ones. So if the user's query contains two fields called `id` we would need to prefix it. I can change it so that it only prefixes the duplicated column names.

dataworkspace/dataworkspace/apps/explorer/utils.py
abbas123456: Nope I haven't seen this, googled for a while and ended up copying it from the django source code. Will update it thanks

dataworkspace/dataworkspace/apps/explorer/utils.py
samuelhwilliams: Ah yeah you're right, my mistake. Maybe we can just do some detection and early reporting on column names that we're going to end up taking over the limit?

dataworkspace/dataworkspace/apps/explorer/utils.py
abbas123456: I've just tried this and it doesn't work unfortunately, STRING is not a valid postgres type:

```
>>> from psycopg2.extensions import string_types
>>> string_types[25].name
'STRING'
```

samuelhwilliams (approved):

dataworkspace/dataworkspace/apps/explorer/utils.py
abbas123456: Postgres automatically truncates identifier names so I don't think we need any validation, perhaps just a message to let the user know, in a similar way Postgres does:

**Explicit create statement:**

```
datasets=# create table foo ("saflkhfdhsakfhdaskjfhdkasjfhdksjafhdkjsahfdkajsfhkdjasfdsaffadsfdsaf" integer);
NOTICE:  identifier "saflkhfdhsakfhdaskjfhdkasjfhdksjafhdkjsahfdkajsfhkdjasfdsaffadsfdsaf" will be truncated to "saflkhfdhsakfhdaskjfhdkasjfhdksjafhdkjsahfdkajsfhkdjasfdsaffads"
```

**"create table as" statement:**

```
datasets=# create table foo as select 1 as saflkhfdhsakfhdaskjfhdkasjfhdksjafhdkjsahfdkajsfhkdjasfdsaffadsfdsaf;
NOTICE:  identifier "saflkhfdhsakfhdaskjfhdkasjfhdksjafhdkjsahfdkajsfhkdjasfdsaffadsfdsaf" will be truncated to "saflkhfdhsakfhdaskjfhdkasjfhdksjafhdkjsahfdkajsfhkdjasfdsaffads"
```

samuelhwilliams: Oh nice. Well totally ignore me then! I thought it died hard on some long identifiers, but clearly not these ones so let's roll on. 👍

dataworkspace/dataworkspace/tests/explorer/test_views.py
abbas123456: I'm afraid that didn't fix the failing tests in this case, it's a bit annoying that we have to explicitly create the schema for some of the tests but I did spend a while trying to figure it out and I don't think its worth investing any more time on it.

PR #1129, opened 2020-12-18
Feature/execute explorer queries in task

abbas123456:
### Description of change

This PR splits out the execute_query function into 1) a celery task that executes the query and stores the results and 2) a separate method fetch_query_results that fetches the results. The QueryLog model has been updated to include a state field which is updated by celery_task.

The celery task is currenly called synchronously via the delay().get() method and is behind a feature flag. The fetch_query_results method is intentionally assuming that the results table exists. In the future once the celery task is called asynchronously, the fetch_query_results method will be turned into a http endpoint and will need to be more robust in handling edge cases.

### Checklist


dataworkspace/dataworkspace/apps/explorer/models.py
```diff
@@ -117,6 +119,14 @@ def params_for_url(self):
...
 
 class QueryLog(models.Model):
+    STATE_RUNNING = QueryLogState.RUNNING.value
+    STATE_FAILED = QueryLogState.FAILED.value
+    STATE_COMPLETE = QueryLogState.COMPLETE.value
+    _STATE_CHOICES = (
+        (STATE_RUNNING, 'Running'),
+        (STATE_FAILED, 'Failed'),
+        (STATE_COMPLETE, 'Complete'),
+    )
```
samuelhwilliams, line 129: This feels kinda verbose ... I think we can inline it into the field as just `choices=[(tag, tag.value) for tag in QueryLogState]`? Feels easier to manage/maintain in the future if we add new states as well.

dataworkspace/dataworkspace/apps/explorer/models.py
```diff
@@ -126,6 +136,9 @@ class QueryLog(models.Model):
     run_at = models.DateTimeField(auto_now_add=True)
     duration = models.FloatField(blank=True, null=True)  # milliseconds
     connection = models.CharField(blank=True, null=True, max_length=128)
+    state = models.IntegerField(choices=_STATE_CHOICES, default=STATE_RUNNING)
+    rows = models.IntegerField(null=True, blank=True)
+    page = models.IntegerField(null=True, blank=True)
```
samuelhwilliams, line 141: Might be nice for these to not be nullable/blank? I think every query will have these because we add them, and then it's a more explicit/reliable record of exactly what data they received (e.g. rows 1000-1999 if it's page=2, rows=1000)?

dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -216,49 +215,15 @@ def remove_data_explorer_user_cached_credentials(user):
...
-        return [r[ix] for r in self.data]
-
-
-class ColumnHeader:
-    def __init__(self, title):
-        self.title = title.strip()
-
-    def __str__(self):
-        return self.title
+        self.query_log_id = query_log_id
```
samuelhwilliams, line 226: I think I can barely see the value in this class any more ... pretty much everything here is just on the QueryLog model. I think this is now only used in one place - returning from the execute_query function. Can we just drop it altogether? I think probably just returning the querylog ID from that function is maybe fine?

The extra attributes on this model are row_count and timeout. I think maybe adding row_count to QueryLog seems reasonable. Timeout we probably don't need to record ... or if we do we can just add that to QueryLog as well?

dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -290,79 +255,19 @@ def tempory_query_table_name(user, query_log_id):
...
 
 
-def _prefix_column(index, column):
-    return f'col_{index}_{column}'
+def fetch_query_results(query_log_id):
+    query_log = get_object_or_404(QueryLog, pk=query_log_id)
 
-
-def execute_query(query, user, page, limit, timeout):
+    user = query_log.run_by_user
```
samuelhwilliams, line 261: Think you mentioned at one point sticking to using `request.user` - did that end up being inconvenient or do you just think this is fine now (given we have `user_explorer_connection`  that should mean we have the DB still enforcing permissions/correctness?

dataworkspace/dataworkspace/apps/explorer/views.py
```diff
@@ -476,7 +479,28 @@ def query_viewmodel(
...
+                    query.final_sql(),
+                    query.connection,
+                    query.id,
+                    request.user.id,
+                    page,
+                    rows,
+                    timeout,
+                )
+
+            headers, data, query_log = fetch_query_results(res['query_log_id'])
```
samuelhwilliams, line 503: Ah sorry, this is probably slightly different than I expected. In my mind, with the feature flag toggle, the default path should use the exact same code as we currently have, and the flagged code path should have all new methods/code path. Working that way, we have confidence that we're not breaking the existing code. With this, we're sort of still bearing the risk of changing the code path for all users, which could be broken. It's possibly just something to think about for new work as I suspect it's a fair amount of faff to now go and try to restore all the original functions...

dataworkspace/dataworkspace/tests/conftest.py
```diff
@@ -151,3 +153,8 @@ def change_staticfiles_storage():
         STATICFILES_STORAGE='django.contrib.staticfiles.storage.StaticFilesStorage'
     ):
         yield
+
+
+@pytest.fixture(scope='session', autouse=True)
+def make_celery_eager():
+    celery_app.conf.task_always_eager = True
```
samuelhwilliams, line 160: Do we still need this now we've added the celery worker to docker-compose-test?

abbas123456: I would rather keep it as I don't think the unit tests should be calling celery or relying on the celery worker running for them to pass.

dataworkspace/dataworkspace/apps/explorer/views.py
abbas123456: I did think about this, I think the feature flag will be much more useful once the task is called asynchronously. At this stage having one path that combines the two functions and another path that calls them separately feels a bit unecessary.

dataworkspace/dataworkspace/apps/explorer/utils.py
abbas123456: I was planning on reviewing the security and making the appropriate changes in a separate PR / ticket

dataworkspace/dataworkspace/apps/explorer/utils.py
abbas123456: Yep agree completely, QueryLog already has a rows field and we don't need to store the timeout as it's only relevant when executing the user's query and not when fetching the records from the results table. Will remove this class.

dataworkspace/dataworkspace/apps/explorer/models.py
abbas123456: Agreed to keep it as it is to maintain consistency with other choice fields such as Dataset.type and Tag.type

dataworkspace/dataworkspace/apps/explorer/models.py
abbas123456: The rows field has to be nullable because this isn't available when the QueryLog object is first created. It gets populated later upon success

dataworkspace/dataworkspace/apps/explorer/tasks.py
```diff
@@ -50,3 +59,79 @@ def cleanup_temporary_query_tables():
...
+        sql = query_sql.rstrip().rstrip(';')
+
+        table_name = tempory_query_table_name(user, query_log.id)
+        try:
+            cursor.execute(f'SET statement_timeout = {timeout}')
+            # This is required to handle multiple select columns with the same name.
+            # It adds a prefix of col_x_ to duplicated column returned from the query and
+            # these prefixed column names are used to create a table containing the
+            # query results. The prefixes are removed when the results are returned.
+            cursor.execute(f'SELECT * FROM ({sql}) sq limit 0')
```
samuelhwilliams, line 96: If this errors, the user sees something like this: `column "foo" does not exist LINE 1: SELECT * FROM (select "foo") sq limit 0 ^`

I wonder ... if this query errors, is it safe to re-run just the bare query and return the error from that instead - so that the user isn't confused by the wrapping we're doing..?

dataworkspace/dataworkspace/apps/explorer/models.py
samuelhwilliams: Ah ok. I suppose we could put a `0` in there when we create the querylog, but that's not much different than keeping a NULL. Probably worse in fact. So yes OK 👍

dataworkspace/dataworkspace/apps/explorer/utils.py
samuelhwilliams: Ah yes as we discussed - cool 👍

dataworkspace/dataworkspace/apps/explorer/tasks.py
abbas123456: I did have a similar thought at the time which was to do a simple string replace to remove the `SELECT * FROM (` and `sq limit 0 ` from the error before returning to the user but wasn't sure about it. If that doesn't sound good enough to you I'm happy to do what you suggested

abbas123456: Have decided to extract the first part of the error message returned by postgres and return that to the user, e.g:

`column "foo" does not exist LINE 1: SELECT * FROM (select "foo") sq limit 0 ^`

becomes

`column "foo" does not exist`

samuelhwilliams: As per slack chat - go for the string replacement if it's easy and reliable.

samuelhwilliams (approved):

samuelhwilliams:
Good work - this ended up being a lot chunkier than we expected I think. 👏  Probably deserves some fairly thorough testing in staging ...

PR #1131, opened 2020-12-20
feat: extract tables from logs show related datasets

niross:
### Description of change

- Exctract tables from pgaudit logs and save in a new audit log table model
- Add links to related datasets to the audit log admin page

### Checklist


samuelhwilliams (approved):
Nice 👍 

I feel like we've trod this ground before and I forget ... but it might be nice to show the related master datasets on the tool query log list page as well as the detail page?

samuelhwilliams (approved):

PR #1142, opened 2020-12-23
Fix/more rds sync fixes

niross:
### Description of change

- Allow for very large sql queries in the pgaudit logs by upping the csv field size limit to the system maximum
- Truncate queries displayed on the admin list and detail pages so queries with millions of characters don't break the page
    - Currently set to 200 and 5000 respectively but happy to discuss

### Checklist


dataworkspace/dataworkspace/apps/datasets/admin.py
```diff
@@ -724,3 +738,18 @@ def has_delete_permission(self, request, obj=None):
...
+            return query[:length] + '...'
+        return query
+
+    def get_list_truncated_query(self, obj):
+        return self._truncate_query(obj.query_sql, 200)
+
+    get_list_truncated_query.short_description = 'Query SQL'
+
+    def get_detail_truncated_query(self, obj):
+        return self._truncate_query(obj.query_sql, 5000)
```
samuelhwilliams, line 753: Maybe we can bung these values into env-configurable options so we can bump them more easily if we do want to change them? These seem like reasonable values though.

niross: yes good call :+1:

samuelhwilliams (approved):

PR #1148, opened 2020-12-30
Add datasets aurora db and associated resources to terraform

abbas123456:
### Checklist


infra/rds_datasets.tf
```diff
@@ -0,0 +1,49 @@
+resource "aws_rds_cluster" "datasets" {
+  availability_zones            = ["${var.aws_availability_zones}"]
+  backup_retention_period       = "${var.datasets_rds_cluster_backup_retention_period}"
+  cluster_identifier            = "${var.datasets_rds_cluster_cluster_identifier}"
+  database_name                 = "${var.datasets_rds_cluster_database_name}"
+  db_subnet_group_name          = "${aws_db_subnet_group.datasets.name}"
+  engine                        = "aurora-postgresql"
+  engine_version                = "10.11"
```
samuelhwilliams, line 8: Possibly these should be local vars so that we can reuse them across the cluster and the cluster instance?

infra/rds_datasets.tf
```diff
@@ -0,0 +1,49 @@
...
+  master_password               = "${random_string.aws_rds_cluster_instance_datasets_password.result}"
+  master_username               = "${var.datasets_rds_cluster_master_username}"
+  storage_encrypted             = "${var.datasets_rds_cluster_storage_encryption_enabled}"
+  vpc_security_group_ids        = ["${aws_security_group.datasets.id}"]
+  skip_final_snapshot           = true
+
+  lifecycle {
+    ignore_changes = ["master_password"]
+  }
```
samuelhwilliams, line 17: This is probably fine. I think we could import the existing password - asking just because I don't kniw: is there a reason not to do that?

infra/rds_datasets.tf
```diff
@@ -0,0 +1,49 @@
...
+resource "aws_rds_cluster_instance" "datasets" {
+  cluster_identifier            = "${aws_rds_cluster.datasets.cluster_identifier}"
+  db_subnet_group_name          = "${aws_db_subnet_group.datasets.name}"
+  engine                        = "aurora-postgresql"
+  engine_version                = "10.11"
+  identifier                    = "${var.datasets_rds_cluster_instance_identifier}"
+  instance_class                = "${var.datasets_rds_cluster_instance_class}"
+  performance_insights_enabled  = "${var.datasets_rds_cluster_instance_performance_insights_enabled}"
+  promotion_tier                = 1
+  db_parameter_group_name       = "aurora-postgresql-10-pgaudit-db-parameter-group"
```
samuelhwilliams, line 29: Could we make/have we made a tech debt/maintenance ticket to set this up in TF?

samuelhwilliams (commented):
Looks good 👍

infra/rds_datasets.tf
abbas123456: Yes the existing password is already on my import plan, however, without this line terraform thinks the master_password attribute has changed even if they match and that will mark the database as pending-reboot if applied

infra/rds_datasets.tf
abbas123456: Yeah this is actually optional on the cluster_instance so I will delete it from there and add it to the lifecycle ignore_changes list to avoid upgrade attempts as explained here: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster_instance#engine_version

infra/rds_datasets.tf
abbas123456: Yes I wasn't really sure what's involved in creating this in terraform as it has hundred of parameters, do you know if we need to manually configure every single one?

samuelhwilliams: No idea I'm afraid. Hopefully not :/

abbas123456: Will create a task for this

samuelhwilliams (approved):
Great work 👏

PR #1151, opened 2020-12-31
test: add retries to integration tests

samuelhwilliams:
### Description of change
Some of our integration tests are flakey at the moment. While we should
address this, quick investigation hasn't yielded results and we'll need
to prioritise spending more time on it. For now, we'll just re-run any
failed integration tests a few times to try to bypass any transient
failures.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ docker-test-unit: docker-build
 
 .PHONY: docker-test-integration
 docker-test-integration: docker-build
-	docker-compose -f docker-compose-test.yml -p data-workspace-test run data-workspace-test pytest test/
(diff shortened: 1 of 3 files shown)


niross (approved):
Nice! Very helpful while we sort out what's going on with them :+1:

samuelhwilliams:
Well this is an unmitigated disaster. The retries just consistently fail as well... 🤔  Something must be fairly fishy with how the tests are running in circleci...

samuelhwilliams:
Doesn't actually work 😢 

I think it would be great for this to be fixed though ... the flaky integration tests are a bit of a nightmare.

PR #1167, opened 2021-01-14
feature: revoke temporary user privileges so that DROP USER succeeds

abbas123456:
The delete_unused_datasets_users task sometimes fails if the temporary user role has privileges. These need to be revoked before attempting to delete the role.

See https://sentry.ci.uktrade.digital/organizations/dit/issues/34154/ for an example error that this PR should fix

### Checklist


samuelhwilliams (approved):
Nice

PR #1183, opened 2021-01-20
Tweak ownership to make dash installable

michalc:
### Description of change

The python package dash installs into /usr/local. We already attempted to make all of it writable by the user, but it wasn't done quite right. Some files/folders were still owned by root since they were added in the Dockerfile as root.

### Checklist

### DIFF
--- a/tools/jupyterlab-python/Dockerfile
+++ b/tools/jupyterlab-python/Dockerfile
@@ -82,7 +82,6 @@ COPY jupyterlab-python/jupyterlab_template_notebooks /jupyterlab_template_notebo
 COPY requirements.txt python-setup.sh /root/
 
 RUN \
-    chown -R jovyan:jovyan /usr/local && \
     /root/python-setup.sh && \
     python3 -m pip install jupyterlab_template_notebooks/server/ && \
     jupyter serverextension enable --system --python jupyterlab_template_notebooks && \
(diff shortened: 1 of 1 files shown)


tools/jupyterlab-python/Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN \
         libxext6 \
         libxrender1 \
         man-db=2.8.5-2 \
-        openssl=1.1.1d-0+deb10u3 \
+        openssl=1.1.1d-0+deb10u4 \
```
samuelhwilliams, line 52: Not totally related to this PR, but ... It feels like we just have to constantly keep bumping these versions. Do we really need to pin them? If we're using a package manager it feels like we just have to accept that they want to give us. At the moment our builds can break at any moment because they release a new version of something we use and remove the old one. Do we need the level of oversight over version changes that is currently required by having to make these manual updates? Are we meant to be reviewing each new version as it comes out before bumping it?

samuelhwilliams (approved):
Change itself looks reasonable 👍  Side question though...

tools/jupyterlab-python/Dockerfile
michalc: > Are we meant to be reviewing each new version as it comes out before bumping it?

No, I don't think so (just like we don't review our other dependencies)

> At the moment our builds can break at any moment because they release a new version of something we use and remove the old one. Do we need the level of oversight over version changes that is currently required by having to make these manual updates

I guess I'm happy with the build breaking with a fairly clear reason: dependency has gone away, rather than what I think is the alternative with unpinning: it succeeding, but it has changed/broken something, but we don't know that, and so don't know where to start looking for the cause. (Or even, worse than that, since typically the rebuild would be due to something changing, so attention would be drawn to that change, which would be a red-herring). I guess I've been burned by unpinned dependencies too many times, and to me, while a bit faffy/manual/annoying, this is the way of less pain.

But, I don't think by any means "this is the one true way": can unpin

PR #1185, opened 2021-01-21
fix: add db_persistent_role to ApplicationInstanceDbUsers

abbas123456:
### Description of change

Adds a new field called db_persistent_role to the ApplicationInstanceDbUsers which is used at a later stage when deleting unused datasets users.

This is set when launching a tool or visualisation and allows us to reassign objects owned by the temporary db user.

https://trello.com/c/r3nnSGNs/1380-sentry-error-persistent-user-role-does-not-exist-when-deleting-unused-datasets-users

### Checklist


dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -294,6 +294,7 @@ def get_new_credentials(database_obj, tables):
             'db_host': database_data['HOST'],
             'db_port': database_data['PORT'],
             'db_user': db_user,
+            'db_role': db_role,
```
samuelhwilliams, line 297: I would maybe call this something more explicit like `db_persistent_role`? Roles and users are sort of similar concepts in Postgres so it's perhaps not immediately clear what the difference is.

samuelhwilliams (approved):
Looks good to me...

PR #1186, opened 2021-01-21
Remove the old design system form

samuelhwilliams:
### Description of change
In a recent ticket, the way that we render design system forms was rewritten to better support GOV.UK Design System components being rendered correctly (e.g. as "single question pages" and with error components). This left some tech debt as there were two forms claiming to render design system components.

This cleans up that debt and also further rationalises the way that design system fields/widgets are declared. The interface on this ticket more closely resembles standard django fields/widgets, where things like labels are declared on the field and the widget only takes rendering config.

Under the hood we do still need to pass some of this content to the widget (e.g. labels, errors) in order to render the components correctly according to the Design System.


niross (approved):
Good stuff :+1:

PR #1188, opened 2021-01-22
fix: create db creds in background

samuelhwilliams:
### Description of change
At the moment we create the DB credentials for the tool/visualisation
within the initiail request/response of the tool creation POST. This
step can take a little while though, so it would make sense to push it
into the background where the rest of the tool is setup so that the user
more quickly lands on the "tool loading" page with a spinner rather than
sitting for a second or so before anything even loads onto the page.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/api_v1/views.py
+++ b/dataworkspace/dataworkspace/apps/api_v1/views.py
@@ -1,4 +1,3 @@
-import datetime
 import json
 import logging
 import re
@@ -14,7 +13,6 @@ from psycopg2 import connect, sql
 
 from dataworkspace.apps.applications.models import (
     ApplicationInstance,
-    ApplicationInstanceDbUsers,
     ApplicationTemplate,
     UserToolConfiguration,
 )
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/apps/api_v1/views.py
```diff
@@ -172,32 +165,12 @@ def application_api_PUT(request, public_host):
 
     app_type = application_template.application_type
 
-    (source_tables, db_role_schema_suffix, db_user) = (
-        (
-            source_tables_for_user(request.user),
-            db_role_schema_suffix_for_user(request.user),
-            postgres_user(request.user.email),
-        )
+    db_role_schema_suffix = (
```
abbas123456, line 168: I don't think we need to generate the `db_role_schema_suffix` here as it's already being generated in spawner.py

abbas123456: Sorry I should have said we don't need to pass `app_schema` to the task as it gets generated there anyway

samuelhwilliams: Ok fixed up

abbas123456 (approved):

PR #1190, opened 2021-01-24
show code snippets for each source table on dataset page

sekharpanja:
### Description of change
This PR is to show code snippets for each table within dataset page in data workspace.

image

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/static/data-workspace.css
+++ b/dataworkspace/dataworkspace/static/data-workspace.css
@@ -221,3 +221,7 @@ Comment: https://github.com/alphagov/govuk-design-system-backlog/issues/28#issue
         float: right;
     }
 }
+
+.app-table_cell--no-border {
+  border-bottom: none;
+}
--- a/dataworkspace/dataworkspace/templates/datasets/master_dataset.html
+++ b/dataworkspace/dataworkspace/templates/datasets/master_dataset.html
@@ -62,20 +62,20 @@
         </thead>
         <tbody>
 
-        {% for data_link, _ in data_links_with_link_toggle %}
+        {% for data_link, _, code_snippets in data_links_with_link_toggle %}
           <tr class="govuk-table__row">
-            <td class="govuk-table__cell">{{ data_link.name }}</td>
-            <td class="govuk-table__cell">"{{ data_link.schema }}"."{{ data_link.table }}"</td>
-            <td class="govuk-table__cell">{{ data_link.get_frequency_display }}</td>
-            <td class="govuk-table__cell">{{ data_link.get_data_last_updated_date|default_if_none:"N/A" }}</td>
+            <td class="govuk-table__cell app-table_cell--no-border">{{ data_link.name }}</td>
+            <td class="govuk-table__cell app-table_cell--no-border">"{{ data_link.schema }}"."{{ data_link.table }}"</td>
+            <td class="govuk-table__cell app-table_cell--no-border">{{ data_link.get_frequency_display }}</td>
+            <td class="govuk-table__cell app-table_cell--no-border">{{ data_link.get_data_last_updated_date|default_if_none:"N/A" }}</td>
             {% if user.is_superuser %}
-              <td class="govuk-table__cell">
+              <td class="govuk-table__cell app-table_cell--no-border">
                 {% if data_link.get_google_data_studio_link and data_link.accessible_by_google_data_studio %}
                   <a href="{{ data_link.get_google_data_studio_link }}">Google Data Studio</a>
                 {% endif %}
               </td>
             {% endif %}
-            <td class="govuk-table__cell">
+            <td class="govuk-table__cell app-table_cell--no-border">
               {% if data_link.type == source_table_type %}
                 {% if has_access %}
                   <a class="govuk-link"
@@ -89,6 +89,22 @@
                   No preview available
               {% endif %}
             </td>
+            {% if code_snippets %}
+              <tr class="govuk-table__row">
+                <td colspan="6" class="govuk-table__cell">
+                  <details class="govuk-details-!-margin-bottom-2" data-module="govuk-details">
+                    <summary class="govuk-details__summary">
+                      <span class="govuk-details__summary-text">
+                        Code snippets <span class="govuk-visually-hidden">for "{{ data_link.schema }}"."{{ data_link.table }}"</span>
+                      </span>
+                    </summary>
+                    <div class="govuk-details__text">
+                      {% include 'partials/code_snippets.html' with code_snippets=code_snippets data_link=data_link %}
+                    </div>
+                  </details>
+                </td>
+              </tr>
+            {% endif %}
           </tr>
         {% endfor %}
         {% if not data_links_with_link_toggle %}
@@ -103,8 +119,6 @@
       {% endif %}
     </div>
   </div>
-
-  {% include 'partials/code_snippets.html' %}
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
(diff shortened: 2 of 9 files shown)


dataworkspace/dataworkspace/templates/datasets/master_dataset.html
```diff
@@ -89,6 +89,22 @@ <h2 class="govuk-heading-l govuk-!-margin-top-8">Dataset Details</h2>
...
+              <tr class="govuk-table__row">
+                <td colspan="6" class="govuk-table__cell">
+                  <details class="govuk-details--no-margin-bottom" data-module="govuk-details">
+                    <summary class="govuk-details__summary">
+                      <span class="govuk-details__summary-text">
+                        Code Snippets
+                      </span>
+                    </summary>
+                    <div class="govuk-details__text">
+                      {% include 'partials/code_snippets.html' with code_snippets=code_snippets %}
```
samuelhwilliams, line 102: I think, visually, we need to update this partial now - it feels like the `margin-top-8` adds way too much space above the header now we're inlining this. I think it can just be removed.

dataworkspace/dataworkspace/static/data-workspace.css
```diff
@@ -221,3 +221,15 @@ Comment: https://github.com/alphagov/govuk-design-system-backlog/issues/28#issue
...
+.govuk-table__cell--no-border {
+  padding: 10px 20px 10px 0;
+  border-bottom: none;
+  text-align: left;
+  vertical-align: top;
+}
+
+.govuk-details--no-margin-bottom {
+  margin-bottom: 10px;
+}
```
samuelhwilliams, line 235: I think we should rename these classes, as at the moment they are masquerading as official GOV.UK Design System classes, but actually they're our custom modifications. For that it's normally recommended to use an `app-` prefix. E.g. `app-table_cell--no-border`.

dataworkspace/dataworkspace/static/data-workspace.css
```diff
@@ -221,3 +221,15 @@ Comment: https://github.com/alphagov/govuk-design-system-backlog/issues/28#issue
...
     }
 }
+
+/* helpers */
+.govuk-table__cell--no-border {
+  padding: 10px 20px 10px 0;
+  border-bottom: none;
+  text-align: left;
+  vertical-align: top;
+}
```
samuelhwilliams, line 231: GOV.UK Design System uses BEM methodology to name CSS classes. This is a modifier on an element (a table cell), so only the attributes for that attribute should be included - which based on the name I think is just `border-bottom: none`? And maybe the class name should end `--no-bottom-border` to be more explicit?

dataworkspace/dataworkspace/static/data-workspace.css
```diff
@@ -221,3 +221,15 @@ Comment: https://github.com/alphagov/govuk-design-system-backlog/issues/28#issue
...
+
+.govuk-details--no-margin-bottom {
+  margin-bottom: 10px;
```
samuelhwilliams, line 234: I'm not sure we need this class - can we just use an override class from the Design System: `govuk-!-margin-bottom-<x>`? Where x is maybe something in the range 1-3, depending on what looks reasonable.

dataworkspace/dataworkspace/templates/datasets/master_dataset.html
samuelhwilliams: We're also going to now have multiple buttons that read the same text: "Open in Data Explorer". We should add some hidden text to these buttons for screenreaders to help disambiguate them. Something like "Open <first 50 rows of "public"."dataset"> in Data Explorer"?

dataworkspace/dataworkspace/templates/datasets/master_dataset.html
```diff
@@ -62,20 +62,20 @@ <h2 class="govuk-heading-l govuk-!-margin-top-8">Dataset Details</h2>
...
+        {% for data_link, _, code_snippets in data_links_with_link_toggle %}
           <tr class="govuk-table__row">
-            <td class="govuk-table__cell">{{ data_link.name }}</td>
-            <td class="govuk-table__cell">"{{ data_link.schema }}"."{{ data_link.table }}"</td>
-            <td class="govuk-table__cell">{{ data_link.get_frequency_display }}</td>
-            <td class="govuk-table__cell">{{ data_link.get_data_last_updated_date|default_if_none:"N/A" }}</td>
+            <td class="govuk-table__cell--no-border">{{ data_link.name }}</td>
+            <td class="govuk-table__cell--no-border">"{{ data_link.schema }}"."{{ data_link.table }}"</td>
+            <td class="govuk-table__cell--no-border">{{ data_link.get_frequency_display }}</td>
+            <td class="govuk-table__cell--no-border">{{ data_link.get_data_last_updated_date|default_if_none:"N/A" }}</td>
```
samuelhwilliams, line 70: After the CSS updates above, the class attribute here may look something like `govuk-table__cell app-table__cell--no-border`

samuelhwilliams (commented):
Looks good overall. It's maybe slightly squiffy to have a separate row for the code snippets as that feels less logically grouped together ... but I don't know a better alternative to use on this design :)

A few suggestions then I think should be good to go.

dataworkspace/dataworkspace/templates/datasets/master_dataset.html
```diff
@@ -92,14 +92,14 @@ <h2 class="govuk-heading-l govuk-!-margin-top-8">Dataset Details</h2>
             {% if code_snippets %}
               <tr class="govuk-table__row">
                 <td colspan="6" class="govuk-table__cell">
-                  <details class="govuk-details--no-margin-bottom" data-module="govuk-details">
+                  <details class="govuk-details-!-margin-bottom-10px" data-module="govuk-details">
```
samuelhwilliams, line 95: This looks like a typo - the override class is `govuk-!-margin-bottom-<0...9>`. See "Using the override classes" here: https://design-system.service.gov.uk/styles/spacing/

dataworkspace/dataworkspace/templates/partials/code_snippets.html
```diff
@@ -28,6 +28,7 @@ <h3 class="govuk-heading-m">SQL</h3>
         </pre>
       </div>
 
+      <div style="display: none">Open first 50 rows of "{{ data_link.schema }}"."{{ data_link.table }}" in Data Explorer</div>
       <a id="launch-data-explorer" class="govuk-button govuk-!-margin-top-5" href="{% url 'explorer:index' %}?sql={{ code_snippets.sql|quote_plus }}">Open in Data Explorer</a>
```
samuelhwilliams, line 32: Ah sorry, this is not quite what I meant. We want this as the button text. It needs to be the text of the actual button so that, when screenreaders read out the text of the button, they read out the full non-ambiguous details. It's useful to be aware that some/most screenreader users don't interact with a page in a top-down fashion, but jump around the page reading headings, links and buttons. If they're in "read buttons" mode, they may not read the text in the div just before it.

We can also use a Design System style to hide the text rather than roll our own display: none.

What we want, I think, is something like this:

```html
      <a id="launch-data-explorer" class="govuk-button govuk-!-margin-top-5" href="{% url 'explorer:index' %}?sql={{ code_snippets.sql|quote_plus }}">Open <span class="govuk-visually-hidden">first 50 rows of "{{ data_link.schema }}"."{{ data_link.table }}" </span>in Data Explorer</a>
```

samuelhwilliams (commented):
Looks good. I think this should be ready after you squash your fixup commits 👍

dataworkspace/dataworkspace/templates/partials/code_snippets.html
```diff
@@ -5,7 +5,7 @@
   {% if code_snippets %}
   <link rel="stylesheet" href="{% static 'assets/vendor/highlight/styles/a11y-light.css' %}">
 
-  <h2 class="govuk-heading-l govuk-!-margin-top-8">Code snippets</h2>
+  <h2 class="govuk-heading-l">Code snippets</h2>
```
samuelhwilliams, line 8: I've just noticed - I think this should be an h3 now?

samuelhwilliams (approved):
Good work 👍

PR #1198, opened 2021-01-26
fix: Make the visualisation header full width

niross:
### Description of change

Make the visualisation header full screen

Screenshot from 2021-01-26 09-17-32


samuelhwilliams (approved):
Beautiful 👏

PR #1200, opened 2021-01-26
feat: alert on queries running > 5 mins

niross:
### Description of change

Celery task that runs every 5 minutes and will alert if there are any active queries on the datasets db that have been running for more than 5 minutes. Alerts will be sent to the dw sentry channel.

Can be enabled/disabled via the `enable_long_running_query_alerts` waffle flag.

### Checklist


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1446,3 +1446,39 @@ def sync_tool_query_logs():
...
+        response = requests.post(
+            settings.SLACK_SENTRY_CHANNEL_WEBHOOK, json={'text': text},
+        )
+        response.raise_for_status()
+
+
+@celery_app.task()
+def long_running_query_alert():
+    if waffle.switch_is_active('enable_long_running_query_alerts'):
+        interval = '5 minutes'
```
samuelhwilliams, line 1462: I reckon 5 minutes may be a bit aggressive - maybe 15 minutes? And possibly a config change.

samuelhwilliams (approved):

PR #1204, opened 2021-01-27
Feature/dns rewrite proxy nlb

abbas123456:
### Description of change

This creates an NLB to sit in front of the dns-rewrite-proxy ECS service. The NLB has a fixed ip address attached to it in its subnet_mapping meaning that tools should always be able to reach a nameserver. Previously the nameserver was set to the ECS task's ip address meaning that a forced redeployment of the service would result in the tool's dns resolution failing.

This also removes the TCP 53 container port as the ECS loadbalancer block doesn't work with multiple protocols for the same port and it turns out the dns-rewrite-proxy doesn't actually listen on TCP anyway.

Setting the NLB's ip address required an upgrade of the aws provider to version 3.2.0 which required an upgrade to terraform v0.12 in order for it to be compatible. The terraform upgrade required quite a lot of syntax changes and was done as a seperate commit. The accompanying terraform-data-workspace changes can be seen in this PR: https://gitlab.ci.uktrade.digital/data-infrastructure/terraform-data-workspace/-/merge_requests/12

This PR has been tested on catalogue-dev and works. Once both PRs are merged anyone running a terraform command will need to download v0.12 and run a `terraform init`.

https://trello.com/c/icYVH6Hh/821-dns-resolver-ha-with-fixed-private-ip-address-to-avoid-rebooting-breaking-tools
### Checklist

### DIFF
--- a/infra/ecs_main_dns_rewrite_proxy.tf
+++ b/infra/ecs_main_dns_rewrite_proxy.tf
@@ -88,7 +88,7 @@ data "template_file" "dns_rewrite_proxy_container_definitions" {
   template = "${file("${path.module}/ecs_main_dns_rewrite_proxy_container_definitions.json")}"
 
   vars = {
-    container_image    = "${var.dns_rewrite_proxy_container_image}:feature_dns_rewrite_proxy_nlb"
+    container_image    = "${var.dns_rewrite_proxy_container_image}:${data.external.dns_rewrite_proxy_current_tag.result.tag}"
     container_name     = "${local.dns_rewrite_proxy_container_name}"
     container_cpu      = "${local.dns_rewrite_proxy_container_cpu}"
     container_memory   = "${local.dns_rewrite_proxy_container_memory}"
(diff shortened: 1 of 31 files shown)


infra/main.tf
```diff
@@ -1,6 +1,14 @@
...
 data "aws_caller_identity" "aws_caller_identity" {}
 
+provider "aws" {}
+provider "aws" {
+  alias = "route53"
+}
+provider "aws" {
+  alias = "mirror"
+}
+
```
abbas123456, line 11: See https://github.com/hashicorp/terraform/issues/21472#issuecomment-497508239

samuelhwilliams (approved):

samuelhwilliams:
Great job 👍

infra/ecs_main_dns_rewrite_proxy.tf
```diff
@@ -1,13 +1,59 @@
+resource "aws_lb" "dns_rewrite_proxy" {
+  name                             = "${var.prefix}-dns-rewrite-proxy"
+  load_balancer_type               = "network"
+  enable_cross_zone_load_balancing = "true"
+
+  internal = true
+  
+  subnet_mapping {
+    subnet_id            = "${aws_subnet.private_with_egress.*.id[0]}"
+    private_ipv4_address = cidrhost("${aws_subnet.private_with_egress.*.cidr_block[0]}", 5)
```
abbas123456, line 10: This gets the 6th ip address in the CIDR range, the first 4 are reserved

PR #1209, opened 2021-02-01
Feature/your files create table mvp

niross:
### Description of change

- Move your files out of core and into it's own app
- Show the "create table" link if file has a .csv extension
- Fires off a post request via js with the s3 file path
- Validates that:
  1. The file has a .csv extension
  2. The path to the file in s3 is the same as the users s3 prefix
  3. The file exists in s3
- If validation fails a generic error message is shown to the user
- If validation passes:
  1. Generate a table name from the file name
  2. Determine the field types to be used for the table
  3. Copy the file from the user's s3 dir to the data workspace uploads
s3 dir
  4. Call the data-flow api to trigger the dag (not yet implemented on
the data flow side)
  5. Show a generic success message to the user

### Checklist


dataworkspace/dataworkspace/apps/your_files/utils.py
```diff
@@ -0,0 +1,101 @@
...
+    'text': PostgresDataTypes.TEXT.value,
+}
+
+
+def get_s3_csv_column_types(path):
+    client = boto3.client('s3')
+    file = client.get_object(Bucket=settings.NOTEBOOKS_BUCKET, Key=path)
+
+    csv_data = ''
+    for count, line in enumerate(file['Body'].iter_lines()):
```
michalc, line 31: Just erring on the side of paranoia, what would happen if the first line is, say, 5gb in size? Can we protect it from taking out the server?

michalc: (And I guess I don't expect maliciousness, but perhaps people renaming some other large files to see if they can get them ingested...)

michalc: (But also...if this is tricky, happy for a ticket to be created and done later)

dataworkspace/dataworkspace/apps/your_files/utils.py
```diff
@@ -0,0 +1,101 @@
...
+    return re.sub(r'[-\s]+', '_', file_name)
+
+
+def copy_file_to_uploads_bucket(from_path, to_path):
+    client = boto3.client('s3')
+    client.copy(
+        {'Bucket': settings.NOTEBOOKS_BUCKET, 'Key': from_path},
+        settings.AWS_UPLOADS_BUCKET,
+        to_path,
+    )
```
michalc, line 101: (If too tricky, happy for not right now + a ticket created), but can copy_object be used? I _think_ copy passes all the bytes through code, but `copy_object` does it all internally in AWS.

michalc: (Although... I think `copy` does a managed transfer, firing up threads, doing lots of fancy things... I've been slightly burned before by memory leaks in the boto3 client in similar situations, I would probably lean to using `copy_object` form the get go if possible...)

dataworkspace/dataworkspace/static/explorer.js
```diff
@@ -146,6 +146,60 @@ angular.module('aws-js-s3-explorer').controller('ViewController', (Config, s3, $
...
+                if (cookie.substring(0, name.length + 1) === (name + '=')) {
+                    cookieValue = decodeURIComponent(cookie.substring(name.length + 1));
+                    break;
+                }
+            }
+        }
+        return cookieValue;
+    }
+
+    $scope.createTable = function() {
```
michalc, line 165: I would be tempted for the button to probably not be done via Ajax from the get-go... for one thing, in a few tickets time, it would be converted to a regular `<a>` link to another page. I would be more tempted for it to be a full page POST. (And could just be a button styled as a link? Or even, just a regular button for now...)

(But also, as my other comments, can be done later/not at all, etc)

michalc (approved):

dataworkspace/dataworkspace/static/explorer.js
niross: Yeah I was wavering on this myself. My initial thought was that we'd then need either a form for each csv file on the page or some javascript hackery... which seemed weird. But yes I agree this way is not exactly ideal eaither so i'll change it to be form based.

dataworkspace/dataworkspace/apps/your_files/utils.py
niross: Yep makes sense. I might do as suggested though and create a separate ticket for this just to unblock the work waiting behind this PR.

PR #1213, opened 2021-02-02
fix: add header and new column names

niross:
### Description of change

When triggering the data flow pipeline pass:

- The header name from the csv file
- The column for the new table
- The data type for the column

Also fixes naming of schema/table passed to data flow


dataworkspace/dataworkspace/apps/your_files/utils.py
```diff
@@ -37,10 +37,16 @@ def get_s3_csv_column_types(path):
     schema = Schema()
     schema.infer(list(reader), confidence=1, headers=1)
 
-    field_map = {}
+    field_map = []
```
samuelhwilliams, line 40: Maybe rename this `fields` if it's no longer a dict/map?

samuelhwilliams (approved):

PR #1222, opened 2021-02-04
Fix/dag trigger fixes

niross:
### Description of change

- Pass a unique dag run id to data flow so we can check the status of the dag run
- Pass the role name to data flow so we can (optionally) switch to it to create tables.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/your_files/views.py
+++ b/dataworkspace/dataworkspace/apps/your_files/views.py
@@ -1,3 +1,5 @@
+import time
+
 from csp.decorators import csp_update
 from django.conf import settings
 from django.contrib import messages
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/apps/your_files/views.py
```diff
@@ -79,8 +81,11 @@ def form_valid(self, form):
         column_definitions = get_s3_csv_column_types(path)
         import_path = settings.DATAFLOW_IMPORTS_BUCKET_ROOT + '/' + path
         copy_file_to_uploads_bucket(path, import_path)
+        dag_run_id = f'{schema}-{table_name}-{int(time.time())}'
```
samuelhwilliams, line 84: Does this end up being the DAG ID that is displayed in the data-flow UI? Would it be friendlier to have an iso-8601 timestamp?

samuelhwilliams (approved):

dataworkspace/dataworkspace/apps/your_files/views.py
niross: I've been thinking that, if we don't want users to overwrite tables, then actually removing the timestamp altogether is the better alternative. That way there is no risk of overwriting as the request will be rejected via the data flow api... What do you reckon?

niross: Nevermind, seen the discussion about designs

PR #1225, opened 2021-02-05
fix: open data hub link in the same browser window

niross:
### Checklist


samuelhwilliams:
🎉 

But maybe we do want to open it in a new tab?

samuelhwilliams (approved):

PR #1227, opened 2021-02-08
feat: create table validate -> ingest -> success

niross:
### Description of change

- Adds new templates for each step - validate, ingest, success, fail
- Validate waits for dag status to be "running" or "success" and redirects to ingest when it gets a match
- Ingest waits for dag status to be "success" and redirects to success when it gets a match
- If either step gets an http error or a dag status of "failed" it will redirect to the failure template.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/templates/your_files/create-table-base.html
+++ b/dataworkspace/dataworkspace/templates/your_files/create-table-base.html
@@ -1,9 +1,5 @@
 {% extends '_main.html' %}
 {% load static %}
-{% block head %}
-  {{ block.super }}
-  <script nonce="{{ request.csp_nonce }}" src="{% static 'your-files-create-table.js' %}"></script>
-{% endblock %}
 {% block breadcrumbs %}
   <div class="govuk-breadcrumbs">
     <ol class="govuk-breadcrumbs__list">
@@ -17,3 +13,7 @@
     </ol>
   </div>
 {% endblock breadcrumbs %}
+{% block footer_scripts %}
+  {{ block.super }}
+  <script nonce="{{ request.csp_nonce }}" src="{% static 'your-files-create-table.js' %}"></script>
+{% endblock footer_scripts %}
--- a/dataworkspace/dataworkspace/templates/your_files/create-table-base.html
+++ b/dataworkspace/dataworkspace/templates/your_files/create-table-base.html
@@ -1,0 +1,19 @@
+{% extends '_main.html' %}
+{% load static %}
+{% block head %}
+  {{ block.super }}
+  <script nonce="{{ request.csp_nonce }}" src="{% static 'your-files-create-table.js' %}"></script>
+{% endblock %}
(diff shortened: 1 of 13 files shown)


dataworkspace/dataworkspace/templates/your_files/create-table-base.html
```diff
@@ -0,0 +1,19 @@
+{% extends '_main.html' %}
+{% load static %}
+{% block head %}
+  {{ block.super }}
+  <script nonce="{{ request.csp_nonce }}" src="{% static 'your-files-create-table.js' %}"></script>
+{% endblock %}
```
samuelhwilliams, line 6: I think we generally want to include JS files in the footer - I think there's a footer_scripts block that's good for this.

dataworkspace/dataworkspace/templates/your_files/create-table-success.html
```diff
@@ -0,0 +1,24 @@
...
+      <div class="govuk-panel govuk-panel--confirmation">
+        <h1 id="pipeline_header" class="govuk-panel__title">Table created</h1>
+      </div>
+    </div>
+  </div>
+  <div class="govuk-grid-row" id="upload_complete">
+    <div class="govuk-grid-column-full">
+      <h2 class="govuk-heading-m">What next?</h2>
+      <p class="govuk-body">You can immediately access your data in tools at <strong>"{{ schema }}"."{{ table_name }}"</strong>.</p>
+      <a class="govuk-button govuk-button--primary" target="_blank" href="{% url 'explorer:index' %}?sql=SELECT+%2A+FROM+%22{{ schema }}%22.%22{{ table_name }}%22+LIMIT+50">
```
samuelhwilliams, line 16: Can we centralise this bit of SQL generation somewhere so that it updates both here and on the master dataset page if we ever want to update it? Probably feels like we'd want to do both at the same time, and probably without a central place it would be easy to forget either this or the other.

dataworkspace/dataworkspace/templates/your_files/create-table-failed.html
```diff
@@ -0,0 +1,24 @@
+{% extends 'your_files/create-table-base.html' %}
+{% block page_title %}Table creation failed{% endblock page_title %}
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">
+        We can't process your CSV file
```
samuelhwilliams, line 7:
```suggestion
        We can’t process your CSV file
```

Super tiny nit, but GDS designers have sort of etched the use of "fancy apostrophes" into my brain... so might be nice to update. But optional.

dataworkspace/dataworkspace/templates/your_files/create-table-success.html
niross: Done :+1:

samuelhwilliams (approved):

PR #1234, opened 2021-02-11
Dataset column types

samuelhwilliams:
### Description of change
We want to surface information about the colums in master dataset tables
to users as we think this will give them more confidence about what the
data is. We show up to 12 columns on the master dataset page. After
that, if it has more than 12 columns, we direct them to a dedicated page
for the source table where they can see all of the column info.

Also includes a refactor to the `get_context` method of DatasetDetailView, which had become a fairly tangled mess... Probably deserves some close review.

### Checklist


abbas123456 (approved):

abbas123456:
Good job

PR #1236, opened 2021-02-12
feat: allow users to choose their own table name

niross:
### Description of change

- If the table name is invalid, allow it to be corrected
- If the table already exists in their schema, confirm overwrite
- If the table does not exist, run the DAG

closes: https://trello.com/c/spFdGRVN/1391-csv-upload-choose-table-name-table-already-exists-page-4

### Checklist

### Confirm
Screenshot from 2021-02-12 10-05-11

### Confirm table name
Screenshot from 2021-02-12 10-05-19

### Invalid table name
Screenshot from 2021-02-12 10-28-49

### Table exists
Screenshot from 2021-02-12 10-29-02

### DIFF
--- a/dataworkspace/dataworkspace/templates/your_files/create-table-table-exists.html
+++ b/dataworkspace/dataworkspace/templates/your_files/create-table-table-exists.html
@@ -1,0 +1,41 @@
+{% extends '_main.html' %}
+
+{% block page_title %}Table already exists{% endblock page_title %}
+
+{% block go_back %}
+  <a href="{% url 'your_files:create-table-confirm-name' %}?path={{ request.GET.path }}&table_name={{ form.table_name.value }}" class="govuk-back-link">Back</a>
+{% endblock %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <form method="POST" action="{% url 'your_files:create-table-confirm-name' %}?{{ request.GET.urlencode }}" novalidate>
+        <div class="govuk-form-group{% if form.table_name.errors %} govuk-form-group--error{% endif %}">
+          <h1 class="govuk-label-wrapper">
+            <label class="govuk-label govuk-label--xl" for="table">
+              Table already exists
+            </label>
+          </h1>
+          <div class="table-hint">
+            <p class="govuk-body">
+              A table already exists in your personal database schema with the name "{{ form.table_name.value }}". Do you want to overwrite the existing table?
+            </p>
+          </div>
+        <div class="govuk-warning-text">
+          <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+          <strong class="govuk-warning-text__text">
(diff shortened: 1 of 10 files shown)


dataworkspace/dataworkspace/templates/your_files/create-table-table-exists.html
```diff
@@ -7,19 +7,18 @@
 {% endblock %}
 
 {% block content %}
+  {{ form.errors }}
```
samuelhwilliams, line 10: I think this should still use the error summary partial?

samuelhwilliams: Ah, this page "shouldn't" really ever throw validation errors right? 🤔  So probably fine.

dataworkspace/dataworkspace/templates/your_files/create-table-table-exists.html
```diff
@@ -0,0 +1,43 @@
...
+          </strong>
+        </div>
+          {% csrf_token %}
+          {{ form.table_name }}
+          {{ form.path }}
+          {{ form.schema }}
+          {{ form.force_overwrite }}
+        </div>
+        <button type="submit" class="govuk-button">Yes, I understand - Continue</button>
+        <a href="{% url 'your_files:create-table-confirm-name' %}?path={{ request.GET.path }}&table_name={{ form.table_name.value }}" class="govuk-button govuk-button--secondary">
```
samuelhwilliams, line 37: Does this also need to send the schema to populate the back form correctly?

dataworkspace/dataworkspace/templates/your_files/create-table-table-exists.html
```diff
@@ -0,0 +1,43 @@
...
+      <form method="POST" action="{% url 'your_files:create-table-confirm-name' %}?{{ request.GET.urlencode }}" novalidate>
+        <div class="govuk-form-group{% if form.table_name.errors %} govuk-form-group--error{% endif %}">
+          <h1 class="govuk-label-wrapper">
+            <label class="govuk-label govuk-label--xl">
+              Table already exists
+            </label>
+          </h1>
+          <div class="govuk-hint">
+            A table already exists in your personal database schema with the name "{{ form.table_name.value }}". Do you want to overwrite the existing table?
+          </div>
```
samuelhwilliams, line 22: I can't quite put my finger on it, but this HTML feels a bit off to me. I don't think we need the `label` element, particularly as we're not pointing the label at another element.

I wonder if maybe we just need to have `aria-describedby=<some_id>` on the "Continue" button, where <some_id> is the id of the div that describes what happens ("A table already exists ... do you want to overwrite?"). That feels like it provides more context on the action.

samuelhwilliams (approved):

dataworkspace/dataworkspace/templates/your_files/create-table-table-exists.html
niross: Indeed, good catch :+1:

PR #1238, opened 2021-02-12
feature: delete unreferenced docker images on the gitlab-runner ec2

abbas123456:
The gitlab-runner ec2 instance runs out of disk space when there are too many docker images and this causes visualisation builds to fail. This PR adds a crontab entry to run a docker image prune once a day to delete images more than 24 hours old.

https://trello.com/c/SQRttjsg/605-reduce-chance-of-visualisation-build-failures-do-something-about-running-out-of-space-in-the-gitlab-runner

This has been tested on dev.

### DIFF
--- a/infra/ecs_main_gitlab.tf
+++ b/infra/ecs_main_gitlab.tf
@@ -569,6 +569,8 @@ resource "aws_launch_configuration" "gitlab_runner" {
   echo "concurrent = 10" >> /etc/gitlab-runner/config.toml
   echo "check_interval = 1" >> /etc/gitlab-runner/config.toml
 
+  echo "0 0 * * * /usr/bin/docker image prune -f -a --filter until=168h" >> /var/spool/cron/ec2-user
+
   gitlab-runner install --user=gitlab-runner --working-directory=/home/gitlab-runner
   gitlab-runner start
   # Connects via HTTP, but uses private ip, not public internet


infra/ecs_main_gitlab.tf
```diff
@@ -569,6 +569,8 @@ resource "aws_launch_configuration" "gitlab_runner" {
   echo "concurrent = 10" >> /etc/gitlab-runner/config.toml
   echo "check_interval = 1" >> /etc/gitlab-runner/config.toml
 
+  echo "0 0 * * * /usr/bin/docker image prune -f -a --filter until=24h" >> /var/spool/cron/ec2-user
```
samuelhwilliams, line 572: Do we know how often it runs out of disk space? I'm just wondering if this will significantly reduce the amortised speed of visualisation builds if it's having to build without any cached images quite regularly? Could we do 7d, 14d, 28d and it be reasonably safe, and then maybe retain more performance for users?

Maybe caching isn't super relevant here though, in which case 👍

samuelhwilliams (approved):

infra/ecs_main_gitlab.tf
abbas123456: Yeah that's a fair point, maybe 1 week is good middle ground

PR #1239, opened 2021-02-12
fix: use error summary for validation errors

samuelhwilliams:
### Description of change
Disable HTML5 validation as per GOV.UK Design System recommendation and
include an error summary at the top of the page linking to the form
field being validated.

We have to do some slight hacks to fill the dynamically-generated label
and hint text for this page by reaching into the widget's custom context
and manually overwriting that as well. This is some small tech debt from
the Design System form/fields implementation to work with Django's form
system.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/forms.py
+++ b/dataworkspace/dataworkspace/apps/datasets/forms.py
@@ -5,6 +5,13 @@ from django import forms
 
 from dataworkspace.apps.datasets.constants import DataSetType
 from .models import Tag
+from ...forms import (
+    GOVUKDesignSystemForm,
+    GOVUKDesignSystemCharField,
+    GOVUKDesignSystemTextWidget,
+    GOVUKDesignSystemTextareaField,
+    GOVUKDesignSystemTextareaWidget,
+)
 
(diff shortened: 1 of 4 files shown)


dataworkspace/dataworkspace/apps/datasets/forms.py
```diff
@@ -54,9 +61,45 @@ def get_context(self, name, value, attrs):
         return context
 
 
-class RequestAccessForm(forms.Form):
-    email = forms.CharField(widget=forms.TextInput, required=True)
-    goal = forms.CharField(widget=forms.Textarea, required=True)
+class RequestAccessForm(GOVUKDesignSystemForm):
+    email = GOVUKDesignSystemCharField(
+        label="Contact email",
+        help_text='foobar',
```
abbas123456, line 67: foobar?

abbas123456 (approved):

dataworkspace/dataworkspace/apps/datasets/forms.py
samuelhwilliams: 💻

PR #1247, opened 2021-02-16
fix: use responsive dashboard height

samuelhwilliams:
### Description of change
Enable QuickSight's dashboard to automatically set the height of the 
visualisation based on the content, rather than fitting the window and 
scrolling within the iframe.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/templates/quicksight_running.html
+++ b/dataworkspace/dataworkspace/templates/quicksight_running.html
@@ -5,11 +5,19 @@
 {% endblock extra_head %}
 {% block javascript %}
   <script nonce="{{ request.csp_nonce }}">
-    QuickSightEmbedding.embedDashboard({
+    dashboard = QuickSightEmbedding.embedDashboard({
         url: '{{ visualisation_src|safe }}',
         container: document.getElementById('visualisation'),
         locale: "en-gb",
         printEnabled: true,
+        height: "AutoFit",
+        loadingHeight: "1000px"
+    });
+
+    dashboard.on("SHOW_MODAL_EVENT", () => {
+        window.scrollTo({
+            top: 0 // iFrame top position
+        });
     });
   </script>
 {% endblock javascript %}


dataworkspace/dataworkspace/templates/quicksight_running.html
```diff
@@ -5,11 +5,19 @@
...
         container: document.getElementById('visualisation'),
         locale: "en-gb",
         printEnabled: true,
+        height: "AutoFit",
+        loadingHeight: "1000px"
+    });
+
+    dashboard.on("SHOW_MODAL_EVENT", () => {
+        window.scrollTo({
+            top: 0 // iFrame top position
```
michalc, line 19: To check, when the page is really zoomed in, does this behave as expected? Wondering if the modals are near the top of the iframe content, which might not be at the top of the outer page

michalc (approved):

dataworkspace/dataworkspace/templates/quicksight_running.html
samuelhwilliams: This is the method recommended specifically by the AWS embedding SDK [here](https://github.com/awslabs/amazon-quicksight-embedding-sdk) to ensure the modal is in-view. After a quick search I'm not particularly sure how to even generate a modal in a quicksight visualisation, so I haven't actually tested it...

michalc: It could be the thing that pops up when you click the print button, e.g. in the top right of https://data.trade.gov.uk/visualisations/link/73bcb95a-7809-4aae-958d-da463e6fd4cb ?

samuelhwilliams: Ah interesting... The print button actually isn't visible at 400% magnification. On my laptop it disappears above about 250%.

Also with autofit set, the print button ends up off screen if you scroll down, so you can't even really click it unless you're at the top of the screen 🤔

PR #1248, opened 2021-02-16
feature: spread scheduled tasks out as there is a limit of 5 per rule

abbas123456:
### Description of change

There is currenly one rule that runs at 4am with 5 scheduled tasks attached to it but there should be 6. There is a hard limit of 5 tasks per rule and therefore running `terraform apply` results in the following error:

`Error: Updating CloudWatch Event Target failed: LimitExceededException: The requested resource exceeds the maximum number allowed.`

This PR creates an additional rule and places the 6th scheduled task there to get around this.

### Checklist

### DIFF
--- a/infra/ecs_main_mirrors_sync.tf
+++ b/infra/ecs_main_mirrors_sync.tf
@@ -15,7 +15,7 @@ data "template_file" "mirrors_sync_container_definitions_conda" {
   template = "${file("${path.module}/ecs_main_mirrors_sync_container_definitions.json")}"
 
   vars = {
-    container_image    = "${aws_ecr_repository.mirrors_sync.repository_url}"
+    container_image    = "${aws_ecr_repository.mirrors_sync.repository_url}:latest"
     container_name     = "${local.mirrors_sync_container_name}"
     container_cpu      = "${local.mirrors_sync_container_cpu}"
     container_memory   = "${local.mirrors_sync_container_memory}"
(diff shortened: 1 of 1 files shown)


infra/ecs_main_mirrors_sync.tf
```diff
@@ -367,6 +367,13 @@ resource "aws_cloudwatch_event_rule" "daily_at_four_am" {
   schedule_expression = "cron(0 4 * * ? *)"
 }
 
+resource "aws_cloudwatch_event_rule" "daily_at_five_am" {
+  count = "${var.mirrors_bucket_name != "" ? 1 : 0}"
+  name                = "daily-five-am"
+  description         = "daily-five-am"
+  schedule_expression = "cron(0 5 * * ? *)"
+}
+
```
samuelhwilliams, line 376: I guess it doesn't make much difference, but maybe we could just do a count of 2 on the four am rule, and then point `mirrors_sync_debian_task` at `daily_at_four_am[1]`?

Or at least add a comment explaining that this is just because of the limit of targets per rule rather than because something about the debian task means it needs to happen at 5am, which might be implied by the current setup.

samuelhwilliams (approved):
Looks OK, but one comment about the implications of the new rule time.

infra/ecs_main_mirrors_sync.tf
abbas123456: Yeah that's a fair point, will change to have two 4am rules

PR #1256, opened 2021-02-18
fix: do not show deleted visualisations in search

samuelhwilliams:
### Description of change
We were missing a `.live()` call on the VisualisationCatalogueItem
queryset on the dataset search page, so deleted items could still show
up in results. This fixes that and brings it in line with the other
three kinds of dataset.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/tests/datasets/test_views.py
+++ b/dataworkspace/dataworkspace/tests/datasets/test_views.py
@@ -275,6 +275,30 @@ def test_find_datasets_combines_results(client):
     )
 
 
+def test_find_datasets_does_not_show_deleted_entries(client, staff_client):
+    factories.DataSetFactory.create(
+        deleted=True, published=True, name='Unpublished search dataset'
+    )
+    factories.DataSetFactory.create(
+        deleted=True, published=True, name='A search dataset'
+    )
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/tests/datasets/test_views.py
```diff
@@ -275,6 +275,30 @@ def test_find_datasets_combines_results(client):
     )
 
 
+def test_find_datasets_does_not_show_deleted_entries(client, staff_client):
+    factories.DataSetFactory.create(
+        deleted=True, published=True, name='Unpublished search dataset'
```
niross, line 280: Should this be `published=False` or is it just named wrong?

niross (approved):
1 tiny question :+1:

dataworkspace/dataworkspace/tests/datasets/test_views.py
samuelhwilliams: Ah they're published but also deleted - this is actually the case the bug was reported on. So I meant for it to be both, but the test is only named after 'deleted' because that's the sort of more-important case. They could be unpublished and still show up in results (but you need to check the "show unpublished" box).

PR #1263, opened 2021-02-19
feat: stop using `slim` buster variant in tools

samuelhwilliams:
### Description of change
We're currently using `buster-slim` as the base image for our
docker-based tools (e.g. theia, jupyterlab). After a chat with the devs
we don't really have a good reason for this. Our images are already
pretty gargantuan - around 4GB, so the difference between `buster` and
`buster-slim` is insignificant in comparison.

The `-slim` variant specifically excludes man pages from the build, so
in some tools we've had to hack this exclusion to re-introduce them, and
after doing that we're still missing some manpages for common tools,
e.g. grep. So since size isn't a concern, let's go back to using the
"full-fat" `buster` base image. Then we can remove our hacks and benefit
from a more complete man-db with no extra work.

### Checklist

### DIFF
--- a/rstudio/Dockerfile
+++ b/rstudio/Dockerfile
@@ -1,10 +1,4 @@
-FROM debian:buster-20200607-slim
-
-# Allow man pages to be installed: slim has hacks to exclude them, so we
-# hack the hacks https://unix.stackexchange.com/a/480460/361685
-RUN \
-    sed -i '/path-exclude \/usr\/share\/man/d' /etc/dpkg/dpkg.cfg.d/docker && \
-    sed -i '/path-exclude \/usr\/share\/groff/d' /etc/dpkg/dpkg.cfg.d/docker
+FROM debian:buster
 
 RUN \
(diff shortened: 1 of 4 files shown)


niross (approved):
Nice :+1:

rstudio/Dockerfile
```diff
@@ -1,10 +1,4 @@
-FROM debian:buster-20200607-slim
-
-# Allow man pages to be installed: slim has hacks to exclude them, so we
-# hack the hacks https://unix.stackexchange.com/a/480460/361685
-RUN \
-    sed -i '/path-exclude \/usr\/share\/man/d' /etc/dpkg/dpkg.cfg.d/docker && \
-    sed -i '/path-exclude \/usr\/share\/groff/d' /etc/dpkg/dpkg.cfg.d/docker
+FROM debian:buster
```
michalc, line 1: I guess I do lean to pinning a bit more strongly...

michalc (approved):

PR #1270, opened 2021-02-22
fix: allow tables to be focused and scrolled

samuelhwilliams:
### Description of change
Makes sure that the tables for master/reference dataset previews, and Data Explorer results, can be tabbed into and scrolled using the keyboard. It also adds a GOV.UK Design System-style focus on the tables, with some padding added by default to ensure that the table doesn't jump around slightly when the focus border is added.

### Checklist


niross (approved):
Works/Looks good on firefox

PR #1271, opened 2021-02-23
fix: use old school function as callback

niross:
### Description of change

IE11 doesn't like es6 style functions so revert back to function() for the quicksight "show modal" event callback

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/templates/quicksight_running.html
+++ b/dataworkspace/dataworkspace/templates/quicksight_running.html
@@ -14,7 +14,7 @@
         loadingHeight: "1000px"
     });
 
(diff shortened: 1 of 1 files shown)


michalc (approved):

dataworkspace/dataworkspace/templates/quicksight_running.html
```diff
@@ -14,7 +14,7 @@
         loadingHeight: "1000px"
     });
 
-    dashboard.on("SHOW_MODAL_EVENT", () => {
+    dashboard.on("SHOW_MODAL_EVENT", function() {
```
samuelhwilliams, line 17: 🤦

PR #1276, opened 2021-03-01
feature: add Celery task to update QuickSight visualisations' last updated dates

abbas123456:
### Description of change

This PR adds a scheduled celery task that loops over the QuickSight visualisations and updates their modified_date based on the rules outlined at https://github.com/uktrade/data-workspace/pull/1276/files#diff-d711fa82ab90b61eda5a84e72f8c9b1ae27af5aaa2ff120a9e711ce099c7b213R123-R143

https://trello.com/c/pnh3ZJLz/1528-data-visualisations-last-updated
### Checklist


dataworkspace/dataworkspace/datasets_db.py
```diff
@@ -64,3 +65,23 @@ def get_tables_last_updated_date(database_name: str, tables: Tuple[Tuple[str, st
         if dt is None:
             return None
         return dt.replace(tzinfo=pytz.UTC)
+
+
+def extract_queried_tables_from_sql_query(database_name, query):
```
michalc, line 70: Does this effectively give master access to the database to users writing queries in Quicksight?

abbas123456: Oh yeah that's a good point. Will change to use the QuickSight user's connection to run the queries.

abbas123456: Decided to not attempt to extract the tables at all and use the max of Dashboard.LastPublishedTime, Dashboard.LastUpdatedTime and DataSet.LastUpdatedTime

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -97,3 +113,121 @@ def get_code_snippets(source_table):
...
+
+                            last_updated_dates.append(last_updated_date_candidate)
+
+            logger.info(
+                'Setting last updated date of %s for DashboardId %s',
+                max(last_updated_dates).strftime('%d-%m-%Y %H:%M:%S'),
+                dashboard_id,
+            )
+            VisualisationLink.objects.filter(pk=visualisation_link.pk).update(
+                modified_date=max(last_updated_dates)
```
michalc, line 228: Extremely minor thing, but why the `filter` + `update`, rather than `visualisation_link.save(update_fields=['modified_date'])`?

Also, the modified date: if someone changes the label on the link in Django admin, would it then modify the modified_date? Should it not? (So, wondering if it should be a separate field on the model `data_sources_last_updated` or similar)

dataworkspace/dataworkspace/datasets_db.py
michalc: Should this function be removed from the PR?

abbas123456: It's still being used in https://github.com/uktrade/data-workspace/pull/1276/files#diff-011f7fbf79adb1dfbc32b51bc1feb5939659ac973ff11302a2616a63e8aa324cR604-R606

dataworkspace/dataworkspace/apps/datasets/utils.py
abbas123456: > why the filter + update

As the modified_date field has `auto_now=True`, changing the modified_date on the object and then saving will result in the change being overwritten with the current date and time. Doing a filter + update gets around this problem.

> if someone changes the label on the link in Django admin, would it then modify the modified_date?

Yes it would temporarily until the task runs again, probably not the best user experience so I'm happy to add a seperate field for this.

dataworkspace/dataworkspace/datasets_db.py
michalc: Ah sorry, missed it was a refactor

dataworkspace/dataworkspace/apps/datasets/utils.py
michalc: > Doing a filter + update gets around this problem.

Ah, good to know

> Yes it would temporarily until the task runs again, probably not the best user experience so I'm happy to add a separate field for this.

I think I lean to this, not just for the user experience in this case, but also it feels a bit... safer(?), in terms of possible future features/refinements to keep the dates separate.

niross (approved):

PR #1286, opened 2021-03-11
Feature/tool query logs api endpoint

niross:
### Description of change

Adds an api endpoint for tool query audit logs and associated tables to be consumed by data flow.

### Checklist


dataworkspace/dataworkspace/apps/api_v1/pagination.py
```diff
@@ -0,0 +1,5 @@
+from rest_framework.pagination import CursorPagination
+
+
+class TimestampCursorPagination(CursorPagination):
+    ordering = ('timestamp',)
```
michalc, line 5: Just out of crazy paranoia, do should we have a second field here in the case of identical timestamps? e.g. `id`?

michalc (approved):

PR #1289, opened 2021-03-11
feat: add elasticsearch for datasets finder

samuelhwilliams:
### Checklist

### DIFF
--- a/infra/elasticsearch_datasets.tf
+++ b/infra/elasticsearch_datasets.tf
@@ -1,0 +1,84 @@
+resource "aws_iam_service_linked_role" "datasets_finder" {
+  # This is a shared resource between all envs ... couldn't find a way to have
+  # one per env as it automatically assigns the same name (and custom suffixes aren't
+  # allowed for ES service linked roles).
+  aws_service_name = "es.amazonaws.com"
+}
+
+resource "aws_elasticsearch_domain" "datasets_finder" {
+  domain_name           = var.prefix
+  elasticsearch_version = "7.9"
+
+  cluster_config {
+    dedicated_master_enabled = false
+    instance_type = var.datasets_finder_instance_type
+    instance_count = 1
+  }
+
+  domain_endpoint_options {
+    enforce_https = true
+    tls_security_policy = "Policy-Min-TLS-1-2-2019-07"
+  }
+
+  encrypt_at_rest {
+    enabled = true
+  }
+
+  ebs_options {
+    ebs_enabled = true
+    volume_type = "gp2"
+    volume_size = var.datasets_finder_ebs_size
+  }
+
+  vpc_options {
+    subnet_ids = [aws_subnet.datasets[0].id]
+    security_group_ids = [aws_security_group.datasets.id]
+  }
+
+  snapshot_options {
+    automated_snapshot_start_hour = 23
+  }
+
+  access_policies = <<CONFIG
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "${aws_iam_role.admin_task.arn}"
+      },
+      "Action": "es:*",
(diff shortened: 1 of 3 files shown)


infra/security_groups.tf
```diff
@@ -1501,3 +1501,27 @@ resource "aws_security_group_rule" "datasets_db_egress_all_to_quicksight" {
...
+  from_port   = "443"
+  to_port     = "443"
+  protocol    = "tcp"
+}
+
+resource "aws_security_group_rule" "elasticsearch_ingress_from_paas" {
+  description = "ingress-elasticsearch-https-from-paas-ie-data-flow"
+
+  security_group_id = "${aws_security_group.datasets.id}"
+  cidr_blocks       = ["10.0.0.0/16"]
```
michalc, line 1521: It's unlikely to change in any rush, but should the terraform find `10.0.0.0/16` at runtime from `aws_vpc_peering_connection.datasets_to_paas` for clarity?

infra/elasticsearch_datasets.tf
```diff
@@ -0,0 +1,84 @@
+resource "aws_iam_service_linked_role" "datasets_finder" {
+  # This is a shared resource between all envs ... couldn't find a way to have
+  # one per env as it automatically assigns the same name (and custom suffixes aren't
+  # allowed for ES service linked roles).
+  aws_service_name = "es.amazonaws.com"
+}
+
+resource "aws_elasticsearch_domain" "datasets_finder" {
+  domain_name           = var.prefix
```
michalc, line 9: Can/should this include something like "dataset-finder", just for clarity on the AWS-side?

infra/elasticsearch_datasets.tf
```diff
@@ -0,0 +1,84 @@
...
+  access_policies = <<CONFIG
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "${aws_iam_role.admin_task.arn}"
+      },
+      "Action": "es:*",
```
michalc, line 51: Can this be locked down to be read only?

infra/elasticsearch_datasets.tf
```diff
@@ -0,0 +1,84 @@
...
+      "Resource": "arn:aws:es:${data.aws_region.aws_region.name}:${data.aws_caller_identity.aws_caller_identity.account_id}:domain/${var.prefix}/*"
+    },
+    {
+      "Effect": "Allow",
+      "Principal": {
+        "AWS": "${aws_iam_user.datasets_finder_data_flow.arn}"
+      },
+      "Action": "es:*",
+      "Resource": "arn:aws:es:${data.aws_region.aws_region.name}:${data.aws_caller_identity.aws_caller_identity.account_id}:domain/${var.prefix}/*"
+    }
```
michalc, line 61: Erring on the side of paranoia, can the data_flow one be restricted to the PaaS IP range, and the admin one to the IP range of the VPC its in?

infra/security_groups.tf
samuelhwilliams: I think we'd need to make that a terraform data source to retrieve - so going to use `var.paas_cidr_block` instead.

infra/elasticsearch_datasets.tf
samuelhwilliams: 👍

infra/elasticsearch_datasets.tf
samuelhwilliams: 👍

infra/elasticsearch_datasets.tf
samuelhwilliams: 👍

samuelhwilliams:
@michalc good for re-review .. flaky tests are flaky.

michalc (approved):

infra/elasticsearch_datasets.tf
samuelhwilliams: Actually, looks like AWS doesn't let you apply IP based policies to a VPC endpoint ES instance...

samuelhwilliams: I think probably fine to go without - IP controls are in place on the security group to limit access?

PR #1293, opened 2021-03-12
feature: add user satisfaction survey and expose submissions via api

abbas123456:
### Description of change

https://trello.com/c/1u78r6M9/1597-user-satisfaction-survey-form

Related to https://github.com/uktrade/data-flow/pull/563

Designs can be found at https://trello.com/c/oRtWCzZ1/1595-support-and-feedback-user-flow-design

Screenshot_2021-03-15 Leave feedback

### Checklist


dataworkspace/dataworkspace/templates/design_system/checkbox_option.html
```diff
@@ -1,5 +1,5 @@
 <div class="govuk-checkboxes__item">
-  <input class="govuk-checkboxes__input" name="{{ widget.name }}" type="radio" {% if widget.value != None %} value="{{ widget.value|stringformat:'s' }}"{% endif %}{% include "django/forms/widgets/attrs.html" %}>
+  <input class="govuk-checkboxes__input" name="{{ widget.name }}" type="checkbox" {% if widget.value != None %} value="{{ widget.value|stringformat:'s' }}"{% endif %}{% include "django/forms/widgets/attrs.html" %}>
```
abbas123456, line 2: This feels like it will break something else but I couldn't find any reference to this html file?

samuelhwilliams: I think this is OK ... I can't find any references either. 👍

dataworkspace/dataworkspace/apps/core/forms.py
```diff
@@ -30,3 +40,49 @@ class SupportForm(forms.Form):
...
+    'Trying to access data',
+    'Analyse data',
+    'Use a tool',
+    'Create a data visualisation',
+    'Share data',
+    'Share a data visualisation',
+    'View a data visualisation',
+    'Other',
+    'Don’t know',
+]
```
samuelhwilliams, line 64: What do you think about making these enums? Django models/forms support those quite nicely now, and it feels a bit more semantically correct. Also if we ever need to reference any of the choices in code, something like `HowSatisified.VERY_SATISFIED` feels nicer than `HOW_SATISFIED_CHOICES[0]` or having to hard-code the string again?

dataworkspace/dataworkspace/apps/core/forms.py
```diff
@@ -30,3 +40,49 @@ class SupportForm(forms.Form):
...
+
+
+class UserSatisfactionSurveyForm(GOVUKDesignSystemForm):
+    how_satisfied = GOVUKDesignSystemRadioField(
+        required=True,
+        label='Overall how satisfied are you with the current Data Workspace?',
+        widget=GOVUKDesignSystemRadiosWidget(label_size='m'),
```
samuelhwilliams, line 71: I think you'll also need to set `heading='h2'` - by default it makes them as `h1`s.

dataworkspace/dataworkspace/apps/core/forms.py
samuelhwilliams: And possibly we can set the enums on the model itself, which may translate to a Postgresql Enum for a slightly stronger guarantee from the DB itself.

samuelhwilliams (approved):
Nice 👍

PR #1297, opened 2021-03-15
dataset finder app

sekharpanja:
### Description of change

This PR adds dataset finder that was prototyped during firebreak as an app and suitably refactor. It will be restricted to admins only for now.

### Checklist


dataworkspace/dataworkspace/settings/base.py
```diff
@@ -136,6 +137,9 @@ def aws_fargate_private_ip():
 APPSTREAM_STACK_NAME = os.environ.get('APPSTREAM_STACK_NAME')
 APPSTREAM_FLEET_NAME = os.environ.get('APPSTREAM_FLEET_NAME')
 AWS_ECR_ENDPOINT_URL = os.environ.get('AWS_ECR_ENDPOINT_URL')
+DATA_FINDER_ES_HOST = os.environ.get('DATA_FINDER_ES_HOST')
+DATA_FINDER_ES_LOCAL = os.environ.get('DATA_FINDER_ES_LOCAL', False)
```
samuelhwilliams, line 141: I've opted to use `_INSECURE` rather than `_LOCAL` - think it's a bit more explicit. But otherwise yes, I did roughly the same thing.

dataworkspace/dataworkspace/templates/datasets/index.html
```diff
@@ -51,7 +51,7 @@ <h1 class="govuk-heading-xl">Welcome to Data Workspace</h1>
       <div class="search-field govuk-!-margin-bottom-9">
         <label class="govuk-label search-field__label" for="search">Enter your search term(s)</label>
         <div class="search-field-wrapper">
-          <input type="search" name="q" id="search" title="Search" class="govuk-input search-field__item search-field__input js-class-toggle" value="{{ query }}" aria-controls="search-summary-accessible-hint-wrapper">
+          <input type="search" name="q" id="search" title="Search" class="govuk-input search-field__item search-field__input js-class-toggle" value="{{ query }}" aria-controls="">
```
samuelhwilliams, line 54: Not clear why this line has been edited?

dataworkspace/dataworkspace/apps/finder/elasticsearch.py
```diff
@@ -0,0 +1,71 @@
...
+    count: int
+    name: Optional[str] = None
+
+
+class ElasticsearchClient:
+    def __init__(self, *args, **kwargs):
+        if settings.DATASET_FINDER_ES_INSECURE:
+            self._client = Elasticsearch(
+                hosts=[settings.DATASET_FINDER_ES_HOST],
+                http_auth=('elastic', 'changeme'),
```
samuelhwilliams, line 24: Do we need to have this at all? It looks v. weird having "credentials" hard-coded. Would rather just have no credentials? If not, then let's at least make them config.

sekharpanja: Will try to update the defaults image to have no credentials.

samuelhwilliams (approved):
Some minor suggestions, but then I think we can get this in 👍

samuelhwilliams:
Also, I think we should probably just squash all of these commits into a single one.

PR #1306, opened 2021-03-19
log dataset finder searches and show them in django admin

sekharpanja:
### Description of change
This PR adds ability to log the dataset finder queries and show that log in django admin.

### Checklist


dataworkspace/dataworkspace/apps/finder/models.py
```diff
@@ -0,0 +1,18 @@
+from django.contrib.auth import get_user_model
+from django.db import models
+
+
+class DatasetFinderQueryLog(models.Model):
+    id = models.BigAutoField(primary_key=True)
+    query = models.CharField(max_length=255)
```
samuelhwilliams, line 7: I wonder if we should drop off this `max_length` constraint as we don't enforce that on the search term itself (although it's unlikely people will be searching longer than that, they could, and not sure why we shouldn't be able to store that).

samuelhwilliams (approved):

dataworkspace/dataworkspace/tests/finder/test_views.py
```diff
@@ -72,6 +75,8 @@ def test_find_datasets_with_results(client, mocker, dataset_finder_db):
     assert result.table_matches[0].table == 'data'
     assert result.table_matches[0].count == 1260
 
+    assert DatasetFinderQueryLog.objects.all().count() == 1
```
samuelhwilliams, line 78: Might be nice to make some assertions about the content of the querylog?

PR #1308, opened 2021-03-19
feature: clear data explorer cached credentials if permissions change

abbas123456:
### Description of change

When a user is granted or revoked access to a dataset, their data explorer cached credentials should be cleared to ensure their database permissions are valid.

Also, when a dataset's user_access_type is changed, all unauthorized users should have their credentials cleared as well. This ensures that users who couldn't access a now "public" dataset get access, and users who had access to a now "private" dataset lose access.

https://trello.com/c/gUDVTJ9n/1591-people-really-get-data-explorer-credentials-when-they-are-granted-clear-cached-data-explorer-credentials-on-permissions-change

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/admin.py
+++ b/dataworkspace/dataworkspace/apps/datasets/admin.py
@@ -51,6 +51,9 @@ from dataworkspace.apps.dw_admin.forms import (
 from dataworkspace.apps.eventlog.models import EventLog
 from dataworkspace.apps.eventlog.utils import log_permission_change
 from dataworkspace.apps.explorer.schema import clear_schema_info_cache_for_user
+from dataworkspace.apps.explorer.utils import (
+    remove_data_explorer_user_cached_credentials,
+)
 
 logger = logging.getLogger('app')
 
@@ -292,7 +295,8 @@ class BaseDatasetAdmin(PermissionedDatasetAdmin):
 
         super().save_model(request, obj, form, change)
 
-        changed_user_sso_ids = set()
+        changed_users = set()
+        unauthorized_users = set()
 
         clear_schema_info_cache = False
         for user in authorized_users - current_authorized_users:
@@ -304,7 +308,7 @@ class BaseDatasetAdmin(PermissionedDatasetAdmin):
                 {'for_user_id': user.id},
                 f"Added dataset {obj} permission",
             )
-            changed_user_sso_ids.add(str(user.profile.sso_id))
+            changed_users.add(user)
             clear_schema_info_cache = True
 
         for user in current_authorized_users - authorized_users:
@@ -316,7 +320,7 @@ class BaseDatasetAdmin(PermissionedDatasetAdmin):
                 {'for_user_id': user.id},
                 f"Removed dataset {obj} permission",
             )
-            changed_user_sso_ids.add(str(user.profile.sso_id))
+            changed_users.add(user)
             clear_schema_info_cache = True
 
         if clear_schema_info_cache:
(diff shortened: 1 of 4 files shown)


dataworkspace/dataworkspace/apps/datasets/admin.py
```diff
@@ -331,13 +335,27 @@ def save_model(self, request, obj, form, change):
...
+            # As the dataset's access type has changed, clear cached credentials for all
+            # unauthorised users to ensure they either:
+            #   - lose access if it went from REQUIRES_AUTHENTICATION to REQUIRES_AUTHORIZATION
+            #   - get access if it went from REQUIRES_AUTHORIZATION to REQUIRES_AUTHENTICATION
+            unauthorized_users = set(
+                get_user_model().objects.exclude(datasetuserpermission__dataset=obj)
+            )
+
+        for user in changed_users | unauthorized_users:
+            remove_data_explorer_user_cached_credentials(user)
```
samuelhwilliams, line 347: My fear here is that the admin dataset page is already v slow - and this is going to reset creds for roughly 4000 users - are there any performance concerns here? I'm not sure if we can pare this down at all ... could we perhaps only revoke for users with tool privileges?

michalc: One way could be to have fetching cached credentials as a two-step process:

- Fetch a global `cache_version` key that is an integer
- Then the credentials key for a person would be something like `f"explorer_credentials__{cache_version}__{user.profile.sso_id}"`

And then if a dataset auth mode is changed, do `cache_version += 1`, and everyone's creds are expired with only one call?

abbas123456: > could we perhaps only revoke for users with tool privileges?

That seems reasonable to me

samuelhwilliams: I kinda like Michal's idea ... feel nice to invalidate them all in constant time.

abbas123456: Ok but just to confirm, that approach means we will clear credentials for all DW users each time?

michalc: > Ok but just to confirm, that approach means we will clear credentials for all DW users each time?

Ah: I was thinking only in the case when changing auth mode of the dataset?

(But... if it is too much of a project, just tool users sounds reasonable)

abbas123456: > Ah: I was thinking only in the case when changing auth mode of the dataset?

Sorry yes, I meant each time the user_access_type changes

> But... if it is too much of a project, just tool users sounds reasonable

I think it should be fine

.envs/test.env
```diff
@@ -89,7 +89,7 @@ VISUALISATION_EMBED_DOMAINS__1=https://authorized-embedder.com
 
 EFS_ID=some-id
 
-EXPLORER_CONNECTIONS={"Postgres": "my_database"}
+EXPLORER_CONNECTIONS={"Postgres": "my_database", "Postgres_1": "test_external_db", "Postgres_2": "test_external_db2"}
```
samuelhwilliams, line 92: I thought we were moving towards only handling one dataset DB in data workspace? I think that's why I've avoided doing something like this before.

abbas123456: This caused `explorer/test_views.py` to fail so I have since removed it

dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -121,8 +121,11 @@ class QueryException(Exception):
     pass
 
 
+cache_version = 1
```
samuelhwilliams, line 124: This needs to be persisted/retrieved from Redis, no? Otherwise we'll fall back to v1 cached credentials with every deploy, which feels broken.

samuelhwilliams (approved):

PR #1322, opened 2021-03-26
dataset results fetching from ES

sekharpanja:
### Description of change
This PR adds an additional view for the user to checkout the results from Dataset finder searches. It fetches data direct from Elasticsearch for performance reasons and handles paging directly from ES.

### Checklist


dataworkspace/dataworkspace/apps/finder/elasticsearch.py
```diff
@@ -9,6 +9,7 @@
 
 @dataclass
 class _TableMatchResult:
+    index: str
```
samuelhwilliams, line 12: As discussed, I don't think we need to store this - this is storing the raw underlying index (timestamped). I think we should just stick to building the index alias with `{schema}--{table}`

dataworkspace/dataworkspace/apps/finder/views.py
```diff
@@ -64,38 +64,37 @@ def find_datasets(request):
...
-    condition = f"to_tsvector(concat_ws(',', {tsvector_fragment})) @@ to_tsquery('{phrase_fragment}')"
-
-    query = f"""SELECT *
-FROM "{schema}"."{table}"
-WHERE {condition}"""
-
-    return HttpResponseRedirect(f"{reverse('explorer:index')}?sql={quote(query)}")
+def show_results(request, schema, table):
+    form = DatasetFinderResultsForm(request.GET)
+    search_term = request.GET.get("q")
```
samuelhwilliams, line 69: Doesn't look like we're actually using the form here - should we just not instantiate it?

dataworkspace/dataworkspace/apps/finder/utils.py
```diff
@@ -134,3 +135,32 @@ def get_index_aliases_for_all_published_source_tables():
...
+    def __getitem__(self, item):
+        assert isinstance(item, slice)
+
+        resp = self._client.search(
+            body={"query": {"match_phrase": {"_all": self.phrase}}},
+            index=self.index_alias,
+            ignore_unavailable=True,
+            from_=item.start,
+            size=item.stop - item.start,
+        )
```
samuelhwilliams, line 164: I think we need to be careful here and perhaps de-duplicate this - it feels like it would be easy to have a different query body here and on the in `ElasticsearchClient.search_for_phrase`. Can we just make `search_for_phrase` take `from`/`size` parameters?

samuelhwilliams: Also - I'm not sure that we're going to have consistent results between pages here as we haven't specified any ordering. Can you look into adding a reliable order to the query when we're using it to actually show pages of results?

dataworkspace/dataworkspace/apps/finder/elasticsearch.py
sekharpanja: This was a left over from earlier plans. Yes we don't need this.

PR #1323, opened 2021-03-26
feature: use simple_query_string for finder search queries

abbas123456:
### Description of change

The current "match_phrase" query type means that terms must appear in the same order as they do in the document for a match to occur. Using the "match" query type will get around this but it doesn't seem to support searching for an exact phrase which a user may want to do.

Using "simple_query_string" appears to support both, meaning that searching for terms in any order will still return the correct results, i.e "United Kingdom" and "Kingdom United" will return a similar number of results. However, if a user wants to search for an exact phrase they can still do so by wrapping the phrase in double quotes.

See https://trello.com/c/KRB21anx/1604-dataset-finder-evaluation-of-performance-and-results-and-suggest-improvements for more details

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/finder/elasticsearch.py
+++ b/dataworkspace/dataworkspace/apps/finder/elasticsearch.py
@@ -70,7 +70,13 @@ class ElasticsearchClient:
         for batch in self._batch_indexes(index_aliases):
             resp = self._client.search(
                 body={
-                    "query": {"match_phrase": {"_all": phrase}},
+                    "query": {
+                        "simple_query_string": {
+                            "query": phrase,
+                            "fields": ["_all"],
+                            "flags": "FUZZY|PHRASE",
+                        }
+                    },
                     "aggs": {"indexes": {"terms": {"field": "_index"}}},
                 },
                 index=','.join(batch),


dataworkspace/dataworkspace/apps/finder/elasticsearch.py
```diff
@@ -70,7 +70,13 @@ def search_for_phrase(
         for batch in self._batch_indexes(index_aliases):
             resp = self._client.search(
                 body={
-                    "query": {"match_phrase": {"_all": phrase}},
+                    "query": {
+                        "simple_query_string": {
+                            "query": f"{phrase}~",
```
samuelhwilliams, line 75: This feels slightly weird - I think doing a search for `united kingdom` would end up searching `united kingdom~` - so only fuzziness on the last word? That feels unintended/confusing. Perhaps I've misunderstood?

abbas123456: Yeah this was slightly intentional, with the problem being if a ~ is added to each word in the user's query the number of results returned are much higher than expected even if everything is spelled correctly. For example searching "United Arab Emirates" returns 500 results but "United~ Arab~ Emirates~" returns 2.4k results.

I agree it is confusing though, so I think the two options available are to a) only add a ~ if the query contains one word or b) not use fuzziness at all. What do you think?

samuelhwilliams: I maybe lean towards us not adding it, because of the inconsistency. Maybe we can guide users to including it themselves if they want/need it?

abbas123456: Ok that sounds good, we will need some messaging anyway to guide users to add double quotes to search for phrases. Thanks

samuelhwilliams (approved):

dataworkspace/dataworkspace/apps/finder/elasticsearch.py
```diff
@@ -70,7 +70,13 @@ def search_for_phrase(
         for batch in self._batch_indexes(index_aliases):
             resp = self._client.search(
                 body={
-                    "query": {"match_phrase": {"_all": phrase}},
+                    "query": {
+                        "simple_query_string": {
+                            "query": f"{phrase}",
```
samuelhwilliams, line 75: Don't think this needs to be a format string any more, but 🤷

abbas123456: Haha yes sorry will remove

PR #1337, opened 2021-04-01
feature:generate db creds and inject in headers when proxying superset

abbas123456:
https://trello.com/c/zU1FSXQJ/1707-step-2-superset-use-own-database-credentials

1) Add a new api_v1 endpoint to generate/retrieve credentials for any user accessing the superset service provided they have permissions to access tools

2) Clear / invalidate cached superset credentials in the same places that data explorer credentials are cleared / invalidated

3) Change the proxy to call the new api_v1 endpoint and add the credentials returned to the headers for all requests that are proxied to the superset service

4) Change the superset service to mutate the existing database connection (that is manually added via the UI) with the superset user's temporary credentials

5) Change the superset service to assign an 'Editor' role (that is manually added via the UI) to any users accessing the superset-edit subdomain


dataworkspace/dataworkspace/apps/api_v1/core/views.py
```diff
@@ -13,3 +27,61 @@ class UserSatisfactionSurveyViewSet(viewsets.ModelViewSet):
...
+
+
+credentials_version_key = 'superset_credentials_version'
+
+
+def get_cached_credentials_key(user_profile_sso_id):
+    credentials_version = cache.get(credentials_version_key, None)
+    if not credentials_version:
+        credentials_version = 1
+        cache.set(credentials_version_key, credentials_version)
```
michalc, line 39: Could there be a bit of technical race condition here. I guess crazy unlikely, but just in case, should we use the lower-level redis SETNX to only set if it doesn't exist?

dataworkspace/dataworkspace/apps/api_v1/core/views.py
```diff
@@ -13,3 +27,61 @@ class UserSatisfactionSurveyViewSet(viewsets.ModelViewSet):
...
+def remove_superset_user_cached_credentials(user):
+    cache_key = get_cached_credentials_key(user.profile.sso_id)
+    cache.delete(cache_key)
+
+
+def invalidate_superset_user_cached_credentials():
+    credentials_version = cache.get(credentials_version_key, 0)
+    with cache.lock(
+        'get_superset_credentials_version', blocking_timeout=30, timeout=180,
+    ):
```
michalc, line 85: To avoid the lock, could it use Redis's INCR to make sure the version is incremented?

michalc: (Think I'm quite pro keeping the locks down, since I now suspect we already lock a bit too much...)

dataworkspace/proxy.py
```diff
@@ -200,9 +200,24 @@ def application_headers(downstream_request):
             else ()
         )
 
-    def superset_headers(downstream_request):
-        return (
+    async def superset_headers(downstream_request):
+        host_api_url = admin_root + '/api/v1/core/get-superset-role-credentials'
```
michalc, line 204: I wonder if worth _not_ doing this call if the path starts with `/static/` ?

dataworkspace/dataworkspace/apps/api_v1/core/views.py
abbas123456: Yes agree, this was a copy paste from the explorer invalidate function, will update both to use INCR

dataworkspace/dataworkspace/apps/api_v1/core/views.py
abbas123456: What would the race condition be? As it's being set to 1 rather than incremented, not sure what would go wrong if it was called twice?

superset/superset_config.py
```diff
@@ -76,5 +78,14 @@ class DataWorkspaceSecurityManager(SupersetSecurityManager):
     authremoteuserview = DataWorkspaceRemoteUserView
 
 
+def DB_CONNECTION_MUTATOR(uri, params, username, security_manager, source):
+    uri.host = request.headers['Credentials-Db-Host']
+    uri.username = request.headers['Credentials-Db-User']
+    uri.database = request.headers['Credentials-Db-Name']
+    uri.password = request.headers['Credentials-Db-Password']
+    uri.port = request.headers['Credentials-Db-Port']
```
michalc, line 86: The tests I think suggest that the last part of the header name has an underscore rather than a dash, e.g. `db_port`. Should it use the underscore instead?

(Or... should it just use dashes in the header names? I don't think underscores are common in HTTP headers, so I guess I fear a bit of an annoying issue down the line if something doesn't expect them)

abbas123456: Yes the credential headers are sent with underscores but I think Flask replaces the underscores with dashes and upper cases the words. I think to avoid confusion I will change it to send them with dashes.

dataworkspace/dataworkspace/apps/api_v1/core/views.py
michalc: I think very unlikely, but between the `cache.get(credentials_version_key, None)`, `cache.set`, potentially a lot can happen. Someone could have gained access, and then their credentials invalidated, and so the version bumped to 2. Then this request finishes and sets it back to 1.

I think the situation would have to be fairly extreme... but: this is all event loop based, so there is the possibility that something could block the event loop exactly at the wrong time. (And the requests above that cause the issue go to another server without a blocked event loop)

abbas123456: Oh yeah I didn't think of the invalidate getting called in between, will change to use SETNX

michalc (approved):

PR #1359, opened 2021-04-12
Feature/data flow ide

abbas123456:
Adds a new tool for developing Data Flow pipelines, code for which can be found at https://github.com/uktrade/data-flow-sandbox:

Screenshot 2021-04-12 at 09 40 21

**Main changes:**

1) Add the terraform infrastructure for the tool. The ECS task is made up of two containers, the main one being the front end which is a Theia image, and the other being Data Flow itself. The Data Flow image has been stripped down and ships with a sample pipeline that is automatically opened when the tool is launched. Static env vars are stored in the tool's spawner options, and dynamic ones (the database URL to store the datasets/airflow tables and the S3 bucket to store the fetched data) are passed when the tool launches.

2) Set the temporary db user's search_path to the user's schema to allow the airflow tables to be created there. This is suggested in the airflow docs because sqlalchemy doesn't provide a way to specify a schema in the connection URL, see https://airflow.apache.org/docs/apache-airflow/stable/howto/set-up-database.html#setting-up-a-postgresql-database. 
Currently this is only done for the Data Flow IDE tool in case it causes some unwanted side effects elsewhere, but I think it should be safe to do for all tools. 

    The airflow tables are stored in the user's schema for the following reasons:
     - Simplifies the infrastructure, no separate schema or database required for the airflow tables
     - No permission model required to ensure users can only interact with their dags
     - No need for "master" credentials to be passed to the tool via env vars, which can be easily accessed by logging os.environ

3) Add a 'frame-src' CSP header to the proxy to allow the IDE iframe to load subdomain URLs

**Known issues**
- Adding the dags volume to s3sync doesn't work, mobius3 deletes all the files locally for some reason. I had some issues with it picking up docker volumes locally so it may be related. This means any changes to the sample pipeline will be lost when the tool is stopped and launched again
- The spawner options for a tool application template get overwritten after a jenkins / terraform deployment so needs to be manually updated each time
- The environment variables contained in the spawner options are passed to all containers in the task definition, this means S3Sync and Theia receive Data Flow specific environment variables


dataworkspace/dataworkspace/apps/applications/spawner.py
```diff
@@ -55,6 +55,9 @@ def spawn(
 ):
     user = get_user_model().objects.get(pk=user_id)
     application_instance = ApplicationInstance.objects.get(id=application_instance_id)
+    set_search_path = (
+        application_instance.application_template.host_basename == 'data-flow-ide'
+    )
```
abbas123456, line 60: Hardcoded for now, if we decide it isn't safe to do on all tools, I will add a boolean field to the ApplicationTemplate model to control this.

infra/ecs_notebooks_data_flow_container_definitions.json
```diff
@@ -0,0 +1,90 @@
...
+      "options": {
+        "awslogs-group": "${log_group}",
+        "awslogs-region": "${log_region}",
+        "awslogs-stream-prefix": "s3sync"
+      }
+    },
+    "mountPoints": [
+      {
+      "sourceVolume": "home_directory",
+      "containerPath": "/home/s3sync/data/data-flow-ide/theia"
```
abbas123456, line 69: The dags volume is not included here as it results in mobius3 deleting it for some reason. This needs to be investigated further to ensure and changes made to the sample dag are persisted when the tool is stopped and re-launched.

dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -143,6 +144,12 @@ def get_new_credentials(database_obj, tables):
                     sql.Identifier(db_user), sql.Literal(settings.PGAUDIT_LOG_SCOPES)
                 )
             )
+            if set_search_path:
+                cur.execute(
+                    sql.SQL("ALTER USER {} SET search_path = {};").format(
+                        sql.Identifier(db_user), sql.Identifier(db_schema)
+                    )
+                )
```
michalc, line 152: Could the sample pipeline determine the schema name dynamically, and we wouldn't have to set the search path?

dataworkspace/proxy.py
```diff
@@ -135,6 +135,7 @@ def csp_application_running_direct(host, public_host):
             f"connect-src 'self' wss://{host};"
             "font-src 'self' data:;"
             "form-action 'self';"
+            f"frame-src 'self' *.{root_domain};"
```
michalc, line 138: This would allow an application to frame lots of things: other visualisations, superset etc. While potentially there could be a uses cases for those, think better to minimise this a bit. Can it be limited to just the `--9000` version of the current host?

dataworkspace/dataworkspace/apps/core/utils.py
michalc: (The schema name should be in an environment variable. I think `APP_SCHEMA` or similar)

abbas123456: This is required for the airflow tables and is controlled by AIRFLOW__CORE__SQL_ALCHEMY_CONN which expects a SQLAlchemy URL. As there is no way to specify a schema in the connection string this is recommended way to control where the tables get created.

michalc: Ah I got confused: this is for Airflow's metadata tables? I wonder if it's fine/possible to use a SQLite database that is not persisted between launches?

abbas123456: Yes it's for the metadata tables.

I think it might be a bit annoying to have to re-run the pipeline from scratch every time the tool is re-opened, i.e if the run-fetch tasks takes a long time. It also removes the ability for the pipeline to run on a scheduled basis (I know this might not be possible anyway as tasks _should_ get killed after 2 hours of inactivity but at least the option is there)

Also if the pipeline fails half way through and the tool is stopped and re-launched, the user may be left with a temp table that can't be deleted and essentially a pipeline that can't start. Although this can be fixed be marking the first few tasks as succeeded, this requires an advanced understanding of Data Flow which we shouldn't expect tool users to have.

dataworkspace/proxy.py
abbas123456: I tried to pass a pattern like this but it was invalid:

`f"frame-src 'self' *--9000.{root_domain};"`

Any idea how to do this?

PR #1360, opened 2021-04-15
allow columns to be filtered before downloading

sekharpanja:
### Description of change
This PR allows columns to be filtered before downloading. from data cut query preview pages. This change will have to be rolled out separately for source links.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/views.py
+++ b/dataworkspace/dataworkspace/apps/datasets/views.py
@@ -1065,10 +1065,15 @@ class CustomDatasetQueryDownloadView(DetailView):
         dataset.number_of_downloads = F('number_of_downloads') + 1
         dataset.save(update_fields=['number_of_downloads'])
 
+        filtered_query = query.query
+        columns = request.GET.getlist('columns')
+        if columns:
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -1065,10 +1065,15 @@ def get(self, request, *args, **kwargs):
         dataset.number_of_downloads = F('number_of_downloads') + 1
         dataset.save(update_fields=['number_of_downloads'])
 
+        filtered_query = query.query
+        columns = request.GET.getlist('columns')
+        if columns:
+            filtered_query = f"SELECT {','.join(columns)} FROM ({query.query.replace(';', '')}) as data;"
```
niross, line 1071: Might need to be a bit craftier about how we remove the trailing semicolon. As it stands I think this will break some of the more funky string aggs we've got. E.g.

```
STRING_AGG(CONCAT(dit.data_hub__advisers.first_name, ' ', dit.data_hub__advisers.last_name, ' <', dit.data_hub__advisers.contact_email, '>', ' (', dit.data_hub__teams.name, ')'), '; ') AS adviser_details,
```

Joins strings on `;`

I believe we've used `query.query.rstrip().rstrip(';')` elsewhere

niross (approved):

PR #1366, opened 2021-04-19
feature: only add permissions for dashboards the user has access to

abbas123456:
In addition to passing the database credentials to superset via headers, the proxy now passes a list of dashboard ids that the user has access to. This is taken from the identifier values on any SUPERSET catalogue items the user has access to. The supserset config then loops over these dashboards and adds the datasource_access permission for each dashboard slice.

https://trello.com/c/ipyYuNGd/1709-step-4-superset-make-a-visualisation-type-for-superset-in-django

### Checklist


niross (approved):
Nice :+1:

PR #1374, opened 2021-04-23
Bookmarks

sekharpanja:
### Description of change
This PR adds provision for users to bookmark catalogue pages. Search results show if an item is bookmarked or not. And Admin shows number of times an item is bookmarked and allows admins to sort by.

image

image

image

### Checklist


dataworkspace/dataworkspace/apps/core/models.py
```diff
@@ -140,3 +140,10 @@ class UserSatisfactionSurvey(TimeStampedModel):
         null=True, blank=True, choices=TryingToDoType.choices
     )
     improve_service = models.TextField(null=True, blank=True)
+
+
+class BaseUserModel(models.Model):
```
abbas123456, line 145: Where is this being used?

niross (commented):
I can't see `bookmark-no.svg` or `bookmark-yes.svg` being used anywhere. Were they supposed to be included in templates?

niross (commented):
Just had a go locally and I don't think the "bookmarked" icon is working correctly on the reference dataset detail page. It doesn't ever seem to show the dataset as bookmarked when I click it.

niross:
Also i'm unable to view the admin pages for any of the bookmark types

Screenshot from 2021-04-27 13-12-42

I do wonder if we would have been better off using a `GenericForeignKey` rather than multiple models...

dataworkspace/dataworkspace/templates/datasets/index.html
```diff
@@ -88,16 +88,35 @@ <h2 class="govuk-body">
...
+              <strong class="govuk-tag govuk-tag--grey">
+                Unpublished
+              </strong>
+            </div>
+          {% endif %}
+            </div>
+        <div class="govuk-grid-column-one-quarter govuk-!-margin-top-2" style="text-align: right;">
+          {% if dataset.is_bookmarked %}
+          <svg version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink" width="16px" height="18px" xmlns="http://www.w3.org/2000/svg">
+            <g transform="matrix(1 0 0 1 -666 -184 )">
```
niross, line 107: Can we use the included "svg" files used elsewhere here as well? Or are these different images?

sekharpanja:
> Also i'm unable to view the admin pages for any of the bookmark types
> 
> Screenshot from 2021-04-27 13-12-42
> 
> I do wonder if we would have been better off using a `GenericForeignKey` rather than multiple models...

This was mistakenly checked in. Bookmarks are shown along with downloads in list admin page instead of their own admin page.

niross (approved):

PR #1391, opened 2021-05-04
feat: add ag-grid for master datasets

niross:
### Description of change

Adds configurable ag-grid data grid to `SourceTable` objects.

- To enable. check "data grid enabled" for the source table on the master dataset admin page.
- Fields included in the table need to be defined in the "data grid column config" json field.
  - "data grid column config" must contain a list of json objects
  - At a minimum the json object must define the `field` attribute. This is the name of the db column to include in the table
  - Other attributes available are:
    - `headerName`: The header name for the column in the grid
    - `sortable`: Whether this column can be sorted on - [true|false] (optional, default: true)
    - `filter`: Whether this column can be filtered. - [true|false] (optional, default: true)
    - Anything else ag-grid accepts in a column definition

### Checklist


abbas123456 (approved):

abbas123456:
Great job

PR #1406, opened 2021-05-10
fix: query params should be kwargs

niross:
### Description of change

Fixes issue where downloading a query with % in the sql was failing.... Go figure

### Checklist


markhigham (approved):
LGTM

PR #1413, opened 2021-05-13
feat: hide other users slices/datasources/dashboards

niross:
- Adds a new user-level editor role
- For each request ensure users only have access to the datasets they created
- Clean up superset config to fix linter errors

### DIFF
--- a/superset/superset_config.py
+++ b/superset/superset_config.py
@@ -1,18 +1,22 @@
 import os
 
-from superset import db, security_manager
-from superset.security import SupersetSecurityManager
-
 from flask_appbuilder.security.manager import AUTH_REMOTE_USER
 from flask_appbuilder.security.views import AuthView
-from flask_login import login_user
 from flask_appbuilder import expose
+from flask_login import login_user
+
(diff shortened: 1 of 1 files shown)


michalc (approved):

superset/superset_config.py
```diff
@@ -134,19 +138,27 @@ def apply_public_role_permissions(sm, user, role_name):
...
-    uri.password = request.headers['Credentials-Db-Password']
-    uri.port = request.headers['Credentials-Db-Port']
-    return uri, params
+    for datasource in db.session.query(Slice).filter_by(created_by_fk=user.get_id()):
+        permission_view_menu = sm.add_permission_view_menu(
+            'datasource_access', datasource.perm
+        )
+        sm.add_permission_role(role, permission_view_menu)
+
+    user.roles.append(role)
```
michalc, line 160: Just to confirm... this doesn't end up re-appending things so in the DB we can end up with thousands of duplicates?

niross: Yep so `add_permission_role` and `add_permission_view_menu` have no effect if the roles/permissions already exist

PR #1449, opened 2021-05-21
feat: bulk delete files from "your files"

niross:
- makes use of the s3 deleteObjects api call
- behind a feature flag because... paranoia


dataworkspace/dataworkspace/static/explorer.js
```diff
@@ -503,6 +503,155 @@ angular.module('aws-js-s3-explorer').controller('TrashController', (s3, $scope,
...
+        // Build an array of keys
+        let objectKeys = [];
+        for (let i=0; i<objects.length; i++) {
+            objectKeys.push({Key: objects[i].Key});
+        }
+
+        // Attempt to bulk delete all the objects
+        let response;
+        try {
+            response = await s3.deleteObjects({Bucket: model.bucket, Delete: {Objects: objectKeys}}).promise();
```
michalc, line 524: If, somehow, the use has selected > 1000 objects in the UI, would this fail?

michalc: ... but, can `$scope.deleteObjects`  _not_ be on on the `$scope`, since I think it's essentially a "private" function?

dataworkspace/dataworkspace/static/explorer.js
```diff
@@ -503,6 +503,155 @@ angular.module('aws-js-s3-explorer').controller('TrashController', (s3, $scope,
...
+            // Attempt to list objects under the prefix. If the list objects
+            // call fails, update the model and try the next prefix.
+            let response;
+            try {
+                response = await s3.listObjectsV2({
+                    Bucket: model.bucket,
+                    Prefix: prefix.Prefix,
+                    ContinuationToken: continuationToken
+                }).promise()
+                continuationToken = response.NextContinuationToken;
```
michalc, line 606: Just to check on the logic here... is `continuationToken` ever used? Are later pages of keys from this call ever deleted?

dataworkspace/dataworkspace/static/explorer.js
michalc: I wonder also, instead of plain function, it should be something with a but more... state?
```javascript
function Deleter() {
  var keys = []  

  async function deleteNow() {
    // code to bulkDelete all keys in a single call, and raise an exception if there is an error
    // for the calling code to handle as needed
  }

  async function scheduleDelete(key) {
      keys.push(key)
      if (keys.length >= bulkDeleteMaxFiles) {
         await deleteNow();
      }
  }

  async function flushDelete() {
    if (keys.length) await deleteNow();
  }
  return [scheduleDelete, flushDelete];
}
```
so calling code can create a deleter as needed:
```javascript
let [scheduleDelete, flushDelete] = Deleter();
```

dataworkspace/dataworkspace/apps/your_files/views.py
```diff
@@ -50,7 +50,11 @@ def file_browser_html_GET(request):
     return render(
         request,
         'your_files/files.html',
-        {'prefix': prefix, 'bucket': settings.NOTEBOOKS_BUCKET},
+        {
+            'prefix': prefix,
+            'bucket': settings.NOTEBOOKS_BUCKET,
+            'YOUR_FILES_BULK_DELETE_FLAG': settings.YOUR_FILES_BULK_DELETE_FLAG,
```
michalc, line 56: I guess I do lean to not having a flag and just doing it?

dataworkspace/dataworkspace/static/explorer.js
niross: > ... but, can `$scope.deleteObjects` _not_ be on on the `$scope`, since I think it's essentially a "private" function?

Originally I did it that way, can you still access $scope from within the private function?

dataworkspace/dataworkspace/static/explorer.js
niross: It's a direct copy form the original delete method. I will check.

dataworkspace/dataworkspace/static/explorer.js
michalc: Yes, I think so

dataworkspace/dataworkspace/static/explorer.js
niross: Ah yeah i see what you mean. Sorted now.

dataworkspace/dataworkspace/static/explorer.js
```diff
@@ -433,74 +433,104 @@ angular.module('aws-js-s3-explorer').controller('TrashController', (s3, $scope,
...
+            let isTruncated = true;
+            while(isTruncated) {
+                let response;
+                try {
                     response = await s3.listObjectsV2({
                         Bucket: model.bucket,
                         Prefix: prefix.Prefix,
-                        ContinuationToken: continuationToken
+                        ContinuationToken: continuationToken,
+                        MaxKeys: 1
```
michalc, line 468: Left over from dev?

niross: Very much so, thanks!

dataworkspace/dataworkspace/static/explorer.js
```diff
@@ -433,74 +433,103 @@ angular.module('aws-js-s3-explorer').controller('TrashController', (s3, $scope,
...
-                let isTruncated = true;
-                let continuationToken = null;
-                while (isTruncated) {
+            setObjectMessage(prefix, 'deleteStarted', true);
+
+            // Attempt to list objects under the prefix. If the list objects
+            // call fails, update the model and try the next prefix.
+            let continuationToken = null;
+            let isTruncated = true;
+            while(isTruncated) {
```
michalc, line 461: Although it was doing it before I realise, should it also stop if `model.aborted` is true, and so have `&& !model.aborted` clause?

michalc (approved):

michalc:
Looks good. Small comment on the aborted thing, but it was already not stopping on abort, not a biggy I think.

PR #1453, opened 2021-05-27
fix: remove unused npm package files

niross:
### Description of change

Removes npm package files that were left over from the original data-explorer implementation. While npm is used in the dev env, the package is installed directly rather than through a package file.

Removing the file fixes [CVE-2021-23364](https://github.com/advisories/GHSA-w8qv-6jwh-64r5).

Happy to discuss if we think these files are necessary or not.


markhigham (approved):
LGTM

PR #1456, opened 2021-05-28
feature: allow specifying a query and inject this data into the config

abbas123456:
### Description of change

This PR means the vega visualisation data no longer needs to be hardcoded in the config. It is instead read from the database based on the query provided in the admin and injected into the config.

### Checklist


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -1552,8 +1552,21 @@ def get(self, request, dataset_uuid, object_id, **kwargs):
             return HttpResponse(status=404)
 
         visualisation = dataset.visualisations.get(id=object_id)
+
+        with psycopg2.connect(
+            database_dsn(settings.DATABASES_DATA[visualisation.database.memorable_name])
+        ) as connection:
+            with connection.cursor(
+                name='data-set-visualisation-data',
```
michalc, line 1560: Maybe it doesn't need to be a named/server-side cursor, if it fetches all the results into memory anyway?

dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -1552,8 +1552,21 @@ def get(self, request, dataset_uuid, object_id, **kwargs):
...
+                cursor_factory=psycopg2.extras.RealDictCursor,
+            ) as cursor:
+                cursor.execute(visualisation.query)
+                data = cursor.fetchall()
+
         vega_definition = json.loads(visualisation.vega_definition_json)
 
+        vega_definition['data'][1]['values'] = data
```
michalc, line 1568: Why the [1]?

(/ Should attempting to run the query be optional, so it's still fine to have a completely hard coded graph)

michalc (approved):

PR #1458, opened 2021-06-01
feat: more closing of connections after celery tasks

michalc:
### Description of change

As recommended at https://github.com/jneight/django-db-geventpool#using-orm-when-not-serving-requests

### Checklist


niross (approved):
Great catch :raised_hands:

PR #1468, opened 2021-06-03
creating team schemas from django admin

markhigham:
### Description of change
New models for team and membership which are editable from django admin only.
Amended spawner to create postgres schema for the team and granted permissions to the current user's permanent db_role

### Checklist


dataworkspace/dataworkspace/apps/applications/spawner.py
```diff
@@ -78,6 +82,15 @@ def spawn(
         valid_for=datetime.timedelta(days=31),
     )
 
+    logger.info(credentials)
+    teams = get_teams_for_user(user)
+
+    schemas = get_or_create_team_schemas(teams, source_tables, credentials)
```
michalc, line 88: Ah: I think this should be inside the `new_private_database_credentials` function for the cases that need credentials but not from a spawned tool, e.g. Data Explorer.

dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -73,6 +81,49 @@ def db_role_schema_suffix_for_app(application_template):
...
+                )
+
+                created_schemas[team] = schema_name
+                for cred in user_credentials:
+                    role_name = cred["db_persistent_role"]
+                    logger.info(role_name)
+                    cur.execute(
+                        sql.SQL('GRANT USAGE ON SCHEMA {} TO {};').format(
+                            sql.Identifier(schema_name), sql.Identifier(role_name)
+                        )
```
michalc, line 121: I suspect USAGE is not enough for people to create tables in the schema? (And I wonder if the simplest thing to make it possible would be a team role in the database, the team set to the owner of the schema, and the user's persistant role GRANTed the team role)

michalc: Ah...and I realise that if it's the permanent role that's GRANTed, something should revoke access if/when the user leaves the team. If it's the temporary user that's granted, then I think we don't need to revoke them... (although: maybe we do need to revoke in the job that deletes the temporary users...)

I originally thought that https://github.com/uktrade/data-workspace/pull/1467 shouldn't go in before this PR, but I wonder now if it should. It's got logic for granting/revoking things to the permanent db_role, and it only changes code inside `new_private_database_credentials`, and it changes no tests (since its ~a refactor from the point of view of the outside world), so it might not even have any conflicts with this PR...

michalc:
Done in https://github.com/uktrade/data-workspace/pull/1495

PR #1494, opened 2021-06-14
using psycopg2.sql to build custom sql

markhigham:
### Description of change
using psycopg2 string formatting to avoid potential sql injection attacks

### Checklist


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -1187,7 +1186,13 @@ def get(self, request, *args, **kwargs):
         filtered_query = query.query
         columns = request.GET.getlist('columns')
         if columns:
-            filtered_query = f"SELECT {','.join(columns)} FROM ({query.query.rstrip().rstrip(';')}) as data;"
+            trimmed_query = query.query.rstrip().rstrip(';')
+            filtered_query = sql.SQL("SELECT {fields} from ({query}) as data;").format(
+                fields=sql.SQL(',').join(
+                    [sql.Identifier(column) for column in columns]
+                ),
+                query=sql.Identifier(trimmed_query),
```
michalc, line 1194: This looks like it would treat the query as an identifier, so probably wrap it in double quotes. Is this right?

markhigham: it's more complicated than that. I didn't realise that this code was called from the none-datagrid download.

michalc (approved):

PR #1495, opened 2021-06-14
feat: team schemas

michalc:
### Description of change

Creates a Team model where users can be made members. With each team, a single schema is maintained where all members of that team have read and write access. This is via a permanent team role to which each user's permanent role is GRANTed.

The Django Admin side is a bit basic: may make this a bit better in a later PR.

### Checklist


sekharpanja (approved):
Looks good to me

PR #1503, opened 2021-06-16
adding event_logging for data downloads

markhigham:
### Description of change
Adds a row to EventLog of type `TYPE_DATASET_CUSTOM_QUERY_DOWNLOAD_COMPLETE` on completion of a download from the DataGridDataView. (Enhanced Preview)

This event contains metrics about the time of execution, the number of rows downloaded vs total available, and the number of columns downloaded vs available.
A correlation_id value is added to the event and to the related `TYPE_DATASET_CUSTOM_QUERY_DOWNLOAD` event

### Checklist


dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -10,6 +10,8 @@
 import csv
 from typing import Tuple
 
+from timeit import default_timer as timer
```
michalc, line 13: Probably not much in it, but to me, seems a bit odd to use timeit in prod, and would more expect something from just `time`, e.g. `time.monotonic`

michalc (approved):

dataworkspace/dataworkspace/apps/core/utils.py
markhigham: I looked past the first answer on stack overflow and found this.
https://stackoverflow.com/a/25823885/155965
I liked that it "provides the best clock available on your platform and version of Python automatically"

PR #1529, opened 2021-06-29
fix: display uuid[] as text[] in data explorer output

niross:
### Description of change

Add `uuid[]` to column types and map to string in the ui to stop them looking like `UUID(xxxx-xxxx...)` 

### Checklist


markhigham (approved):
LGTM

PR #1536, opened 2021-07-07
Fix/query timeouts

niross:
### Description of change

1. Sets a timeout on "idle in transaction" queries to match the existing query timeout.
2. Gives streaming download cursors more descriptive names.

For custom dataset query downloads the cursor name includes the id and the slugified name of the model.

It will look something like:

```
FETCH FORWARD 1000 FROM "custom-query--1--a-long-running-query"
```

And for data grid queries it will include the model type the query is running against and the id of that model.

It will look something like:

```
FETCH FORWARD 1000 FROM "data-grid--CustomDatasetQuery--1"
```

or 

```
FETCH FORWARD 1000 FROM "data-grid--SourceTable--2"
```

### DIFF
--- a/dataworkspace/dataworkspace/apps/core/utils.py
+++ b/dataworkspace/dataworkspace/apps/core/utils.py
@@ -888,7 +888,10 @@ def streaming_query_response(
 
         with connect(
             database_dsn(settings.DATABASES_DATA[database]),
-            options=f'-c idle_in_transaction_session_timeout={query_timeout}',
+            options=(
+                '-c idle_in_transaction_session_timeout={timeout} '
+                '-c statement_timeout={timeout}'
+            ).format(timeout=query_timeout),
         ) as conn:
             conn.set_session(readonly=True)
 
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/apps/core/utils.py
```diff
@@ -884,7 +886,10 @@ def get_row_count_from_unfiltered(conn):
     def run_queries():
         should_run_query_metrics = unfiltered_query and query_metrics_callback
 
-        with connect(database_dsn(settings.DATABASES_DATA[database])) as conn:
+        with connect(
+            database_dsn(settings.DATABASES_DATA[database]),
+            options=f'-c idle_in_transaction_session_timeout={query_timeout}',
+        ) as conn:
```
michalc, line 892: I wonder, can the statement timeout and transaction timeout be specified in the same way, either both here, or both below?

(Also, I was always a bit skeptical of the pattern with the inner cursor setting things that should be applicable outside of it... I wouldn't be against having those outside of the server side cursor setup)

michalc (approved):

dataworkspace/dataworkspace/apps/core/utils.py
niross: Yeah can do. I wasn't sure you could actually set `idle_in_transaction_session_timeout` directly on the cursor. I couldn't find any info on it anyway. 

> (Also, I was always a bit skeptical of the pattern with the inner cursor setting things that should be applicable outside of it... I wouldn't be against having those outside of the server side cursor setup)
Ok so i'll try to set what I can on the connection itself yes?

michalc: I think yes? But no biggie I think if it seems to be taking too long

PR #1541, opened 2021-07-08
feature: change local superset setup to use proxy

abbas123456:
### Description of change

The previous local set up meant requests to superset were not going through the proxy.

### Checklist


niross:
Much better :+1:

niross (approved):

PR #1543, opened 2021-07-09
fix: use custom json encoder to allow superset query history to work

abbas123456:
### Description of change

/superset/sqllab/history was previously returning 500 as it was trying to serialize a decimal to JSON. This adds a custom JSON encoder to get around this.

### Checklist

### DIFF
--- a/superset/superset_config.py
+++ b/superset/superset_config.py
@@ -1,3 +1,4 @@
+import decimal
 import os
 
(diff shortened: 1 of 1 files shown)


niross (approved):
Looks good, I wonder if we should rename `data_workspace_permission_handler` now though as it is a bit more multi purpose

PR #1548, opened 2021-07-14
feat: intercept map style file requests from superset

michalc:
### Checklist


michalc:
Closing since requests get blocked by the CSP before they get to the service worker

PR #1549, opened 2021-07-15
Avoiding keyboard trap when using data-grid with dynamic background data load

markhigham:
### Description of change
Focus can be trapped in the datagrid when you navigate with keyboard only _and_ data is dynamically loaded. This happens because the grid requests more data when focus lands on the last row. For large data sets it becomes impossible to reach the bottom and tab out of the grid.

This change supresses tab key navigation between grid cells. Now if you press tab while in the grid the focus moves to the next element on the page 

### Checklist


niross (approved):
Nice :+1:

PR #1551, opened 2021-07-16
fix: replace requests to mapbox with our own mirror

niross:
### Description of change

Overrides `fetch()` and `Worker.postMessage()` to inject our map tiles into the mapbox and deck.gl superset charts.

Screenshot from 2021-07-16 16-59-55


michalc (approved):

markhigham (approved):
LGTM

PR #1562, opened 2021-07-26
feat: allow dataset access to be granted automatically based on email

abbas123456:
### Description of change

This adds a field to the Dataset and VisualisationCatalogueItem models to allow a list of authorized email domains to be specified. Any user with an email address domain that is authorized gets granted access to the dataset automatically.

https://trello.com/c/eITmaxWB/2079-data-workspace-catalogue-items-are-made-open-to-users-with-a-specific-email-domain

### Checklist


niross (approved):

dataworkspace/dataworkspace/apps/datasets/admin.py
```diff
@@ -334,7 +336,10 @@ def save_model(self, request, obj, form, change):
             changed_users.add(user)
             clear_schema_info_cache_for_user(user)
 
-        if original_user_access_type != obj.user_access_type:
+        if (
+            original_user_access_type != obj.user_access_type
+            or 'authorized_email_domains' in form.changed_data
+        ):
```
michalc, line 342: The log event looks like the user_access_type has been changed, which not is now always the case if this bit of code runs. Should it be a different EventLog type?

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -2054,6 +2059,9 @@ class VisualisationCatalogueItem(DeletableTimestampedUserModel):
     # Used as a parallel to DataSet.type, which will help other parts of the codebase
     # easily distinguish between reference datasets, datacuts, master datasets and visualisations.
     type = DataSetType.VISUALISATION
+    authorized_email_domains = ArrayField(
+        models.CharField(max_length=256), blank=True, null=True
```
michalc, line 2063: Does this have to be null?

michalc: (Essentially, can it be the empty array if there are no email domains)

PR #1579, opened 2021-08-17
Convert newlines to <br>, show pretty json with double quotes and indent

sekharpanja:
### Description of change
This PR is to convert newlines to br, show indents and format json in query results

image

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/explorer/utils.py
+++ b/dataworkspace/dataworkspace/apps/explorer/utils.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import re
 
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -245,10 +246,20 @@ def fetch_query_results(query_log_id):
...
+        description = [(re.sub(r'col_\d*_', '', s.name),) for s in cursor.description]
         headers = [d[0].strip() for d in description] if description else ['--']
-        data = [list(r) for r in cursor]
-    return headers, data, query_log
+        data_list = [list(r) for r in cursor]
+        types = [pg_type[t.type_code] for t in cursor.description]
+        data = []
+        for i, row in enumerate(data_list[0], start=0):
+            if types[i] == 'jsonb':
+                data.append(json.dumps(row, indent=2, sort_keys=True))
```
michalc, line 262: Can this _not_ sort the keys? Just so the output is semantically the same as what's in the DB.

(I know dictionaries aren't really ordered... but they sort of are...)

dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -245,10 +246,20 @@ def fetch_query_results(query_log_id):
     )
     table_name = tempory_query_table_name(user, query_log.id)
     with user_explorer_connection(user_connection_settings) as conn:
-        cursor = conn.cursor()
+        cursor = conn.cursor(cursor_factory=psycopg2.extras.DictCursor)
+        cursor.execute('select oid, * from pg_type')
```
michalc, line 250: The most tiny thing, but instead of *, can it be just `typname`, since that's all we need?

(But also, we're only looking for the 1 type, jsonb? So could have `SELECT oid FROM pg_type WHERE typename = 'jsonb'`, and not have any dicts or anything below?)

michalc (approved):

dataworkspace/dataworkspace/apps/explorer/utils.py
```diff
@@ -245,10 +246,20 @@ def fetch_query_results(query_log_id):
...
-        data = [list(r) for r in cursor]
-    return headers, data, query_log
+        data_list = [list(r) for r in cursor]
+        types = [pg_type[t.type_code] for t in cursor.description]
+        data = []
+        for i, row in enumerate(data_list[0], start=0):
+            if types[i] == 'jsonb':
+                data.append(json.dumps(row, indent=2))
+            else:
+                data.append(row)
```
michalc, line 264: Ah... One thing we didn't have before... `append`. Can this still be as a list comprehension, for performance/memory reasons?

michalc (approved):

PR #1585, opened 2021-08-23
fix: return all rows for data explorer queries

niross:
### Checklist


markhigham (approved):
👍

PR #1593, opened 2021-08-25
fix: redirect to home from /request-access if user doesnt need access

abbas123456:
### Checklist


abbas123456:
not a great solution, will add some copy to explain that the user already has access

PR #1597, opened 2021-08-31
feat: do not show vega visualisations to ie11 users

niross:
- Pass the user agent header through to the django application from the proxy
- Adds a new template tag to determine if a user agent matches ie11 or below
- Do not show master dataset visualisations if the user is using internet explorer

### DIFF
--- a/dataworkspace/dataworkspace/apps/core/templatetags/core_tags.py
+++ b/dataworkspace/dataworkspace/apps/core/templatetags/core_tags.py
@@ -1,4 +1,5 @@
 import json
+import re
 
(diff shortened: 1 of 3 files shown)


markhigham (commented):
does this includes a separate feature for related visualisations?

niross:
> does this includes a separate feature for related visualisations?

Do we show vega visualisations in related visualisations? I _think_ it's only on the mast dataset page

markhigham (approved):
LGTM. I testing in IE11 too. Works ok

PR #1610, opened 2021-09-06
Esacape python r code snippets

sekharpanja:
### Description of change
This PR is to escape both Python and R code snippets, so that the code can be run without any errors in JupyterLab and RStudio.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/utils.py
+++ b/dataworkspace/dataworkspace/apps/datasets/utils.py
@@ -99,6 +99,7 @@ def get_code_snippets_for_table(source_table):
         "sql": query,
     }
 
+
 def get_code_snippets_for_reference_table(table):
     query = get_sql_snippet('public', table, 50)
     return {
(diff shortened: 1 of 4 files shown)


niross (approved):

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -137,7 +138,7 @@ def get_r_snippet(query):
 }}
 conn <- getConn(Sys.getenv('DATABASE_DSN__datasets_1'))
 
-res <- dbSendQuery(conn, '{query}')
+res <- dbSendQuery(conn, \"{query}\")
```
michalc, line 141: To confirm, do double quotes need to be escaped in triple quoted Python strings? (from a brief test, they didn't seem to)

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -121,19 +122,25 @@ def get_sql_snippet(schema, table_name, limit=50):
 
 
 def get_python_snippet(query):
+    """
+    sqlalchemy.text() is used to make sure the sql string is in a form that sqlalchemy expects.
+    `backslash`, `"` and `'''` are also escaped
+    """
+    query = query.replace('\\', '\\\\').replace('"', '\\"').replace('"""', '\\"""')
```
michalc, line 129: I think once `.replace('"', '\\"')` is done, then `.replace('"""', '\\"""')` can't do anything, so should be safe to remove it?

PR #1630, opened 2021-09-22
Feature/dataset finder data grid

niross:
Adds a grid to the finder results page.

There are some limitations to the use of the grid due to the our indexes are setup currently.

1. Sort is not supported - we will need to enable this at the index level
2. The number of field level filters have been cut as we would need specific anaylsers setup for each field type (extra work on the elasticsearch side)
    * Available filters are:
        * text: contains/not contains
        * date: equal/not equal, less/greater than, range
        * bool: equals
        * uuid: contains/not contains
3. Fixes an issue where the data grid was expecting a "download json" button that doesn't always exist.


abbas123456 (approved):

abbas123456:
Good work

PR #1636, opened 2021-09-24
fix: use table perm checker in finder results

niross:
### Description of change

Check user's access to a dataset via the `user_has_access` method when showing finder results....

1. Fixes a bug where sometimes the user was shown the "request access" link when they already had access
2. The previous way of checking access does not work for allowed email domains 

### Checklist


markhigham (approved):
LGTM

PR #1637, opened 2021-09-24
Fix for data explorer returning zero results

markhigham:
### Description of change
The textbox for query page size defaults to max 1000 or the number of rows returned.
When the number of rows returned is zero, this textbox is set to zero.
This causes problems with subsequent queries which return zero results too.

The fix is to store the page size when the query is requested against the query model and use this value (as opposed max rows or row count) for subsequent page views.

### Checklist


michalc:
@markhigham This does solve the problem with zero results, but I think it's still a problem that it's limited to the number of results? If there is 1 row of results, then it still shouldn't limit to just the 1 row for the next query?

dataworkspace/dataworkspace/apps/explorer/views.py
```diff
@@ -564,14 +579,19 @@ def get(self, request, querylog_id):
             elif query_log.state == QueryLogState.COMPLETE:
                 template = loader.get_template('explorer/partials/query_results.html')
                 headers, data, _ = fetch_query_results(querylog_id)
+                next_page_request_rows = (
+                    len(data) if data else settings.EXPLORER_DEFAULT_ROWS
+                )
```
michalc, line 584: Does this still make the number of results per page dependant on the data? (From the discussion, I thought it shouldn't any more...)

markhigham: Done 👍

michalc (approved):

PR #1643, opened 2021-10-05
feat: Add local s3 support via localstack

niross:
- Allow for hitting s3 locally for your files
- Config update to allow your files table creation via local data-flow
- Adds missing functions to the datasets db to allow for data-flow integration
- Adds profiles to the def docker compose config. Allowing us to run specific parts of the stack locally, e.g.
  - dataset finder `docker-compose -f docker-compose-dev.yml --profile finder up`
  - superset `docker-compose -f docker-compose-dev.yml --profile superset up`

** you will probably need to update docker-compose to get profiles working


niross:
Partially done as part of the react your files migration

PR #1644, opened 2021-10-05
fix: prevent users from seeing other user's access request details

abbas123456:
### Description of change

This stops users from accessing other user's access request data by changing the id in the url

### Checklist


markhigham (approved):
LGTM

PR #1690, opened 2021-11-10
feat: add admin only page for source table changelog

niross:
### Description of change

Adds a new source table changelog page visible to admins only and currently only showing "structural" changes to the table.

### Link from the source table section of the master dataset catalogue page
Screenshot from 2021-11-10 12-39-50

### The changelog
Screenshot from 2021-11-10 12-40-10

### Checklist


dataworkspace/dataworkspace/templates/datasets/master_dataset.html
```diff
@@ -136,6 +136,9 @@ <h2 class="govuk-heading-l">Data</h2>
           <th class="govuk-table__header">Identifier</th>
           <th class="govuk-table__header">Last updated</th>
           <th class="govuk-table__header">Action</th>
+          {% if user.is_superuser %}
```
michalc, line 139: Ah: do we want this not so visible, say for demos?

niross: So behind a flag? Or not visible at all?

michalc: I did initially think not at all, but maybe yes, a separate feature flag would be good? So we can at least turn it off, including when the whole notification feature goes live?

abbas123456 (approved):

PR #1699, opened 2021-11-16
feat: manage all user notification subscriptions

markhigham:
### Description of change
Creation of a new page to manage notification subscriptions.

### Checklist


dataworkspace/dataworkspace/apps/datasets/subscriptions/utils.py
```diff
@@ -10,3 +17,35 @@ def get_or_create_dataset_subscription_for_user(
...
+
+
+def unsubscribe_from_all(user):
+    logger.info("unsubscribe from all subscriptions for %s", user)
+
+    subscriptions = [
+        s for s in DataSetSubscription.objects.filter(user=user) if s.is_active()
+    ]
+
+    logger.info(len(subscriptions))
```
abbas123456, line 29: Should this be removed?

markhigham: yes 👍

dataworkspace/dataworkspace/apps/datasets/subscriptions/views.py
```diff
@@ -9,9 +10,64 @@
...
+            context={
+                "subscriptions": [s for s in subscriptions if s.is_active()],
+                "question_text": "Are you sure you want to unsubscribe?",
+            },
+        )
+
+    def post(self, request, subscription_id):
+        if subscription_id == "__all__":
+            # Ignore the post values and just delete all subscriptions
+            logger.debug("__all__")
```
abbas123456, line 61: Again, should this be removed

dataworkspace/dataworkspace/apps/datasets/subscriptions/utils.py
```diff
@@ -10,3 +17,35 @@ def get_or_create_dataset_subscription_for_user(
 
     subscription = dataset.datasetsubscription_set.create(user=user)
     return subscription
+
+
+def unsubscribe_from_all(user):
+    logger.info("unsubscribe from all subscriptions for %s", user)
+
+    subscriptions = [
+        s for s in DataSetSubscription.objects.filter(user=user) if s.is_active()
```
abbas123456, line 26: The `is_active` method is used a lot in this PR when getting subscriptions, I think it would be better to add a custom manager to the DataSetSubscription model that can be called like `DataSetSubscription.objects.active(user)`. And the manager's `active` method can be defined as such:

```
from django.db.models import Q

def active(self, user):
    return self.filter(user=user, Q(notify_on_schema_change=True) | Q(notify_on_data_change=True))
```

dataworkspace/dataworkspace/apps/datasets/subscriptions/utils.py
```diff
@@ -10,3 +17,35 @@ def get_or_create_dataset_subscription_for_user(
...
+    ]
+
+    logger.info(len(subscriptions))
+    for subscription in subscriptions:
+        subscription.make_inactive()
+
+    DataSetSubscription.objects.bulk_update(
+        subscriptions, ["notify_on_data_change", "notify_on_schema_change"]
+    )
+
```
abbas123456, line 36: If you change the above to use a custom manager instead of a list, you can update the queryset directly instead of setting the attributes on each subscription and then bulk updating:

`subscriptions.update(notify_on_data_change=False, notify_on_schema_change=False)`

dataworkspace/dataworkspace/apps/datasets/subscriptions/utils.py
markhigham: good call. I'll do it. Thanks

dataworkspace/dataworkspace/apps/datasets/subscriptions/utils.py
markhigham: I learnt something new today!

abbas123456 (approved):

PR #1701, opened 2021-11-17
feature:add sso user id to google analytics

FREYA-KROL:
add sso user id to google analytics. This is to track new users of data workspace and the actions they complete, so we can track value for money for the comms and engagement campaign. 

### Checklist

* [ ] ~Have tests been added to cover any changes?~

### DIFF
--- a/dataworkspace/dataworkspace/templates/datasets/index.html
+++ b/dataworkspace/dataworkspace/templates/datasets/index.html
@@ -17,6 +17,7 @@
   <script nonce="{{ request.csp_nonce }}">
     dataLayer.push(
       {
+        "UserId": "{{ sso_user_id }}",
         "event": "filter",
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/templates/datasets/index.html
```diff
@@ -17,6 +17,7 @@
   <script nonce="{{ request.csp_nonce }}">
     dataLayer.push(
       {
+        "UserId": "{{ sso_user_id }}",
```
michalc, line 20: For consistency with the other dataLayer variables and with the spreadsheet linked from the ticket, should this be `userId`?

FREYA-KROL: UserId changed to userId now

michalc (approved):

PR #1702, opened 2021-11-17
feature: notify users about dataset structure changes by email

abbas123456:
### Description of change

Adds a celery task that sends an email notification when a source table is changed. It uses a `Notification` model to indicate that a changelog row has been processed by the task and uses a `UserNotification` object to indicate that a user has been emailed about a change.

This PR also changes the `get_table_changelog` method to figure out the changelog in Python rather than SQL. This gets around a bug in the current implementation where it treats a column being added and removed again as one change when it should be treated as two.

### Checklist


dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -614,3 +619,43 @@ def store_custom_dataset_query_table_structures():
                         Literal(int(DataSetType.DATACUT)),
                     )
                 )
+
+
+@celery_app.task()
+def send_notification_emails():
+    for table in SourceTable.objects.all():
```
michalc, line 626: So minor, but should this be ordered, say by ID, so it's just a bit more predictable how things get created?

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -614,3 +619,43 @@ def store_custom_dataset_query_table_structures():
...
+        changelog = get_table_changelog(
+            table.database.memorable_name, table.schema, table.table
+        )
+        for change in changelog:
```
michalc, line 630: I can't decide whether for the first item in the changelog no emails should be sent... 

- There is technically a race condition? A new dataset added with a source table, and published, and then someone subscribes before this task managed to run. They will be notified on the first item in the changelog, even if nothing has happened?

- But, say on an existing dataset, a new source table is added. This is probably a structure change, then people _should_ be notified?

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -614,3 +619,43 @@ def store_custom_dataset_query_table_structures():
...
+            notification, created = Notification.objects.get_or_create(
+                changelog_id=change['change_id']
+            )
```
michalc, line 633: Is there some way to be able to break out of the loop early? Essentially to not have to check absolutely every row of the changelog from now to the beginning...

dataworkspace/dataworkspace/datasets_db.py
```diff
@@ -94,51 +94,59 @@ def get_table_changelog(database_name: str, schema: str, table: str):
...
-                table_structure,
-                'Table structure updated' change_type
+            SELECT id, source_data_modified_utc, table_structure
             FROM dataflow.metadata
             WHERE table_schema = %s
             AND table_name = %s
             AND source_data_modified_utc IS NOT NULL
-            GROUP BY table_structure
-            ORDER BY change_date DESC;
+            ORDER BY id;
```
michalc, line 102: Would it be better to order DESC? So most recent changes are at the top?

dataworkspace/dataworkspace/apps/datasets/utils.py
michalc: Thinking about it... if the changelog is ordered with the most recent changes first, I think I would be tempted to only notify on the most recent data or schema change, so you could exit this loop early.

In later versions, we could get more fancy and notify on more exactly what has changed between this notification and the previous notification, but a small step to that is just notifying on the most recent one?

dataworkspace/dataworkspace/datasets_db.py
```diff
@@ -94,51 +94,59 @@ def get_table_changelog(database_name: str, schema: str, table: str):
...
-                record[columns[idx]] = field
-            record['change_date'] = record['change_date'].replace(tzinfo=pytz.UTC)
-            records.append(record)
-        return records
+        return get_changelog_from_metadata_rows(
+            cursor.fetchall(), 'Dataset query structure updated'
+        )
+
+
+def get_changelog_from_metadata_rows(rows, change_type):
```
michalc, line 128: Should this _not_ take the change type as an argument, but instead work it out? And in fact, I wonder if there isn't really a single change type, but it's more complicated. Not sure on the best structure, but something like:
```
{
  'change_id', ...
  'change_date', ...
  'is_initial': True/False,
  'table_structure': ...,
  'table_structure_changed': True/False(/None for `initial`?),
  'changed_data': True/False(/None for `initial`?),
}
```

So then calling code could filter on the changes it wants. Say, structure emails could filter on `table_structure_changed: True`

(And I would expect even more fields added down the line... with more detail on what has changed. What columns added/removed/changed type/re-ordered etc)

dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -2367,3 +2370,13 @@ class ToolQueryAuditLogTable(models.Model):
...
+
+
+class Notification(TimeStampedModel):
+    changelog_id = models.IntegerField(unique=True)
+
+
+class UserNotification(TimeStampedModel):
+    notification = models.ForeignKey(Notification, on_delete=models.PROTECT)
+    subscription = models.ForeignKey(DataSetSubscription, on_delete=models.PROTECT)
+    email_id = models.UUIDField()
```
michalc, line 2382: Should there be some sort of unique index/constraint to ensure only the 1 user notification for each notification/subscription pair?

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -614,3 +619,43 @@ def store_custom_dataset_query_table_structures():
...
+                        personalisation={
+                            'dataset_name': dataset.name,
+                            'change_date': change['change_date'],
+                        },
+                    )
+                    UserNotification.objects.create(
+                        notification=notification,
+                        subscription=subscription,
+                        email_id=email_id,
+                    )
```
michalc, line 661: I wonder if this should be split?

- In a big transaction, create the Notification and all the UserNotification objects. So if something goes wrong for any of them, it all aborts.

- Then afterwards, loop through "unsent" UserNotification objects. Trying to work out how to make this robust but also reduce the chance of sending someone multiple emails... I guess there should be some way to prevent multiple instances of the task from attempting to send the same email, but also some way to eventually retry if something went wrong?

michalc: On the robustness... it would be great if gov.uk notify used some sort of idempotency token so we could safely retry without the risk of actually duplicating emails. I don't _think_ they offer anything like that though...

dataworkspace/dataworkspace/apps/datasets/utils.py
abbas123456: >They will be notified on the first item in the changelog, even if nothing has happened?

Well technically something _has_ happened, the table has been created?

dataworkspace/dataworkspace/datasets_db.py
abbas123456: Agreed offline to do this at a later date once there are different types of emails. For now I will remove the `change_type` argument as this is only used on the front end and therefore belongs in the template

dataworkspace/dataworkspace/datasets_db.py
abbas123456: Achieved by reversing the return value of `get_changelog_from_metadata_rows` instead to avoid having to change the pairwise comparison logic

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -614,3 +619,84 @@ def store_custom_dataset_query_table_structures():
...
+                        email_id = send_email(
+                            settings.NOTIFY_DATASET_NOTIFICATIONS_TEMPLATE_ID,
+                            email_address,
+                            personalisation={
+                                'dataset_name': dataset_name,
+                                'change_date': change_date,
+                            },
+                        )
+                    except EmailSendFailureException:
+                        logger.info('Failed to send email')
```
michalc, line 689: Ah should this be logger.exception? I _think_ more ends up in sentry if this happens.

dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -614,3 +619,84 @@ def store_custom_dataset_query_table_structures():
...
+                    else:
+                        user_notification.email_id = email_id
+                        user_notification.save()
```
michalc, line 692: Tiny thing, but can this just be `save(update_fields=['email_id'])`. Doesn't make too much of a different in behaviour, but I think makes the code slightly clearer

dataworkspace/dataworkspace/notify.py
```diff
@@ -17,12 +17,20 @@ def decrypt_token(token):
...
+    response = client.send_email_notification(
         template_id=template_id,
         email_address=email_address,
         personalisation=personalisation,
         reference=reference,
     )
+    if 'id' in response:
+        return response['id']
+    else:
+        raise EmailSendFailureException
```
michalc, line 36: Should `response` be an argument to the exception? So hopefully it will appear in Sentry (especially if using `logger.exception` I think)

michalc (approved):

PR #1708, opened 2021-11-22
fix: add google-analytics to connect-src to fix blocked requests

markhigham:
### Description of change
- adding https://www.google-analytics.com to the csp `connect-src`
- adding s3ForcePathStyle to your_files client side javascript which make requests to `s3-hostname/bucketname` instead of `bucketname.s3.hostname`

### Checklist


dataworkspace/dataworkspace/apps/your_files/views.py
```diff
@@ -47,7 +47,7 @@ def file_browser_html_view(request):
     return file_browser_html_GET(request) if request.method == "GET" else HttpResponse(status=405)
 
 
-@csp_update(CONNECT_SRC=[settings.APPLICATION_ROOT_DOMAIN, "https://s3.eu-west-2.amazonaws.com"])
+@csp_update(CONNECT_SRC=["https://s3.eu-west-2.amazonaws.com"])
```
michalc, line 50: It really surprises me that this works! The browser makes an ajax request to APPLICATION_ROOT_DOMAIN to fetch credentials, which I thought uses CONNECT_SRC, so something's going on that I don't understand...

michalc (approved):

dataworkspace/dataworkspace/apps/your_files/views.py
michalc: Ah is csp_update more of an `append`?

michalc: Ah I see yes, it is https://django-csp.readthedocs.io/en/latest/decorators.html#csp-update

PR #1757, opened 2021-12-20
feat: filter cookies to and from visualisations

niross:
- Add generic `filter_headers()` function to replace `without_transfer_encoding`
- Set cookie headers to an empty string before forwarding requests to visualisations
- Filter out `Cookie` and `Set-Cookie` headers received from visualisations

### DIFF
--- a/dataworkspace/proxy.py
+++ b/dataworkspace/proxy.py
@@ -167,6 +167,12 @@ async def async_main():
             if key.lower() not in [x.lower() for x in exclude]
         )
 
+    def clear_cookie_header(headers: Union[CIMultiDict, CIMultiDictProxy]):
+        for key in list(headers.keys()):
+            if key.lower() == "cookie":
+                headers[key] = ""
+        return headers
+
     def admin_headers_request(downstream_request):
         # When we make a deliberate request to the admin application from the
         # proxy we don't want to proxy content-length or content-type
@@ -190,11 +196,7 @@ async def async_main():
         )
 
     def mirror_headers(downstream_request):
-        return tuple(
-            (key, value)
-            for key, value in downstream_request.headers.items()
-            if key.lower() not in ["host", "transfer-encoding"]
-        )
+        return filter_headers(downstream_request.headers, ["host", "transfer-encoding"])
 
     def application_headers(downstream_request):
         return (
@@ -705,7 +707,7 @@ async def async_main():
                 async with client_session.ws_connect(
                     str(upstream_url),
                     headers=(
-                        filter_headers(upstream_headers, ["Cookie", "Set-Cookie"])
+                        clear_cookie_header(upstream_headers)
                         if filter_cookies
                         else upstream_headers
                     ),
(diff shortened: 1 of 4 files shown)


dataworkspace/proxy.py
```diff
@@ -156,13 +157,22 @@ def csp_application_running_direct(host, public_host):
...
+        """
         return tuple(
             (key, value)
-            for key, value in request_or_response.headers.items()
-            if key.lower() != "transfer-encoding"
+            for key, value in headers.items()
+            if key.lower() not in [x.lower() for x in exclude]
         )
 
+    def clear_cookie_header(headers: Union[CIMultiDict, CIMultiDictProxy]):
```
michalc, line 170: Just out of paranoia really, can this not modify the passed headers, and instead return new set of headers? (Just in case somewhere either now or in future versions the headers object is used somewhere else, and the code assumes it's the original set of headers)

michalc: Actually... can it use the `filter_headers` function that returns a new set of headers?

niross: Good catch!

dataworkspace/proxy.py
```diff
@@ -797,23 +822,30 @@ async def handle_http(
...
             _, _, _, with_session_cookie = downstream_request[SESSION_KEY]
             downstream_response = await with_session_cookie(
                 web.StreamResponse(
                     status=upstream_response.status,
                     headers=CIMultiDict(
-                        without_transfer_encoding(upstream_response) + response_headers
+                        filter_headers(
+                            upstream_response.headers,
+                            ["transfer-encoding"]
+                            + (["Cookie", "Set-Cookie"] if filter_cookies else []),
```
michalc, line 846: Just to check on this, does it need `Cookie` header to be filtered out here? Would it be weird for an app to send the `Cookie` header to browsers, and even if it did, it would be ignored?

(I know it won't break anything having it filtered out, but since the proxy is complicated enough, want to try to make sure only things that are needed are in it)

dataworkspace/proxy.py
```diff
@@ -156,13 +157,19 @@ def csp_application_running_direct(host, public_host):
...
         return tuple(
             (key, value)
-            for key, value in request_or_response.headers.items()
-            if key.lower() != "transfer-encoding"
+            for key, value in headers.items()
+            if key.lower() not in [x.lower() for x in exclude]
         )
 
+    def clear_cookie_header(headers: Union[CIMultiDict, CIMultiDictProxy]):
+        return CIMultiDict(filter_headers(headers, ["cookie"]) + (("cookie", ""),))
```
michalc, line 171: Does it need the empty cookie header here? Remove if no (and maybe, comment why if yes?)

niross: Yes so originally that was what I did. It broke the visualisation I was testing with. As in it wouldn't load at all. Adding an empty string fixed it. I didn't look into it any further than that though

dataworkspace/proxy.py
niross: Yes definitely just me being overly cautious. I've removed it

michalc (approved):

PR #1770, opened 2022-01-13
changes published date for cloned master datasets to time of cloning

TomMacca:
### Description of change

cloned master datasets now get their published_at date at the time of cloning rather than inheriting it from the previous dataset 

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/models.py
+++ b/dataworkspace/dataworkspace/apps/datasets/models.py
@@ -410,6 +410,9 @@ class DataSet(DeletableTimestampedUserModel):
         New dataset is unpublished and has a name prefixed with
         "Copy of <original dataset name>".
 
+        The new datasets published_at date is set to the moment
+        that it was copied.
+
         Related objects (excluding user permissions) are duplicated
         for the new dataset.
 
@@ -422,6 +425,7 @@ class DataSet(DeletableTimestampedUserModel):
         clone.slug = ""
         clone.number_of_downloads = 0
         clone.published = False
+        clone.published_at = timezone.now()
         clone.save()
 
         for obj in self.related_objects():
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -422,6 +425,7 @@ def clone(self):
         clone.slug = ""
         clone.number_of_downloads = 0
         clone.published = False
+        clone.published_at = timezone.now()
```
michalc, line 428: Suspect this should be `None`, to then allow it to be set at the time the clone gets published

TomMacca: Thats a good point Michal! 

I've not worded it very well but my impression was the datetime should be at time of cloning rather than time of publishing. Should I clarify this or do you think it most likely should be at time of publishing?

michalc: I'm pretty sure it should be the time of publishing... for one thing, it says so in the ticket https://trello.com/c/w4twFpXf/2408-cloning-any-catalogue-item-should-not-copy-published-dates ;-)

(And from what I know of this issue, the ticket is correct. For reporting purposes, a cloned dataset time of publish should be the time it was first published, not the time it was cloned)

dataworkspace/dataworkspace/tests/datasets/test_models.py
```diff
@@ -22,6 +23,7 @@ def test_clone_dataset(db):
     assert clone.slug == ""
     assert clone.number_of_downloads == 0
     assert clone.name == f"Copy of {ds.name}"
+    assert clone.published_at == None
```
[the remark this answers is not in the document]
TomMacca: I fixed it before you sent your message, cut me some slack doc and use the next commit lol

michalc (approved):

PR #1786, opened 2022-01-18
feat: replace any cookie set from Django, rather than adding to it

michalc:
### Description of change

On staging, it's observed that _multiple_ set-cookie headers are received by the browser, and it's suspected that the first is set by Django. I'm not really sure why this hasn't been seen in prod. However, the cookies should be completely controlled by the proxy - any cookie set from Django should not make it back to the browser.

### Checklist


michalc:
This can't be done while storing the CSRF value in cookies... Django has to be able to set it in this case

PR #1790, opened 2022-01-19
fix: ensure pipeline is deleted if dag doesn't exist on data-flow

niross:
### Description of change

Fix the issue where deleting a pipeline failed if it didn't exist on data-flow. We should allow this incase the two systems get out of sync.

### Checklist


michalc (approved):

markhigham (approved):
LGTM. Although perhaps the error wording could be a bit more friendly.

There was a problem deleting the pipeline etc...

PR #1795, opened 2022-01-21
feat: enable virus scanning for file uploads

markhigham:
### Description of change
Enable a malware check for file uploads using DIT clamav service for all file uploads

### Checklist


dataworkspace/dataworkspace/apps/core/storage.py
```diff
@@ -1,23 +1,79 @@
...
+        clamav_response = _upload_to_clamav(content)
+
+        if clamav_response.malware:
+            msg = f"Virus found in {content.name} identified as {clamav_response.reason}"
+            logger.error(msg)
+            raise AntiVirusServiceErrorException(msg)
+
+        # this feels hacky! rewinding the file object so that it can be re-read!
+        # without it, we can't call the av_scan _and_ s3 upload
+        if hasattr(content, "seek"):
```
michalc, line 71: I don't think it seems too hacky! But also, why the `if`? (I would avoid having the `if` for something we always expect to be true...)

dataworkspace/dataworkspace/apps/core/storage.py
```diff
@@ -1,23 +1,79 @@
...
+logger = logging.getLogger("app")
+
+
+class AntiVirusServiceErrorException(UploadFileException):
+    pass
+
+
+class ClamAVResponse:
+    def __init__(self, json_response):
+        self.malware = json_response.get("malware", False)
```
michalc, line 25: If for some reason the `malware` key  is missing, it defaults to it thinking there isn't any malware?

All things being equal, I would lean to an exception if there isn't a malware key, so using `['malware']`, which woudl raise a `KeyError` might be good, rather than `get`?

markhigham: Yes. Good idea. I'll change

dataworkspace/dataworkspace/apps/core/storage.py
markhigham: This was to make the unit tests pass!
`dataworkspace/dataworkspace/tests/core/test_storage.py::test_file_save` which uses a bytes array as the file contents. In that scenario `contant` has no seek attribute

michalc: Can you make content have a seek attribute in the test? Now the production version depends on it, seems right that the version in the test should have it?

markhigham: 👍 done

michalc (approved):

PR #1812, opened 2022-02-01
Feature/display derived pipeline logs

TomMacca:
### Description of change

- adds a view logs page where a user can view the logs for their derived pipelines
image

- this page shows an error message that I've put in the the logs.html for when the data-flow server cannot respond due to 404 
image

### Checklist


dataworkspace/dataworkspace/apps/datasets/pipelines/views.py
```diff
@@ -151,3 +153,29 @@ def post(self, request, pk, *args, **kwargs):
...
+class PipelineLogsDetailView(DetailView, UserPassesTestMixin):
+    model = Pipeline
+    template_name = "datasets/pipelines/logs.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["errors"] = []
+        try:
+            context["logs"] = get_pipeline_logs(self.object)
+            context["errors"].append(("Logs retrieved successfully.", "success"))
```
niross, line 167: Does it feel a bit weird that we've got error = "success" here? I'm wondering if we can only set errors if there is an actual error and then in the template just check `{% if error %}<error message>{% else %}<list logs>{% endif %}`.

TomMacca: oh yeah definitely, I must of missed that - will amend. Theres problems with getting data-flow to build locally on m1's at the moment (me, Mark and Freya it seems) so I can't properly test this yet, will try to get things running and tested locally first and then push up

niross (approved):
Looks good, just need to sort out the tests/linting now :+1:

PR #1856, opened 2022-02-28
feature: validate pipeline sql when creating / editing derived pipelines

abbas123456:
This adds validation to ensure that:

- query is valid sql
- query is a single statement only
- query is not a CREATE or DROP statement
- query doesn't contain duplicate column names

### Checklist


niross (approved):
Nicely done :+1:

PR #1861, opened 2022-03-02
feature: remove poll_for_user_creation argument and dont log to sentry

abbas123456:
### Description of change

No longer need to conditionally poll for QuickSight user creation and no need to raise a sentry error if a user isn't found as register_user is now called when launching QuickSight

### Checklist


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -885,47 +884,26 @@ def sync_quicksight_permissions(user_sso_ids_to_update=tuple(), poll_for_user_cr
     account_id = boto3.client("sts").get_caller_identity().get("Account")
 
     quicksight_user_list: List[Dict[str, str]]
-    max_retries = settings.MAX_QUICKSIGHT_THROTTLE_RETRIES
     if len(user_sso_ids_to_update) > 0:
         quicksight_user_list = []
 
         for user_sso_id in user_sso_ids_to_update:
             # Poll for the user for 5 minutes
```
niross, line 891: Veery tiny, but this comment is wrong now, right?

niross (approved):
One tiny comment but looks good

dataworkspace/dataworkspace/apps/applications/utils.py
abbas123456: yes good spot

PR #1875, opened 2022-03-11
Feat use npm for front end css

markhigham:
### Description of change
Add node-sass as a build tool for front end assets.
Run the build from `npm`

### Checklist


README.md
```diff
@@ -133,6 +133,23 @@ We use [pip-tools](https://github.com/jazzband/pip-tools) to manage dependencies
...
+
+
+```
+  # this will configure node from .nvmrc or prompt you to install
+  nvm use
+  npm install
+  npm run build:css
+```
+
+Currently the generated CSS files are committed to source control. We will eventually automate this.
```
michalc, line 150: Can we not have the bit about the automation - it's a bit "todo" ish?

(And I guess I'm fairly happy with the generated CSS files being committed... reducing steps on deploy...)

michalc:
Realised - we set far future expires headers in nginx config https://github.com/uktrade/data-workspace/blob/master/etc/nginx/nginx.conf#L81

So as it stands, users could use old CSS with new HTML. I am fairly pro keeping the caching - is there a way to have both?

adr/front-end-css.md
```diff
@@ -0,0 +1,29 @@
+# Build frontend CSS using node-sass via NPM
+
+## Status
+
+Proposed: Waiting for review by data-infrastructure team
```
michalc, line 5: I guess once this PR gets merged in, the "Proposed" status is immediately out of data?

Also - should the adr documents be numbered?

package.json
```diff
@@ -0,0 +1,30 @@
...
+    "build:css": "node-sass dataworkspace/dataworkspace/sass --output dataworkspace/dataworkspace/static --include-path node_modules",
+    "debug:css": "node-sass dataworkspace/dataworkspace/sass --output dataworkspace/dataworkspace/static --source-map true --source-map-contents --include-path node_modules"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/uktrade/data-workspace.git"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
michalc, line 21: I think we're MIT license

package-lock.json
```diff
@@ -0,0 +1,2049 @@
+{
+  "name": "data-workspace-frontend",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
```
michalc, line 4: Should this now be version 2?

package.json
```diff
@@ -0,0 +1,30 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "build:css": "node-sass dataworkspace/dataworkspace/sass --output dataworkspace/dataworkspace/static --include-path node_modules",
+    "debug:css": "node-sass dataworkspace/dataworkspace/sass --output dataworkspace/dataworkspace/static --source-map true --source-map-contents --include-path node_modules"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/uktrade/data-workspace.git"
+  },
+  "keywords": [],
+  "author": "",
```
michalc, line 20: Can this be "Department for International Trade" - for consistency with the LICENSE file

package-lock.json
markhigham: normally yes. I wanted to use the same version of node/npm as govuk-frontend and they appear to be pinned at 14.x.x.
I figured that local dev and npm version issues may be reduced if we follow their lead ?

adr/front-end-css.md
markhigham: yes. So we end up needing another change after the PR is approved. Good point.

michalc:
How does this now handle the font issue - will the browser attempt to fetch the GDS font from the server and fallback to something else? (If so - is that... not good?)

markhigham:
> Realised - we set far future expires headers in nginx config https://github.com/uktrade/data-workspace/blob/master/etc/nginx/nginx.conf#L81
> 
> So as it stands, users could use old CSS with new HTML. I am fairly pro keeping the caching - is there a way to have both caching and the building of CSS?
> 
> Edit - actually - maybe there is some magic cachebusting going on? Having a nose around...

Good point. Not sure about cache busting. I half thought we did something inline but can't seem to find

michalc:
> about cache busting. I half thought we did something inline but can't seem to find

I think we do - I can't see the code for it, but looking at the file paths in the browser, they all seem to have some sort of hash added

PR #1884, opened 2022-03-14
fix: allow for filtering the grid on array fields

niross:
### Description of change

Adds equal/not equal and contains/not contains filters for array fields

### Checklist


markhigham (approved):
LGTM

PR #1915, opened 2022-04-11
Feat/chart builder user flows

niross:
### Description of change

- Move the chart builder to within the `core` app
- Adds multiple ways a chart can be created
  - Via data explorer query results
  - Via master dataset source table grid
  - Via data cut dataset custom query grid
  - Directly from a source table
  - Directly from a custom dataset query
- Adds a new chart section to data cut and master dataset catalogue pages

### Checklist


abbas123456 (approved):
Brilliant work

PR #1921, opened 2022-04-14
Feat/derived sharepoint pipelines

niross:
### Description of change

- Adds a new derived pipeline type `sharepoint`
- Requires sharepoint site and list names and lets data-flow handle the rest
- No major change to the ui other than an extra step confirming the type when you click "add new pipeline"
- At the moment this only supports lists stored directly under a site. Which is apparently the way all lists will be store going forward.
- If we really want to support legacy subsites will need a bit more work on the data-flow site (but nothing on the DW side)

### Checklist


markhigham (approved):
LGTM

PR #1929, opened 2022-04-21
feature: allow IAMs/IAOs to edit visualisation catalogue items

abbas123456:
### Description of change

https://trello.com/c/Za9knytm/2537-give-information-asset-owners-and-information-asset-managers-access-to-the-admin-page-for-visualisations

### Checklist


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -1280,28 +1285,74 @@ def get(self, request, **kwargs):
...
 
-class DatasetEditBaseView(View):
+class EditBaseView(View):
     dataset = None
     summary: str = None
 
     def dispatch(self, request, *args, **kwargs):
-        self.dataset = get_object_or_404(DataSet.objects.live(), pk=self.kwargs.get("pk"))
+        try:
+            dataset = DataSet.objects.live().get(pk=self.kwargs.get("pk"))
```
michalc, line 1294: I can't decide if it's too... odd... to assume that a VisualisationCatalogueItem ID can't be a DataSet ID. I mean, it's unlikely to the extreme, but it's not guaranteed by the database?

michalc: Is the dispatch method called explicitly by subclasses? Could that pass either `DataSet` or `VisualisationCatalogueItem` to this method - which would also then make it simpler since it wouldn't need to try both?

PR #1947, opened 2022-05-04
Feature - Add last updated flag for search results

nboyse:
### Description of change
In search results we need to be able to display the last updated date for source datasets / data cuts

### Checklist


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -203,6 +203,55 @@ def _enrich_tags(dataset, tags_dict):
...
+                    id=dataset["id"]
+                ).visualisationlink_set.all()
+                if not model:
+                    return None
+                date = model.first().data_source_last_updated
+                for table in model:
+                    last_updated = table.data_source_last_updated
+                    if last_updated and date:
+                        if last_updated > date:
+                            date = last_updated
```
michalc, line 247: Would it be better to refactor the above above to remove some of the duplication, e.g. something like the below where each dataset populates an iterable of last_updates_dates, and then choose the max of these (and then leveraging the `default` argument of `max`)

```python
if dataset["data_type"] == DataSetType.REFERENCE:
    last_updated_dates = (ReferenceDataset.objects.get(uuid=dataset["id"]).data_last_updated, )
elif dataset["data_type"] == DataSetType.MASTER:
    last_updated_dates = (table.get_data_last_updated_date() for table in MasterDataset.objects.get(id=dataset["id"]).sourcetable_set.all())
elif dataset["data_type"] == DataSetType.DATACUT:
    last_updated_dates = (query.get_data_last_updated_date() for query in DataCutDataset.objects.get(id=dataset["id"]).customdatasetquery_set.all())
elif dataset["data_type"] == DataSetType.VISUALISATION:
    last_updated_dates = (link.data_source_last_updated for link in VisualisationCatalogueItem.objects.get(id=dataset["id"]).visualisationlink_set.all())

return max(last_updated_dates, default=None)
```
There is a tiny bit of duplication between the master and datacut types, but to me, having one `if` block for each is a bit clearer + friendlier to other changes

Also on the exception - I wonder if better make it so the functions that fetch the date _don't_ raise an exception if there is nothing in the metadata table for them (To be honest I thought this was the case already!) But if that's not good/possible, can/should the try block a) catch a much more specific exception and b) only wrap the specific code that can raise that exception

michalc (approved):

PR #1960, opened 2022-05-16
fix: embiggen editor ui and add save status message

niross:
### Description of change

- Widen the chart builder to match data explorer's page width
- Show a confirmation message when the user clicks the save button

Screenshot from 2022-05-16 12-00-38

### Checklist


dataworkspace/dataworkspace/static/js/chart-builder/src/chart-builder/App.css
```diff
@@ -60,3 +62,38 @@
 .mapboxgl-ctrl-attrib {
   display: none !important;
 }
+@media only screen and (min-width: 1020px) {
+    /* Tablet and mobile displays already use full width */
+    .govuk-width-container {
+        max-width: 95%;
```
michalc, line 68: Feels a tiny but odd that this is a %, rather than a fixed margin either side in px, but I see Data Explorer does seem to behave that way as well.

PR #1961, opened 2022-05-17
fix: track users first login

niross:
### Description of change

Add `date_joined` to the user api results to enable querying for new data workspace users in SQL.

### Checklist


michalc (approved):

dataworkspace/dataworkspace/apps/accounts/backends.py
```diff
@@ -32,6 +33,11 @@ def authenticate(self, request, username=None, password=None, **kwargs):
             check_tools_access_if_user_exists=False,
         )
         set_user({"id": str(user.profile.sso_id), "email": user.email})
+
+        if user.profile.first_login is None:
+            user.profile.first_login = datetime.now()
+            user.profile.save()
```
michalc, line 39: There code for updating the user in certain cases in create_user_from_sso already - should this live in there somehow?

michalc (approved):

dataworkspace/dataworkspace/apps/accounts/backends.py
niross: Yes you're right, although I believe this is also called when syncing users from sso so we would run into the same issue as with `date_joined`

PR #1977, opened 2022-05-20
build(deps): move to pglast for better M1 support

michalc:
### Description of change

Suspect there were some unpinned deps as well, so those have been tidied up.

This also sneaks in a slightly more robust check on only allowing SELECT statements in the pipeline builder via an allowlist rather than a denylist.

### Checklist


markhigham (approved):
This should be a massive help for running on M1s
👍

PR #1980, opened 2022-05-20
Feature: create user facing newsletter sign up form

nboyse:
### Description of change
Currently, the distribution list for the monthly round up email is managed in an excel sheet, which users have to request to be added to manually by one of the team.

In line with improving the onboarding journey, we want to promote the newsletter and provide users with an easy way of signing up to this from Data Workspace.

Newsletter sign ups to be stored into data workspace so as a team we can monitor growth over time, export the list to import into Notify, manage preferences etc

### Checklist


markhigham (commented):
Should we put this behind a feature flag? Especially if DW changes are complete prior to the newsletter copy or sending mechanism

niross (approved):

niross (approved):

PR #2019, opened 2022-06-07
Feature: add big int as data type

nboyse:
### Description of change
User's downloads quite often fail due to columns containing whole numbers that exceed postgres' maximum integer size. The workaround to this is to use the numeric type but that can add a decimal point to the stored number which is not always what the user wanted.

We need to add BIGINT to the available data types on the your files and dataset uploaders. We should also confirm that when we try to determine column types automatically we select bigint when appropriate

### Checklist


dataworkspace/dataworkspace/apps/core/constants.py
```diff
@@ -12,7 +12,7 @@ class PostgresDataTypes(models.TextChoices):
 
 
 SCHEMA_POSTGRES_DATA_TYPE_MAP = {
-    "integer": PostgresDataTypes.INTEGER,
+    "bigint": PostgresDataTypes.BIGINT,
```
michalc, line 15: If this was:
```
"integer": PostgresDataTypes.BIGINT,
```
would it mean that we wouldn't have to have the conditional `if name != "bigint" else "Integer"` below?

nboyse: This would convert the integer to text in `your_files/views.py` lines 312-315

michalc: Ah I think I see - but does that feel odd somehow? We have a dictionaries to map between user facing and PostgreSQL types, but then we have to have a special case?

michalc (approved):

PR #2036, opened 2022-06-14
feat: avoid touching the database when extracting database tables

michalc:
### Description of change

The previous version caused issues with dataflow in terms of not having permission to drop the temporary views, as well as some cases of the views hanging around even though they are temporary

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/datasets_db.py
+++ b/dataworkspace/dataworkspace/datasets_db.py
@@ -9,8 +9,7 @@ import psycopg2
 from psycopg2.sql import Literal, SQL
 import pytz
 from django.conf import settings
-from django.db import connections, transaction
-from django.db.utils import DatabaseError
+from django.db import connections
 
 from dataworkspace.apps.datasets.constants import DataSetType
 from dataworkspace.utils import TYPE_CODES_REVERSED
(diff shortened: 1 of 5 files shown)


dataworkspace/dataworkspace/datasets_db.py
```diff
@@ -107,28 +106,44 @@ def last_updated_date_for_database(database_name, tables_for_database: Tuple[Tup
...
-        except DatabaseError as e:
-            logger.error(e)
-            tables = []
-        else:
-            cursor.execute(
-                "select table_schema, table_name from information_schema.view_table_usage where view_name = 'get_tables'"
-            )
-            tables = cursor.fetchall()
-            cursor.execute("drop view get_tables")
+def extract_queried_tables_from_sql_query(query):
```
markhigham, line 109: I think it would be helpful to have return types defined here.
I _think_ its a list of tuples of `schema` `table` but I had to read the contents of the function

niross (approved):

dataworkspace/dataworkspace/tests/test_datasets_db.py
```diff
@@ -23,13 +23,74 @@
...
+                        my_cte_1 AS (SELECT * FROM my_inner_1),
+                        my_inner_2 AS (SELECT * FROM my_inner_2)
+                    SELECT * FROM my_cte_1
+                )
+            """,
+            [("public", "my_inner_1"), ("public", "my_inner_2")],
+        ),
     ),
 )
 @pytest.mark.django_db
```
markhigham, line 91: Do we still need this if we don't touch the db?

markhigham (approved):

PR #2051, opened 2022-06-21
feat: hide internal fields from user facing sort options

niross:
### Description of change

hide internal field names from the search results query string by mapping them to descriptive names. Any requests for the old style sort params will be translated to the new type.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/forms.py
+++ b/dataworkspace/dataworkspace/apps/datasets/forms.py
@@ -9,7 +9,7 @@ from django.contrib.auth import get_user_model
 
 from dataworkspace.apps.datasets.constants import DataSetType, TagType
 from .models import DataSet, SourceLink, Tag, VisualisationCatalogueItem
-from .search import SearchDatasetsFilters
+from .search import LEGACY_SORT_FIELD_MAP, SORT_FIELD_MAP, SearchDatasetsFilters
 from ...forms import (
     GOVUKDesignSystemForm,
     GOVUKDesignSystemCharField,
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/apps/datasets/forms.py
```diff
@@ -410,7 +401,9 @@ def get_filters(self):
         filters.with_visuals = "withvisuals" in self.cleaned_data.get("admin_filters")
         filters.use = set(self.cleaned_data.get("use"))
         filters.data_type = set(self.cleaned_data.get("data_type", []))
-        filters.sort_type = self.cleaned_data.get("sort")
+        filters.sort_type = SORT_FIELD_MAP.get(
+            self.cleaned_data["sort"], self.cleaned_data["sort"]
```
michalc, line 405: Does the default here mean that if the user passes something that isn't in the SORT_FIELD_MAP, then it gets passed through to the Django query functions? Would it be better to not have this change, and instead have both new and legacy in the one map, where the legacy ones map to themselves, e.g.:

```
SORT_FIELD_MAP = {
    "-published": "-published_date,-search_rank,name",
    "published": "published_date,-search_rank,name",
    "alphabetical": "name",
    "relevance": (
        "-is_bookmarked,-table_match,-search_rank_name,-search_rank_short_description,"
        "-search_rank_tags,-search_rank_description,-search_rank,-published_date,name"
    ),
    "popularity": "-average_unique_users_daily,-published_date,name",

    # Legacy - to be removed
    "-published_date,-search_rank,name": "-published_date,-search_rank,name",
    "published_date,-search_rank,name": "published_date,-search_rank,name",
    "name": "name",
    "-is_bookmarked,-table_match,-search_rank_name,-search_rank_short_description,"
    "-search_rank_tags,-search_rank_description,-search_rank,-published_date,name": (
        "-is_bookmarked,-table_match,-search_rank_name,-search_rank_short_description,"
        "-search_rank_tags,-search_rank_description,-search_rank,-published_date,name"
    ),
    "-average_unique_users_daily,-published_date,name"": "-average_unique_users_daily,-published_date,name"
}
```

niross: yes good call. I must admit I totally confused myself with this one due to the django choice fields baffling me for some reason

dataworkspace/dataworkspace/apps/datasets/forms.py
```diff
@@ -256,8 +240,15 @@ def clean_sort(self):
...
 
-    def __init__(self, request, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, request, data, *args, **kwargs):
+        # This translation can be removed 1-2 days after rollout
+        # Ensure legacy search field names work with new field names by
+        # translating old field name -> new field name
+        if data.get("sort") in LEGACY_SORT_FIELD_MAP:
+            data = data.copy()
+            data["sort"] = LEGACY_SORT_FIELD_MAP[data["sort"]]
```
michalc, line 249: Is this now not needed, since everything is in SORT_FIELD_MAP?

niross: The problem I had with removing it is, django validates the form and it will redirect back to the search page if the form is invalid. If we add the legacy values _and_ the new values to the form there will be duplicates in the sort dropdown on the search page. If we remove the legacy values then we must map them to the new values somehow. The whole thing seems overly complicated to be honest.

niross: Yeah as expected removing the mapping fails the unit tests because the old fields don't exist in the sort dropdown on the search page. So form validation fails.

michalc: Ah ok

PR #2061, opened 2022-06-24
Feat/schema table search

niross:
### Description of change

- Allow for searching for tables with fully qualified name `<schema>.<table>`
- Adds support for searching for reference dataset by table name and schema.table

Adding this to the filters unfortunately means doing an aggregation before the first filters are applied. This is so we can concatenate the schema and table before filtering. 

### Checklist


michalc:
On

> Adding this to the filters unfortunately means doing an aggregation before the first filters are applied. 

I wonder if we could make the assumption that all-but-the-last-dot is part of the schema name, so we then can make a good guess as to query_schema, and query_table, and have to aggregate before the first filters.

niross:
> On
> 
> > Adding this to the filters unfortunately means doing an aggregation before the first filters are applied.
> 
> I wonder if we could make the assumption that all-but-the-last-dot is part of the schema name, so we then can make a good guess as to query_schema, and query_table, and so _not_ have to aggregate before the first filters.

Ok yeah i've split the query and used it to match on schema/table

dataworkspace/dataworkspace/apps/datasets/search.py
```diff
@@ -390,19 +407,36 @@ def _annotate_is_bookmarked(datasets, user):
...
     if datasets.model is DataSet:
-        datasets = datasets.annotate(
+        schema, table = _schema_table_from_search_term(query)
+        return datasets.annotate(
             table_match=BoolOr(
                 Case(
-                    When(sourcetable__table=query, then=True),
+                    When(
+                        Q(sourcetable__table=query)
+                        | Q(Q(sourcetable__table=table) & Q(sourcetable__schema=schema)),
```
michalc, line 428: The Q logic looks slightly different to the logic in _filter_by_query...is this fine? (Although it might be equivalent...)

Should it be more clearly the same?

michalc (approved):

PR #2064, opened 2022-06-27
feature: proxy to mlflow UI and generate JWT when launching tools

abbas123456:
### Checklist


mlflow/proxy.py
```diff
@@ -0,0 +1,233 @@
...
+            for key, value in request_or_response.headers.items()
+            if key.lower() != "transfer-encoding"
+        )
+
+    def headers_proxy(downstream_request):
+        return tuple(
+            (key, value)
+            for key, value in downstream_request.headers.items()
+            if key.lower()
+            in required_admin_headers + ("content-length", "content-type", "authorization")
```
michalc, line 70: Ah should 
- `required_admin_headers` be changed to include `("content-length", "content-type", "authorization")`, since only the concatenated list used
- Some of the headers removed from `required_admin_headers` that aren't needed? Some of them look a bit Django-specific
- `required_admin_headers` renamed to something like `allowed_headers` (there is not admin application really here)

mlflow/proxy.py
```diff
@@ -0,0 +1,233 @@
...
+        upstream_url = URL(upstream_root).with_path(path)
+
+        if "Authorization" not in downstream_request.headers:
+            return web.Response(status=401)
+
+        jwt_token = downstream_request.headers["Authorization"][7:].encode()
+        header_b64, payload_b64, signature_b64 = jwt_token.split(b".")
+        payload = json.loads(b64_decode(payload_b64))
+
+        public_key = load_pem_public_key(jwt_public_key.encode())
```
michalc, line 92: Very minor thing, but probably good to do this on startup? A bit of an optimisation, but also I think good to raise an exception then if something isn't right, which would exit, and so hopefully prevent deployment

mlflow/proxy.py
```diff
@@ -0,0 +1,227 @@
...
+            downstream_request,
+            method,
+            CIMultiDict(headers_proxy(downstream_request)),
+            upstream_url,
+            query,
+            await get_data(downstream_request),
+            default_http_timeout,
+        )
+
+    async def get_data(downstream_request):
```
michalc, line 113: I think this doesn't need to be an async function, since it doesn't do any `await`ing inside?

niross (approved):

PR #2082, opened 2022-07-18
feat: add data dictionary for reference datasets

markhigham:
### Description of change
Adding data dictionary for reference datasets.
Because reference datasets are a separate model I created a new `DataDictionaryService` which handles the getting and saving of dictionaries. This provides a couple of helper classes that clients can use to interact with the service.

### Checklist


dataworkspace/dataworkspace/apps/datasets/data_dictionary/service.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _update_source_table(
+        self, source_table, rows: List[DataDictionary.DataDictionaryUpdateRow]
+    ):
+        logger.debug("updating source table %s", source_table)
+
+        for row in rows:
+            field, _ = SourceTableFieldDefinition.objects.get_or_create(
+                source_table=source_table, field=row.name
+            )
```
abbas123456, line 129: Can we check if the `SourceTableFieldDefinition` was created and if so skip the below so it doesn't call save unnecessarily

dataworkspace/dataworkspace/apps/explorer/schema.py
```diff
@@ -138,6 +138,9 @@ def match_datasets_with_schema_info(schema):
     schema_table_names = [
         Func(Value(s.name.schema), Value(s.name.name), function="Row") for s in schema
     ]
+
+    logger.debug(schema_table_names)
```
abbas123456, line 142: Was this left in by mistake?

abbas123456 (approved):

dataworkspace/dataworkspace/apps/datasets/data_dictionary/service.py
```diff
@@ -0,0 +1,145 @@
...
+        return dictionary
+
+    def _update_source_table(
+        self, source_table, rows: List[DataDictionary.DataDictionaryUpdateRow]
+    ):
+        logger.debug("updating source table %s", source_table)
+
+        for row in rows:
+            field, created = SourceTableFieldDefinition.objects.get_or_create(
+                source_table=source_table, field=row.name, description=row.definition
```
abbas123456, line 105: I don't think this will work in the `get` path of `get_or_create` as it will only return `SourceTableFieldDefinition` objects where the `description` matches `row.definition`. You will need to pass the `description` kwarg in the defaults argument like so in order for it to only be used in the `create` path:

```
field, created = SourceTableFieldDefinition.objects.get_or_create(
    source_table=source_table,
    field=row.name,
    defaults={'description': row.definition}
)
```

PR #2085, opened 2022-07-18
feat: add enhanced table, apply to saved queries

niross:
### Description of change

- adds a new js "pckage" to turn an existing govuk styled table into a govuk styled grid
- falls back to standard table if javascript is disabled
- slight refactor of the data grid js to allow for sharing of certain functions
- updates the saved queries page to merge the two tables and add sort functionality
- change grid css to sass so we can add our own styles

The reason I went with table replacement rather than populating with json is:

1. tables can often contain multiple buttons, links and forms within a cell and rendering that into json started getting messy.
2. we can fall back to standard html table s if javascript is disabled

### Example

The top table is a standard govuk table. Below is a similar table with the class `.enhanced-table` added and column config set as data attributes.

dataworkspace test_8000_data-explorer_queries_

### Checklist


dataworkspace/dataworkspace/sass/_grid.scss
```diff
@@ -0,0 +1,97 @@
...
+  .ag-grid-header-cell, .ag-header-cell-label, .ag-cell-label-container, .ag-header-cell-sortable  {
+    cursor: default;
+  }
+  .ag-header-cell-sorted-none button, .ag-header-cell-sorted-asc button, .ag-header-cell-sorted-desc button {
+    color: rgb(29, 112, 184);
+    cursor: pointer;
+  }
+  .ag-header-cell-sorted-none button:focus, .ag-header-cell-sorted-asc button:focus, .ag-header-cell-sorted-desc button:focus {
+    background-color: #fd0;
+    color: #0b0c0c;
```
markhigham, line 54: perhaps we could use the govuk colour constants here?
https://design-system.service.gov.uk/styles/colour/

dataworkspace/dataworkspace/static/js/enhanced-table.js
```diff
@@ -0,0 +1,99 @@
+function initEnhancedTable() {
+  // Given an html table, convert it to a sortable grid
+  var tables = document.getElementsByClassName("enhanced-table");
```
markhigham, line 3: I would pass the classname to the `initEnhancedTable` function rather than hard code it

markhigham (commented):
Some minor comments

markhigham (approved):

PR #2091, opened 2022-07-20
Feat/query log design update

niross:
### Description of change

- Add new actions to the query log table
- Add last run column to the aforementioned table

Screenshot 2022-07-20 at 11 18 05

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/static/explorer_custom.css
+++ b/dataworkspace/dataworkspace/static/explorer_custom.css
@@ -136,3 +136,20 @@
   100% { transform: rotate(360deg); }
 }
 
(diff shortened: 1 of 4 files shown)


dataworkspace/dataworkspace/static/explorer_custom.css
```diff
@@ -136,3 +136,20 @@
   100% { transform: rotate(360deg); }
 }
 
+.button-link {
```
markhigham, line 139: doesn't govuk provide this out of the box?

markhigham (approved):

PR #2100, opened 2022-07-23
feat: include dots in identifiers from the point of view of auto-completion

michalc:
### Description of change
The table list is formed of schema.table_name pairs, but the default regex used to find the current identifier does not include dots. This mis-match caused the autocompletion to ignore the schema name in the existing query, and re-add the whole schema.table_name pair, resulting in schema.schema.table_name, which then resulted in the server complaining about cross database references not being implemented.

### Checklist


markhigham (approved):
This is certainly better than current functionality

PR #2125, opened 2022-08-15
feat: rewrite your_files in react

markhigham:
### Description of change
A like for like translation of the your_files tool from angular.js to react

This is an interim PR without full functionality.
It will be available at the `/files/react` url

### Checklist


niross (approved):
Looks good 🎉 I will deploy what we've got and iterate on the last few bits

PR #2132, opened 2022-08-26
Feat/your files react error modal

niross:
### Description of change

Adds an error modal to match what we have in the angular version

Screenshot 2022-08-26 at 17 19 35

### Checklist


dataworkspace/dataworkspace/static/js/react_apps/src/your-files/utils.js
```diff
@@ -46,3 +46,11 @@ export function getFolderName(prefix, rootPrefix) {
   if (!folder) return `${ROOT_FOLDER_NAME}/`;
   return prefixToFolder(folder);
 }
+
+export function objEntries(obj) {
+    var ownProps = Object.keys(obj);
```
markhigham, line 51: I think we are safe to use const or let here

dataworkspace/dataworkspace/static/js/react_apps/src/your-files/utils.js
```diff
@@ -46,3 +46,11 @@ export function getFolderName(prefix, rootPrefix) {
   if (!folder) return `${ROOT_FOLDER_NAME}/`;
   return prefixToFolder(folder);
 }
+
+export function objEntries(obj) {
```
markhigham, line 50: I think this is complicated. Not sure what it's for or why returning array of arrays(?) from object properties.

dataworkspace/dataworkspace/static/js/react_apps/src/your-files/components/ErrorModal.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
```
markhigham, line 1: should we have components and popups folders separate?

markhigham (approved):
Some minor comments. LGTM 👍

dataworkspace/dataworkspace/static/js/react_apps/src/your-files/utils.js
niross: ohh yes good catch thanks

dataworkspace/dataworkspace/static/js/react_apps/src/your-files/utils.js
niross: Yeah I agree, this is basically a lift and shift from the angular version. I'll have a think about how it can be de-complicated

PR #2137, opened 2022-09-05
Feature/show number of results in data preview

clonecmdr:
### Description of change

https://user-images.githubusercontent.com/19953203/188815390-2417a9c6-4a6b-4c44-8e63-032400b60b4f.mov

### Checklist


dataworkspace/dataworkspace/apps/datasets/utils.py
```diff
@@ -567,7 +568,14 @@ def build_filtered_dataset_query(inner_query, column_config, params):
         SQL(",").join(map(Identifier, sort_fields)),
     )
 
-    return query, query_params
+    where_clause = Composed(where_clause)
+    if download_limit is None:
+        download_limit = 5000
+    limit_clause = Composed([SQL(f" LIMIT {download_limit + 1}")])
```
michalc, line 574: The most minor thing - and if we do it elsewhere but I think we really shouldn't - to not use f-strings when composing SQL, pretty much never. I know download_limit is probably "safe", but just in case in a future version of the code it's not, it feels a bit SQL-injection-risk-y.

Is there another pattern that can be used?

nboyse (approved):

PR #2193, opened 2022-10-11
feat: removed chart builder reference from welcome screen

grahambirdtrade:
### Description of change

* Commented out block about chart builder as this feature is now hidden
* Changed background colours of other blocks to account for missing block
* Switched image alignment of other blocks to account for missing block
* Reordered Find out more bullets to lead with newsletter and Teams

I haven't tested it locally as I am editing in the browser.


nboyse (approved):
Tested locally, looks good

PR #2194, opened 2022-10-11
perf: faster check for existing privileges

michalc:
### Description of change

Sometimes this was so slow it was hitting 504s and preventing tools from loading

Seems to cut times down from > 60 seconds to ~12 seconds. The `has_table_privilege` function I think was the cause.

Potentially could make this faster still - information_schema.table_privilege is a view that probably queries too much. Maybe could get faster by going to the pg_* tables directly.

### Checklist


niross (approved):
Nice!

PR #2196, opened 2022-10-13
perf: speculative removal of server side cursor for data grid

michalc:
### Description of change

For some reason sometimes a server side cursor can make a query less performant. Here, we don't need to take that risk since we're fetching all the rows and loading them into memory anyway, so there is no benefit of a server side cursor.

To help debugging production issues, we now set application_name instead, which is viewable by (for example) pg_activity

### Checklist


clonecmdr (approved):
I'm assuming the label is to filter out data-grid queries with pg_stat_activity

PR #2201, opened 2022-10-18
feat: switch to localstack for s3/sts

niross:
### Description of change

Allows your files to be run locally without any extra env faff.

### Checklist


michalc:
Oh wow

PR #2277, opened 2022-11-08
refactor: remove our code for yellow outline in alerts

michalc:
### Description of change

This was added in 09f7f7f471b149efa245a4f374fea3c3ec70fa78, but now GDS seems to add their own CSS that does exactly this, so removing ours. This was noticed when initially considering removing the yellow outline

DT-780

### Checklist


michalc:
Leaving this for now - not urgent

PR #2298, opened 2022-11-17
Feat/add users to collection

niross:
### Description of change

Add "add user" form to the collection users page

DT-531

https://user-images.githubusercontent.com/594496/202417399-235e95fb-7515-4c6c-a654-0fb0ff481b08.mov

### Checklist


dataworkspace/dataworkspace/apps/data_collections/forms.py
```diff
@@ -53,3 +56,34 @@ class CollectionUserForm(forms.ModelForm):
...
+    def clean_email(self):
+        user = get_user_model().objects.filter(email=self.cleaned_data["email"]).first()
+        if user is None:
+            raise forms.ValidationError(
+                "The user you are sharing with must have a DIT staff SSO account"
+            )
+        if self.collection.owner == user:
+            raise forms.ValidationError(f"{user.email} already has access to this collection")
+        if self.collection.user_memberships.live().filter(user=user).exists():
+            raise forms.ValidationError(f"{user.email} already has access to this collection")
```
michalc, line 88: Is there potentially a race condition here - after this check, someone else could have added the user, and then this user could get a 500 from an integrity error?

Should it instead do same thing that `select_collection_for_membership` does - attempt to add in the user, and catch the integrity error?

michalc (approved):

PR #2482, opened 2023-03-30
feat:DT:1261 create an API to expose teams schema name, members and id

Teccolat:
### Checklist


dataworkspace/dataworkspace/apps/api_v1/core/serializers.py
```diff
@@ -13,3 +13,14 @@ class NewsletterSubscriptionSerializer(serializers.ModelSerializer):
     class Meta:
         model = NewsletterSubscription
         fields = "__all__"
+
+
+class TeamSerializer(serializers.ModelSerializer):
+    member = serializers.SerializerMethodField()
```
niross, line 19: Small one but can we call this `members` and the method `get_members(...)`. As a team can have more than on member

niross (approved):
Great work 🎉

PR #2496, opened 2023-04-06
build: remove /app folder from docker-compose volumes

michalc:
### Description of change

It's the same folder as the dataworkspace folder, so shouldn't need to be in two places

### Checklist


michalc:
Misunderstood this

PR #2659, opened 2023-06-14
feat: Catalogue Editor: Django Admin

nboyse:
### Description of change
This is the implementation for providing a Data Catalogue Editor role. It includes the changes for:

- Django Admin to allow addition of existing users to be the Data Catalogue Editors per catalogue item
- Allow data catalogue editors to manage dataset
- Amend search for data I own or manage to include datasets users are a data catalogue editor for
- Allow information asset managers, owners and superusers to manage editors in a similar way that permissions are granted to datasets
- Allow data catalogue editors to manage permissions of users to the dataset
- Expose links

It applies to Source datasets, Data cuts and Visualisation catalogue items but not Reference datasets as these are not restricted.

### Checklist


niross (approved):
Great work 🔥

PR #2854, opened 2023-11-10
Show source dataset notification 

ian-leggett:
### Description of change

This change surfaces a notification to the user when the dataset is no longer updated. Also this change fixes two small bugs:

1. Adds text to the "Report an issue" flag within the data cut preview page
2. Increases an offset value to reduce multiple scroll bars on data preview pages

### Checklist


oliver-weare (approved):
Looks good!

PR #2954, opened 2024-01-17
Fix styles for the welcome page search bar

ian-leggett:
### Description of change

Currently the styles are broken for the search bar on the `/welcome` page. This change adds the correct styling and is makes the search bar consistent with other internal pages.

### Checklist


oliver-weare (approved):
Looks good!

PR #2985, opened 2024-01-30
Add feedback banner for data download with cookie

CaitBarnard:
### Description of change
A feedback banner appears on the modal after downloading data. This banner will only appear once a week and this is set in a cookie. The banner has a link to the feedback form.

The feedback form, when accessed through the link should pre-select 'Analyse data' in the first question. The feedback form also stores the location of where the user redirected to the form from. If they accessed it the normal route then it is set to 'contact-us'. If they accessed it via the feedback form link on the banner then it is set to 'csat-download-link'

### Screenshots
Screenshot 2024-02-06 at 15 57 43
### Checklist


dataworkspace/dataworkspace/sass/data-workspace.scss
```diff
@@ -531,6 +531,15 @@ p.sidebar-section {
...
 
+#download-dialog {
+  .govuk-link {
+    font-size: 19px;
+  }
+  .text--margin {
+    margin-bottom: 10px
+  }
+}
+
```
ian-leggett, line 542: Would it be possible to move these rules inside the `dataworkspace/dataworkspace/sass/_dialog.scss` file?

ian-leggett (approved):

dataworkspace/dataworkspace/apps/core/views.py
```diff
@@ -255,6 +255,11 @@ class UserSatisfactionSurveyView(FormView):
     form_class = UserSatisfactionSurveyForm
     template_name = "core/user-satisfaction-survey.html"
 
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs["trying_to_do_initial"] = self.request.GET.getlist("from")
```
chopkinsmade, line 260: `from` in a url is usually associated with pagination, could this variable be named `survey_source` instead to match the name of the new model property?

dataworkspace/dataworkspace/apps/core/forms.py
```diff
@@ -72,6 +72,18 @@ def clean(self):
 
 
 class UserSatisfactionSurveyForm(GOVUKDesignSystemForm):
+    def __init__(self, *args, **kwargs):
+        trying_to_do_initial = kwargs.pop("trying_to_do_initial", None)
```
chopkinsmade, line 76: Does this need an additional prop named `trying_to_do_initial` instead of setting it to `survey_source` in the views.py?

chopkinsmade (approved):

PR #2986, opened 2024-01-31
chore/DT-1804 Add homepage E2E tests

ian-leggett:
### Description of change

This change adds E2E/functional tests around the homepage dashboard.

### Checklist


cypress/e2e/datasets/home-page.cy.ts
```diff
@@ -0,0 +1,170 @@
+const endpoints = {
+  recentItems: "/api/v2/recent_items/*",
+  recentCollections: "/api/v2/collections/*",
+  recentTools: "/api/v2/recent_tools/*",
+  yourBookmarks: "/api/v2/your_bookmarks/*",
+} as const;
+
+describe("Homepage dashboard", () => {
+  context("When any user visits the page", () => {
```
chopkinsmade, line 9: There might be a linter for this, but can you add line breaks between the context and it statements to make this a bit more readable?

chopkinsmade (approved):

cypress/e2e/datasets/home-page.cy.ts
ian-leggett: Good spot, I have now spaced the IT statements out for readability. I am going to create another ticket to update the linter so we get these checks put in place 👍

chopkinsmade (approved):

PR #2987, opened 2024-01-31
feat: add cypress testing library

ian-leggett:
### Description of change

This change add [Cypress testing library](https://testing-library.com/docs/cypress-testing-library/intro/). By adding this package we can use the testing library API when writing Cypress tests. This library encourages your applications to be more accessible and allows you to get your tests closer to using your components the way a user will, which allows your tests to give you more confidence that your application will work when a real user uses it.

### Checklist


chopkinsmade (approved):

package.json
```diff
@@ -30,9 +30,11 @@
   },
   "homepage": "https://github.com/uktrade/data-workspace#readme",
   "devDependencies": {
+    "@testing-library/cypress": "^10.0.1",
```
chopkinsmade, line 33: I think these need to be in the cypress/package.json file as well as thats what the docker image uses

PR #2993, opened 2024-02-02
feature/DT-1803-grant-access-to-dataset

chopkinsmade:
### Description of change
When an email address is added to the request approvers field, if there is a matching user with that email also add them to the catalogue editors field.

When a catalogue editor is removed using the template form, also remove them from the request approvers field
### Checklist


ian-leggett (approved):
LGTM

CaitBarnard (approved):

PR #2994, opened 2024-02-02
fix/DT-1824 Guard against null being returned

ian-leggett:
### Description of change

**The problem scenario:**
- You have a dataset listed on the home page as a recently visited item
- The owner of that dataset deletes it
- The "Recent items" component hangs as it tries to retrieve a name from the response object that isn't there

**Solution**
This change just guards for the above scenario and filters out any returned items with no name
 
### Checklist


dataworkspace/dataworkspace/static/js/react/transformers.ts
```diff
@@ -33,7 +33,7 @@ export const transformRecentItemsResponse = (
   response: YourRecentIemsResponse
 ): TransformedYourRecentItemsResponse =>
   response.results.map(({ related_object, extra }) => ({
-    name: related_object.name,
+    name: related_object.name || null,
```
chopkinsmade, line 36: I think the issue was related_object was null? So this would still cause an issue as your accessing name

chopkinsmade (approved):

PR #3006, opened 2024-02-07
bugfix/DT-1770-add-localstack-into-docker-compose

chopkinsmade:
### Description of change

- Add the localstack docker image in to allow local uploading of files to an emulated S3 environment
- Exclude the AV check running when on local

### Checklist


ian-leggett (approved):
🔥

PR #3021, opened 2024-02-12
feature/DT-1823-automate-react-compilation

chopkinsmade:
### Description of change
- Added an new docker stage that will build the react bundles. The output of this step is copied into the later docker stages
- Removed the react bundle files from github to avoid any confusion about where these files are coming from. 
- When developing locally the docker compose will ignore the version in the Dockerfile, and use the version in your dev machine. This allows continuation of using running `npm run dev` and using the hot reload
- On the E2E and a11y scripts, add an empty volume to force the docker container to use the bundle files inside the docker container, and not the versions in the host

### Checklist


dataworkspace/dataworkspace/static/js/react/features/home-page/components/RecentItems/index.tsx
```diff
@@ -50,7 +50,7 @@ const RecentItemsListItem: React.FC<RecentItemProps> = ({ url, name }) => (
 const RecentItems: React.FC<Record<'items', RecentItemProps[]>> = ({
   items
 }) => (
-  <Tile as="article" title="Your recent items">
+  <Tile as="article" title="Your recent items COMPILED FROM DOCKERFILE">
```
chopkinsmade, line 53: Make sure to remove this before merging

PR #3041, opened 2024-02-19
feature/DT-1830-your-files-accessible-modal

chopkinsmade:
### Description of change
Switch the file uploader modals to use a `<dialog>` to be the same as the other modals used in the rest of data workspace

### Checklist


dataworkspace/dataworkspace/static/js/react/features/your-files/popups/ErrorModal.js
```diff
@@ -1,53 +1,63 @@
...
-                type="button"
-                onClick={onClose}
-                className="govuk-button govuk-button--secondary modal-button"
-              >
-                Close
-              </button>
+          <div className="modal-contents">
+            <div className="modal-body">
+              <div className="panel-body">
+                <p className="govuk-body" ng-if="error.errors.length">
```
ian-leggett, line 21: This condition is using an Angular directive. Can we change this to use React instead?

dataworkspace/dataworkspace/static/js/react/features/your-files/popups/ErrorModal.js
```diff
@@ -1,53 +1,63 @@
...
+                  {error.message}
+                </p>
+                <table className="govuk-table" ng-if="error.errors.length">
```
ian-leggett, line 24: Same as above

ian-leggett (commented):
Looks good. Just one thing, I noticed some conditions using Angular directives, can we remove this or do we need it?

dataworkspace/dataworkspace/static/js/react/features/your-files/popups/ErrorModal.js
chopkinsmade: Doesn't seem to have any effect, removed and errors still appear

ian-leggett (approved):

PR #3042, opened 2024-02-19
fix: set quicksight email to DW user email when syncing users

niross:
### Description of change

Fixes issue where quicksight user emails were never updated and users were not receiving notifications

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -837,10 +837,15 @@ def create_update_delete_quicksight_user_data_sources(
 def sync_quicksight_users(data_client, user_client, account_id, quicksight_user_list):
     for quicksight_user in quicksight_user_list:
         user_arn = quicksight_user["Arn"]
-        user_email = quicksight_user["Email"].lower()
         user_role = quicksight_user["Role"]
         user_username = quicksight_user["UserName"]
         sso_id = quicksight_user["UserName"].split("/")[-1]
+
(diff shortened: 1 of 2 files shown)


michalc (approved):

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -837,11 +837,16 @@ def create_update_delete_quicksight_user_data_sources(
 def sync_quicksight_users(data_client, user_client, account_id, quicksight_user_list):
     for quicksight_user in quicksight_user_list:
         user_arn = quicksight_user["Arn"]
-        user_email = quicksight_user["Email"].lower()
         user_role = quicksight_user["Role"]
         user_username = quicksight_user["UserName"]
         sso_id = quicksight_user["UserName"].split("/")[-1]
 
+        # Update the quicksight user's email address to match data workspace
+        try:
```
chopkinsmade, line 845: Instead of handling a django exception, you can use the `first()` function that returns `None` instead if there is no match, something like this (untested, may need tweaking)
```
user_email = quicksight_user["Email"].lower()
user = get_user_model().objects.filter(username=sso_id).first()
if user is not None:
    user_email = user.email
```

chopkinsmade (approved):

dataworkspace/dataworkspace/apps/applications/utils.py
niross: Cheers, I tend to lean towards the "ask for forgiveness" model although granted it seems to be a bit mixed in this codebase

PR #3051, opened 2024-02-23
feature/DT-1829-refactor-your-files-scheduled-task

chopkinsmade:
### Description of change

- Renamed the original scheduled job to make it obvious it is getting the S3 file size for all users
- Added a new celery job that can be used to sync an individual users S3 filesize
- Used a new lock when dealing with a single user sync, to avoid conflicts if the all users job and single user job is running at the same time
- Added test coverage for the task functions

### Checklist


niross (approved):

niross:
Loogs good 👍

PR #3052, opened 2024-02-23
A11y: Skip to content / page title change

CaitBarnard:
### Description of change
DT-1843 -> Update Data Catalogue page title to match page content
DT_1844 -> Fix href for 'Skip to content' on welcome page
DT-1846 -> Remove alt text from welcome page images as they are unnecessary for screen readers

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/templates/_main_no_container.html
+++ b/dataworkspace/dataworkspace/templates/_main_no_container.html
@@ -1,7 +1,7 @@
 {% extends '_base_no_footer.html' %}
 
 {% block main %}
-  <main id="main-content" role="main">
+  <main role="main">
     {% include 'partials/messages.html' %}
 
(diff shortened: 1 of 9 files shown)


dataworkspace/dataworkspace/templates/_main_no_container.html
```diff
@@ -1,7 +1,7 @@
 {% extends '_base_no_footer.html' %}
 
 {% block main %}
-  <main id="main-content" role="main">
+  <main role="main">
```
ian-leggett, line 4: We may need to be careful here as we do want the ID on on some pages as user will not be able to skip the main navigation. As long as the "data preview" page is the only page that uses _main_no_container.html then this seems okay. If we do this we will need to remove the skip to main link on data preview pages

ian-leggett (approved):

PR #3063, opened 2024-02-26
refactor: move out Superset

michalc:
### Description of change

This is more of a cross-repo refactor: moving things out to https://github.com/uktrade/data-workspace-superset.

Note that there is no automated deployment for Superset. Things are built and pushed locally. So this shouldn't break anything, because there is nothing to break.

### Checklist


superset/Dockerfile
chopkinsmade, line 1: There are still references to this Dockerfile in the docker-compose.yml. Will we ever need to run this locally alongside data workspace, or can all references be take out?

chopkinsmade (approved):

PR #3072, opened 2024-03-01
feature/DT-1852-keep-focus-on-filters

chopkinsmade:
### Description of change
When the search has finished, set the focus back to the original element that triggered the search

### Checklist


ian-leggett (approved):

dataworkspace/dataworkspace/static/search-v2.js
```diff
@@ -2298,6 +2298,13 @@ LiveSearch.prototype.formChange = function formChange(e) {
     pageUpdated = this.updateResults();
     pageUpdated.done(
       function () {
+        if (e && e.target && e.target.id) {
```
ian-leggett, line 2301: Just a small one, we could use `e?.target?.id`

PR #3146, opened 2024-04-10
feat: tidy up granting/revoking table permissions

michalc:
### Description of change

This does 4 things when fetching a user's existing permisssions

- Ensures that we don't have repeat schema/table pairs in the list of tables with existing permissions.
- Make sure we include permissions on views, materialized views, and partitioned tables - which we're doing on the (in progress) role-based permissions system.
- Removes what I think are unnecessary filters when finding a user's existing table permissions - the user would have permissions on some of these, but I suspect via ownership than directly granted on pg_class
- Makes sure to clear the user's permissions cache if we have made any changes so next time we don't make unnecessary changes to the database.

We are hopefully going to remove this soon, but I suspect this still has value to make the current setup clearer.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/core/utils.py
+++ b/dataworkspace/dataworkspace/apps/core/utils.py
@@ -837,6 +837,9 @@ def new_private_database_credentials(
                 )
             )
 
+        if tables_to_revoke or tables_to_grant:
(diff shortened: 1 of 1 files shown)


michalc:
Changed my mind on this one... too big/maybe unecessary/and I think slower

PR #3197, opened 2024-05-29
DT-2009/create Data workspace user inline feedback API for data flow

Teccolat:
### Description of change
This PR is to allow Data flow to ingest Data workspace user inline feedback data from Data workspace API.
All data available on data workspace database is exposed through the API.

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/api_v1/core/views.py
+++ b/dataworkspace/dataworkspace/apps/api_v1/core/views.py
@@ -13,12 +13,14 @@ from dataworkspace.apps.api_v1.core.serializers import (
     UserSatisfactionSurveySerializer,
     NewsletterSubscriptionSerializer,
     TeamSerializer,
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/apps/api_v1/core/views.py
```diff
@@ -156,3 +158,13 @@ class TeamViewSet(viewsets.ModelViewSet):
...
+
+
+class UserInlineFeedbackSurveyViewSet(viewsets.ModelViewSet):
+    """
+    API endpoint to list user satisfaction survey results for ingestion by data flow
+    """
+
+    queryset = UserInlineFeedbackSurvey.objects.all()
+    serializer_class = UserInlineFeedbackSurveySerializer
+    # pagination_class = PageNumberPagination
```
chopkinsmade, line 170: Looking at the classes above, I think this needs to be there and not commented out

PR #3212, opened 2024-06-24
Add method to validate user email domain

CaitBarnard:
### Description of change
- Redirect to either self-certify or request access based on user's email domain.
- Content changes for confirmation page

### Checklist

### DIFF
--- a/dataworkspace/dataworkspace/apps/request_access/views.py
+++ b/dataworkspace/dataworkspace/apps/request_access/views.py
@@ -257,15 +257,15 @@ def is_user_email_domain_valid(email):
     email_domain = email.split("@")[-1]
 
     valid_domains = [
-        "businessandtrade",
-        "beis",
-        "trade",
-        "digital.trade",
-        "fcdo",
-        "fco",
+        "businessandtrade.gov.uk",
+        "beis.gov.uk",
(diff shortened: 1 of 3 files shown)


dataworkspace/dataworkspace/apps/request_access/views.py
```diff
@@ -252,15 +252,28 @@ def get_context_data(self, **kwargs):
...
+    valid_domains = [
+        "businessandtrade",
+        "beis",
+        "trade",
+        "digital.trade",
+        "fcdo",
+        "fco",
+    ]
+
+    return any(email_domain.startswith(valid_domain) for valid_domain in valid_domains)
```
CaitBarnard, line 268: Technically this would allow emails such as test@digital.trade.anything.gov.uk etc. Any suggestions on tightening this logic would be appreciated

ian-leggett: What you could do is first check that a "@" exists, if it does split the string and get the last one. Then run that last part against all the domains. Something like 

```
def is_valid_email(email, valid_domains):
  split_email = email.split('@')[-1]
  return split_email in valid_domains
```

ian-leggett (approved):

PR #3234, opened 2024-07-22
Feat/dt 2053 your tables schema

lluli:
This is the table schema page for the service of adding a table to an existing catalogue page. 

- When a dataset has one schema, that will be automatically selected
- When a dataset has multiple schemas, a radio button form will allow you to select your chosen schema
- The content slightly changes depending if there is one or multiple schemas
- The schema gets passed in the URL

### DIFF
--- a/dataworkspace/dataworkspace/templates/datasets/add_table/table_schema.html
+++ b/dataworkspace/dataworkspace/templates/datasets/add_table/table_schema.html
@@ -1,7 +1,7 @@
  {% extends '_main.html' %}
 
 
-{% block page_title %}Your table’s schema{% endblock %}
+ {% block page_title %}Add Table - {{ model.name }} - {{ block.super }}{% endblock %}
 
 {% if backlink %}
   {% block go_back %}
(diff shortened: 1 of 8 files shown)


dataworkspace/dataworkspace/templates/datasets/add_table/table_schema.html
```diff
@@ -0,0 +1,27 @@
...
+
+{% block content %}
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">Your table’s schema</h1>
+
+        <h2 class="govuk-heading-m">Your table will be saved in '{{ schema }}' schema</h2>
+
+            <p class="govuk-body">This is the schema used by other tables in this catalogue item. Schemas are used to categorise data sources. Schemas are often named after the data provider e.g. HMRC.</p>
+                          </p>
```
ian-leggett, line 20: I think we can remove this closing paragraph

ian-leggett (commented):
I think we should also assert the title as we are dynamically setting it. Sorry I forgot to do this on my previous PR.

ian-leggett (approved):

dataworkspace/dataworkspace/tests/datasets/add_table/test_views.py
```diff
@@ -8,28 +9,122 @@
...
+        assert "Select an existing schema from this catalogue page" in header_two_text
+        assert (
+            "This is the schema used by other tables in this catalogue item. Schemas are used to categorise data sources. Schemas are often named after the data provider e.g. HMRC."
+            in paragraph_text
+        )
+        assert self.source.schema and self.source2.schema in radio_names
+
+    def test_table_schema_page_when_multiple_tables_have_same_schema(self):
+        self.source2 = factories.SourceTableFactory.create(
+            dataset=self.dataset, schema="test", table="table2"
```
ian-leggett, line 106: As you are expecting only one radio option to render it will probably be worth asserting the count. I would expect only one radio option to render with the value "test"

PR #3242, opened 2024-07-29
Feat/dt 2190 Remove tools access for users with expired certificates

ian-leggett:
### Description of change

Ticket [DT-2190](https://uktrade.atlassian.net/jira/software/projects/DT/boards/356?isInsightsOpen=true&selectedIssue=DT-2190)

This change adds a new Celery task that checks if any users tools certificates have expired. If they have, then a function runs that removes all tools access. This task will run every morning at 5:00am.

### Checklist

[DT-2190]: https://uktrade.atlassian.net/browse/DT-2190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dataworkspace/dataworkspace/apps/applications/utils.py
+++ b/dataworkspace/dataworkspace/apps/applications/utils.py
@@ -1,4 +1,3 @@
-import calendar
 import csv
 import datetime
 import json
@@ -64,6 +63,7 @@ from dataworkspace.apps.core.utils import (
     transaction_and_lock,
     new_private_database_credentials,
     postgres_user,
+    has_tools_cert_expired,
 )
 from dataworkspace.apps.applications.gitlab import gitlab_has_developer_access
 from dataworkspace.apps.datasets.constants import UserAccessType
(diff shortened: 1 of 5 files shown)


dataworkspace/dataworkspace/tests/applications/test_utils.py
```diff
@@ -1062,3 +1064,58 @@ def test_long_running_queries(self, mock_send_slack_message, mock_connections):
             ":rotating_light: Found 1 SQL query running for longer than 15 minutes "
             "on the datasets db."
         )
+
+
+class Test_remove_tools_access_for_users_with_expired_cert:
```
chopkinsmade, line 1069: Classes are Pascal cased, even when testing a single method

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1975,3 +1977,49 @@ def get_tool_url_for_user(user: get_user_model(), application_template: Applicat
...
+    user_model = get_user_model()
+    permissions_codenames = [
+        "start_all_applications",
+        "develop_visualisations",
+        "access_quicksight",
+        "access_appstream",
+    ]
+    permission_ids = Permission.objects.filter(codename__in=permissions_codenames).all()
+
+    def has_tools_cert_expired(cert_date):
```
chopkinsmade, line 2007: I would move this inner function out, and add tests dedicated to just returning a true or false in the different datetime conditions. For example, I would add a test forcing the current date to be a leap year instead of just using the real today's date which might not be a leap year. Search the code for `@freeze_time` for an example of to do this, where you can set what you want the result of `datetime.date.today()` to be

dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1975,3 +1977,49 @@ def get_tool_url_for_user(user: get_user_model(), application_template: Applicat
...
+
+    def remove_tools_access(user):
+        for permission_id in permission_ids:
+            user.user_permissions.remove(permission_id.id)
+
+    for user in user_model.objects.all():
+        user_has_access = user.user_permissions.filter(codename__in=permissions_codenames).exists()
+        user_profile = Profile.objects.filter(user=user.id)
+
+        if user_has_access and user_profile[0].tools_certification_date:
```
chopkinsmade, line 2021: I would add a test for checking the `user_has_access` is False, as it looks like both of the tests assign a user all permissions.

chopkinsmade (approved):

PR #3251, opened 2024-08-05
Descriptive table name and tests

lluli:
The validation for the Descriptive name page.

### DIFF
--- a/dataworkspace/dataworkspace/tests/datasets/add_table/test_views.py
+++ b/dataworkspace/dataworkspace/tests/datasets/add_table/test_views.py
@@ -195,3 +195,114 @@ class TestAddTable(TestCase):
             "You must not add a table that would change the security classification of the catalogue item you're adding to."
             in paragraph_text
         )
+
+    def test_descriptive_table_name_page(self):
+
+        response = self.client.get(
+            reverse(
+                "datasets:add_table:descriptive-name",
+                kwargs={"pk": self.dataset.id, "schema": self.source.schema},
+            ))
+
+        soup = BeautifulSoup(response.content.decode(response.charset))
+        header_one = soup.find("h1")
+        paragraph = soup.find("p")
+        header_one_text = header_one.contents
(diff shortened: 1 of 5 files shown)


dataworkspace/dataworkspace/tests/datasets/add_table/test_views.py
```diff
@@ -195,3 +195,114 @@ def test_classification_check_page(self):
...
+
+        error_header = soup.find("h2")
+        error_header_text = error_header.contents
+        error_messages = soup.find('ul', class_='govuk-list govuk-error-summary__list')
+        links = error_messages.find('a')
+        link_text = links.contents
+        assert "\n    There is a problem\n  " in error_header_text
+        assert "Descriptive name cannot contain the word underscores" in link_text
+
+    def test_error_shows_when_descriptive_table_name_input_contains_underscores(self):
```
ian-leggett, line 292: This test name is the same as L274. I believe this is just testing an empty input, is that right?

dataworkspace/dataworkspace/tests/datasets/add_table/test_views.py
```diff
@@ -195,3 +195,114 @@ def test_classification_check_page(self):
...
+
+        soup = BeautifulSoup(response.content.decode(response.charset))
+        assert response.status_code == 200
+
+        error_header = soup.find("h2")
+        error_header_text = error_header.contents
+        error_messages = soup.find('ul', class_='govuk-list govuk-error-summary__list')
+        links = error_messages.find('a')
+        link_text = links.contents
+        assert "\n    There is a problem\n  " in error_header_text
```
ian-leggett, line 235: Are we able to remove the carriage returns and just use "assert "There is a problem" in error_header_text"

ian-leggett (commented):
Looking good. I would also add a test that checks that the page title is populated as well as the back link as these are dynamically set based on the context you pass to the template.

PR #3252, opened 2024-08-05
Feat/dt 2056 descriptive name 2

lluli:
Next step of add table journey:  the validation for the Descriptive name page.

### DIFF
--- a/dataworkspace/dataworkspace/apps/datasets/add_table/views.py
+++ b/dataworkspace/dataworkspace/apps/datasets/add_table/views.py
@@ -4,7 +4,7 @@ from django.http import HttpResponseRedirect
 
 from dataworkspace.apps.datasets.utils import find_dataset
 from dataworkspace.apps.datasets.constants import DataSetType
-from dataworkspace.apps.datasets.add_table.forms import TableSchemaForm
+from dataworkspace.apps.datasets.add_table.forms import TableSchemaForm, DescriptiveNameForm
 
 
 class AddTableView(DetailView):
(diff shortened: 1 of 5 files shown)


dataworkspace/dataworkspace/apps/datasets/add_table/views.py
```diff
@@ -77,5 +77,29 @@ def get_context_data(self, **kwargs):
...
+        dataset = find_dataset(self.kwargs["pk"], self.request.user)
+        ctx["model"] = dataset
+        ctx["backlink"] = reverse("datasets:add_table:classification-check", args={self.kwargs["pk"], self.kwargs['schema']})
+        ctx["nextlink"] = ''
+        return ctx
+
+    def form_valid(self, form):
+        clean_data = form.cleaned_data
+        descriptive_name = clean_data["descriptive_name"]
+        print('descriptive_name', descriptive_name)
```
ian-leggett, line 99: We should remove the print here

dataworkspace/dataworkspace/apps/datasets/add_table/views.py
```diff
@@ -77,5 +77,29 @@ def get_context_data(self, **kwargs):
...
+        self.request.session["descriptive_name"] = descriptive_name
```
ian-leggett, line 100: Do we need this?

ian-leggett (commented):
Looking good. I would also add a test that checks that the page title is populated as well as the back link as these are dynamically set based on the context you pass to the template.

PR #3271, opened 2024-08-15
bump aiohttp package from 3.9.4 to 3.10.2 version

Teccolat:
### Description of change

Update aiohttp package from 3.9.4 to 3.10.2 version.

### Checklist


requirements-dev.txt
```diff
@@ -1,10 +1,14 @@
 #
-# This file is autogenerated by pip-compile with python 3.9
-# To update, run:
+# This file is autogenerated by pip-compile with Python 3.11
```
ian-leggett, line 2: The requirements file has been generated using python 3.11 and we are currently using 3.9. Could we re-run this with 3.9?

PR #3360, opened 2024-11-15
DT-2423 add manual 'Add User' button to manage permissions page

lluli:
Button introduced to Manage Permissions page to allow IAM/IAOs to manually add users to datasets. 

Will be redirected to the search feature of the particular dataset.

### DIFF
--- a/dataworkspace/dataworkspace/static/js/react/features/confirm-remove-user/ConfirmRemoveUser.tsx
+++ b/dataworkspace/dataworkspace/static/js/react/features/confirm-remove-user/ConfirmRemoveUser.tsx
@@ -1,7 +1,6 @@
 import React, { useState } from 'react';
 
 import { Button } from '@govuk-react/button';
-import { H2 } from '@govuk-react/heading';
 import Table from '@govuk-react/table';
 import { Paragraph } from 'govuk-react';
 import styled from 'styled-components';
@@ -42,10 +41,9 @@ const ConfirmRemoveUser = ({
   return (
     <>
       {data.length < 1 ? (
-        <H2>There are currently no authorized users</H2>
+        <div></div>
       ) : (
         <div>
-          <H2>Users who have access</H2>
           <Table>
             {data.map((user) => (
               <Table.Row key={user.id}>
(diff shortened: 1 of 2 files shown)


dataworkspace/dataworkspace/static/js/react/features/confirm-remove-user/ConfirmRemoveUser.tsx
```diff
@@ -42,10 +41,9 @@ const ConfirmRemoveUser = ({
   return (
     <>
       {data.length < 1 ? (
-        <H2>There are currently no authorized users</H2>
+        <div></div>
```
ian-leggett, line 44: You can just use an empty fragment here as a `div` could come with global styling such as padding and margin. You can change `<div></div>` to `<></>` or `null` maybe. I think you could just write this as `{data.length && <div>Some HTML goes here....</div>` so you might not need the ternary.

dataworkspace/dataworkspace/templates/datasets/manage_permissions/edit_access.html
```diff
@@ -57,9 +57,18 @@ <h2 class="govuk-heading-m">Users who have requested access</h2>
...
                   </tr>
               {% endfor %}
-          </tbody>
+              
         </table> 
         {% endif %}
+        <h2 class="govuk-heading-l">Users who have access</h2>
+          <a class="govuk-link" href="{% url 'datasets:search_authorized_users' obj.id summary.id %}"><button type="submit" class="govuk-button govuk-button--secondary govuk-!-static-margin-bottom-6" data-module="govuk-button">
+            Add User
+          </button></a>
```
ian-leggett, line 66: Is it possible to update the unit tests for this part?

dataworkspace/dataworkspace/templates/datasets/manage_permissions/edit_access.html
```diff
@@ -57,9 +57,18 @@ <h2 class="govuk-heading-m">Users who have requested access</h2>
...
+    
+                {% if authorised_users.count == 0 %}
+                  <h2 class="govuk-heading-m">There are currently no authorized users</h2>
+                {% endif %}
```
ian-leggett, line 70: Will this ever happen? There will always be at least an IAM that has access.

dataworkspace/dataworkspace/templates/datasets/manage_permissions/edit_access.html
lluli: added

dataworkspace/dataworkspace/templates/datasets/manage_permissions/edit_access.html
lluli: removed

PR #3421, opened 2025-01-21
fix/add initial backend view and url route DO NOT MERGE

JamesPRobinson:
### Description of change
Rough outline of how we could use the accepted/dimissed cookie. We could call the backend with js* or turn the survey into a django form.

* _e.g._
``` js
<link id="accepted" onclick="closeBanner(true)">
<link id="dismissed" onclick="closeBanner(false)">

<script>
const closeBanner = function(hasAccepted) {
const status = "accepted" ? hasAccepted : "dismissed"; 
post("set-notification-cookie", args={status=status, campaign="survey"}) // etc etc
}
</script>

### DIFF
--- a/dataworkspace/dataworkspace/apps/core/views.py
+++ b/dataworkspace/dataworkspace/apps/core/views.py
@@ -16,6 +16,7 @@ from django.http import (
 )
 from django.shortcuts import render
 from django.urls import reverse
+from django.utils.timezone import now, timedelta
 from django.views import View
 from django.views.generic import FormView
 from requests import HTTPError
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: dataworkspace/dataworkspace/apps/core/views.py [hunk 1]
```diff
@@ -463,3 +464,17 @@
...
+        # id_campaign = request.POST.get("id_campaign")
+        id_campaign = "placholder"
+        # config_notification_banner = NotificationBannerConfig.objects.filter(campaign_id=id_campaign)
+        # date_expiry = config_notification_banner.expiry_date
+        date_expiry = now() + timedelta(days=1)
+        if now() >= date_expiry:
+            response = JsonResponse({"message": f"campaign {id_campaign} expired"})
+        else:
+            action = request.POST.get("action")
+            response.set_cookie(f"{id_campaign}_{action}", "true", expires=date_expiry)
```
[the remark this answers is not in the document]
JamesPRobinson: I would have some checks here to make sure they fall within a range of accepted values

dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -174,9 +174,31 @@ def _get_tags_as_dict():
     return tags_dict
 
 
+def show_survey_banner(request):
```
JamesPRobinson, line 177: Unfortunately would need to delete this code after survey expiry is over

dataworkspace/dataworkspace/apps/core/views.py
```diff
@@ -463,3 +464,17 @@ def form_valid(self, form):
         if cleaned["contact_type"] == form.ContactTypes.GET_HELP:
             return HttpResponseRedirect(reverse("support"))
         return HttpResponseRedirect(reverse("feedback"))
+
+
+class SetNotificationCookie(View):
+    def post(self, request, *args, **kwargs):
```
JamesPRobinson, line 470: I've tried to make this generic but maybe would need to hardcode to survey campaign to avoid the security issues flagged below

ian-leggett: Could you then set a base class and extend from it? Or perhaps a util function can be used here?

PR #3429, opened 2025-01-27
Feat/dt 2545 notification banner open link new tab

JamesPRobinson:
### Description of change
POST request when 'accepting' doesn't fire when navigating away from page. Could wait for request to complete or could just have it so link always opens in new tab.


lluli (approved):

dataworkspace/dataworkspace/templates/partials/notification_banner.html
```diff
@@ -1,4 +1,7 @@
 <style>
+  .notification-banner__background {
+    background-color: #003399;
+  }
```
ian-leggett, line 4: We just need to move this out to a separate SASS component file 👍

PR #3435, opened 2025-02-03
fix/run isort import linting tool

JamesPRobinson:
tidy up import order in one PR


ian-leggett (approved):
This would be great if we can get this working in CI so we don't have to manually run it every time. I will create a tech card in the backlog 👍

PR #3442, opened 2025-02-10
fix/data insights register w proxy, more test coverage, filtering

JamesPRobinson:
### Description of change
Requests via proxy.py were broken while unregistered.
Further filtering of users to not return users where no owned datasets.
field 'access_request_count' should return 0 not None when no access requests
Increased test coverage.


dataworkspace/proxy.py
```diff
@@ -332,6 +332,7 @@ def is_dataset_requested(request):
         return (
             request.url.path.startswith("/api/v1/dataset/")
             or request.url.path.startswith("/api/v1/reference-dataset/")
+            or request.url.path.startswith("/api/v1/data_insights/owner_insights")
```
JamesPRobinson, line 335: Think this is the source of the error

dataworkspace/dataworkspace/apps/api_v1/data_insights/views.py
```diff
@@ -1,15 +1,29 @@
...
+from dataworkspace.apps.datasets.models import DataSet, DataSetType
 
 
 class OwnerInsightsViewSet(viewsets.ModelViewSet):
     """
     API endpoint to list EvenLog items for consumption by data flow.
     """
 
-    queryset = get_user_model().objects.all()
+    queryset = (
```
JamesPRobinson, line 16: want to leave out users who don't have any owned datasets in the response

PR #3465, opened 2025-02-21
feat/dt-2601: Approval page for project owner

ian-leggett:
### Description of change

Ticket DT-2601

This is the start of the third approver feature. This change covers start of the approval page for when the user owns the visualisation to which they want to publish. The work is hidden behind a feature flag `THIRD_APPROVER` so you will need to turn this on. As we currently have no way to work with an instance of Gitlab locally I have bypassed the Gitlab API calls using an environment variable called `GITLAB_FIXTURES`. If you set this env var you will use static fixtures that fake the API response object which will then allow you to access `/visualisations/173/approvals` where `173` is the project id from one of the fixtures.

### Checklist


dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -694,7 +697,10 @@ def visualisation_users_give_access_html_view(request, gitlab_project_id):
             "manage visualisations", "To request access"
         )
 
-    gitlab_project = _visualisation_gitlab_project(gitlab_project_id)
+    if settings:
```
JamesPRobinson, line 700: should this be settings.GITLAB_FIXTURES ?

dataworkspace/dataworkspace/apps/applications/gitlab.py
```diff
@@ -145,3 +153,24 @@ def _ensure_user_has_manage_unpublish_perm(user):
...
+    if settings.GITLAB_FIXTURES:
+        (current_gitlab_project_user,) = get_fixture("project_member_fixture.json")
+    else:
+        (current_gitlab_project_user,) = gitlab_api_v4(
+            "GET",
+            f"/projects/{gitlab_project_id}/members/all",
+            params=(("user_ids", str(gitlab_user["id"])),),
+        )
+
+    return bool(current_gitlab_project_user["access_level"] == int(MAINTAINER_ACCESS_LEVEL))
```
JamesPRobinson, line 168: Should this be OWNER_ACCESS_LEVEL at 50?

dataworkspace/dataworkspace/apps/applications/views.py
ian-leggett: Good spot thanks 👍

dataworkspace/dataworkspace/apps/applications/gitlab.py
ian-leggett: This is a bit confusing this one as when you create a visualisation in Gitlab you are set as a Maintainer which is at level 40. So for us we treat maintainer as the owner.

JamesPRobinson (approved):

PR #3481, opened 2025-03-06
Feat/dt 2611 invalid visualisation approval scenario

Teccolat:
### Checklist


dataworkspace/dataworkspace/apps/applications/utils.py
```diff
@@ -1943,3 +1943,46 @@ def has_all_three_approval_types(approvals: list[dict]) -> bool:
...
+
+    for approver_type in ["owner", "peer reviewer", "team member"]:
+        approver_types[approver_type].sort(key=lambda x: x["date_approved"])
+
+    project_approvals = [
+        approver_types[approver_type][0]
+        for approver_type in ["owner", "peer reviewer", "team member"]
+        if approver_types[approver_type]
+    ]
+    is_approved_by_all = has_all_three_approval_types(project_approvals)
```
JamesPRobinson, line 1987: can this just just return project approvals and then use that in the function that's calling it to call has_all_three_approval_types to get is_approved_by_all?

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -1524,13 +1498,27 @@ def visualisation_publish_html_view(request, gitlab_project_id):
     return HttpResponse(status=405)
 
 
-def _visualisation_is_approved(application_template):
```
JamesPRobinson, line 1527: I think we'll need to keep both the new way and old way of doing things and wrap it around the feature flag for now just so we don't break compatibility.

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -1614,7 +1626,7 @@ def _set_published_on_catalogue_item(request, gitlab_project, catalogue_item, pu
 
         return redirect(request.path)
 
-    if visualisation_approved is False:
+    if is_approved_by_all is False and len(project_approvals) < 3:
```
JamesPRobinson, line 1629: what does len(project_approvals) < 3 do?

Teccolat: This was a mistake. len(project_approvals) is unnecessary

dataworkspace/dataworkspace/apps/applications/views.py
```diff
@@ -1524,13 +1498,35 @@ def visualisation_publish_html_view(request, gitlab_project_id):
...
+        visualisation=application_template, approved=True
+    ).all()
+
+    if waffle.flag_is_active(request, settings.THIRD_APPROVER):
+        if settings.GITLAB_FIXTURES:
+            project_members = get_fixture("project_members_fixture.json")
+        else:
+            project_members = gitlab_project_members(gitlab_project)
+
+    project_approvals = visualisation_approvals(dw_approvals, project_members)
```
JamesPRobinson, line 1513: Will this new function work with the old journey? Maybe we should build it in staging to visualise it better

JamesPRobinson: As in, how did the old journey get project approvals before?

JamesPRobinson (approved):

PR #3492, opened 2025-03-14
feat: Derived Pipelines custom schedules

tombrooks248:
### Description of change

Allow users who have access to /Pipelines to write custom schedules 

### Why
As DPS we receive requests to run Derived Pipelines at different frequencies than the standard options and would like to be able to meet these requests.

### Checklist


dataworkspace/dataworkspace/apps/datasets/models.py
```diff
@@ -2898,6 +2898,7 @@ class Pipeline(TimeStampedUserModel):
     schedule = models.TextField(
         choices=PipelineScheduleType.choices, default=PipelineScheduleType.DAILY
     )
+    custom_schedule=models.CharField(max_length=16, blank=True)
```
tom-swart, line 2901: I would increase this maybe to 64 or 128 as you can get some chunky cron expressions sometimes

dataworkspace/dataworkspace/apps/datasets/pipelines/forms.py
```diff
@@ -64,6 +64,19 @@ class BasePipelineCreateForm(GOVUKDesignSystemModelForm):
             label_is_heading=False, extra_label_classes="govuk-!-font-weight-bold"
         ),
     )
+    custom_schedule = GOVUKDesignSystemCharField(
```
JamesPRobinson, line 67: I know there's a note here explaining the field but it would be worth adding some code to the form's 'clean' function  just to make sure there's no confusion if someone selects 'daily' and then adds some custom cron schedule or vice-versa.

So something like
``` python
def clean(self):
    if self.schedule_field == 'customschedule' and not self.custom_schedule: # is null or empty
        raise ValueError("schedule field selected with 'Custom Schedule' but cron expression field empty")
    elif self.schedule_file != 'customschedule' and self.custom_schedule:
         raise ValueError("Custom CRON expressions can only be entered with 'Custom Schedule' selected")
```

tombrooks248: Hey - been a mad week - just coming back to this. 

Thanks sounds good - I've added in something which seems to elicit all the behaviours I think we are looking for. It uses the self.cleaned_data a few times, which I don't really understand I'm just using it cos I saw the other functions use it and it seems to work. 

If there are better/ cleaner ways of doing this, I'm super happy to hear any direction.

JamesPRobinson (approved):

dataworkspace/dataworkspace/apps/datasets/models.py
tombrooks248: Sure done.

PR #3534, opened 2025-04-01
3rd approver visualisation current permissions

JamesPRobinson:
### Description of change
Change journey so that approval is saved at point of approval, before who the approver is and their approval would be checked dynamically.

This means:
- adding an approval_type ('team member', 'peer reviewer', 'owner') categorical field on the VisualisationApproval model
- adding an approval_date (formatted to render on the frontend) field on VisualisationApprvoal
- refactoring the tests to include these fields and also not use GitLab API to get this information every time


paulgain (approved):
Thanks for taking me through it :-)

PR #3538, opened 2025-04-01
Feat/dt 2686 create zendesk ticket after unpublish

Teccolat:
### Checklist


dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -1540,6 +1548,39 @@ def form_valid(self, form):
...
+            reverse("datasets:dataset_detail", args=[self.obj.id])
+        )
+        # In Dev Ignore the API call to Zendesk and notify
+        if settings.ENVIRONMENT != "Dev":
+            if isinstance(dataset, (DataSet, VisualisationCatalogueItem)):
+                zendesk.notify_unpublish_catalogue_page(
+                    request,
+                    dataset,
+                )
+                dataset.save()
```
JamesPRobinson, line 1566: is this line needed?

dataworkspace/dataworkspace/apps/datasets/views.py
```diff
@@ -1540,6 +1548,39 @@ def form_valid(self, form):
...
+    def post(self, request, *args, **kwargs):
+        dataset = find_dataset(kwargs["pk"], request.user)
+        dataset.published = False
+        dataset.save()
+        absolute_url = self.request.build_absolute_uri(
+            reverse("datasets:dataset_detail", args=[self.obj.id])
+        )
+        # In Dev Ignore the API call to Zendesk and notify
+        if settings.ENVIRONMENT != "Dev":
+            if isinstance(dataset, (DataSet, VisualisationCatalogueItem)):
```
JamesPRobinson, line 1561: Is this check needed? Is unpublish only for these types of dataset?

Teccolat: We don't need to check if its an instance of DataSet but i need to check if the environment is a dev or not. This is to avoid a call for API while working on it locally.

